PR #2, opened 2020-01-28
Make scripts be multiplatform

NickSeagull:
This PR makes the required changes to make sure that all scripts/tools that we use work on Windows. I've added also a script to _check all the things_


scripts/check-all-the-things.ps1
```diff
@@ -0,0 +1,11 @@
...
+# but in order to maintain backwards compatibility
+# with Powershell <7.0 (what most Windows users have preinstalled)
+# we write it in the old-school style
+lerna clean --yes
+lerna run clean
+if ($?) { lerna bootstrap }
+if ($?) { lerna run compile }
+if ($?) { lerna run fix-lint }
+if ($?) { lerna run lint }
+if ($?) { lerna run test }
```
alvaroloes: What does $? mean? Does it contain the result of the previous command?

NickSeagull: yep

PR #6, opened 2020-02-05
Add docs on how to use the /commands endpoint

NickSeagull:


docs/documentation/03-commands.md
```diff
@@ -2,7 +2,17 @@
...
+
+```json
+{
+  "typeName": "CreateCart"  // name of the command class
+  "value": {
+  "cartId": "aaaa",         // contents of the fields of the class
+    ...
+  }
+}
+```
```
alvaroloes: This is already documented in [here](docs/documentation/09-rest-api.md) although we should improve that documentation.
I think it would be enough to just have a link here pointing to the commands section of the API documentation

NickSeagull: fixed in https://github.com/boostercloud/booster/pull/6/commits/9bf7c6fbdaf0f25f86fa9d2b74a75ab2058573f7 thanks

PR #17, opened 2020-02-19
Use homogeneous case for cartId

JorgeRdg:


docs/examples/cart-service/README.md
```diff
@@ -75,7 +75,7 @@ export class CartChanged {
     readonly quantity: number
   ) {}
 
-  public entityID(): string {
+  public entityId(): string {
```
javiertoledo: This won't compile if you use the camelcase version, we're using all-caps for IDs and acronyms as a convention

PR #18, opened 2020-02-19
Use homogeneous case for 'ID'

JorgeRdg:


docs/documentation/04-events.md
```diff
@@ -23,16 +23,16 @@ Booster will generate a class for you in the `src/events` folder:
 export class CartPaid {
   public constructor(readonly cartID: UUID, readonly confirmationToken: string) {}
 
-  public entityID(): UUID {
+  public entityId(): UUID {
```
javiertoledo: This won't compile this way, we're using all case for IDs and acronyms

PR #89, opened 2020-05-21
Fix-#64: Fix project template lint configuration

### LINKED ISSUE #64: Fix project template lint configuration
**Description**
**Given** that I am a user that has just created a Booster project from template

**When** I run the linter

**Then** I get an error

**Steps to reproduce:**

1. Create a booster project from template
2. run `npm run lint`

**Steps to fix it:**

1. Add an eslint configuration file, e.g. .eslintrc.js
2. Update the following dependencies:

```
"@typescript-eslint/eslint-plugin"
"@typescript-eslint/parser" 
```

glammers1:
## Description
Closes #64 

## Changes
I added some files related to ESLint/Prettier for them to be added to the generated projects. I've used the same configuration we have for Booster so that all Booster's children have the same one. Let me know if I should change any rules.

## Checks
- [X] Project Builds

## Additional information

Screenshot 2020-05-21 at 20 51 47


packages/cli/src/templates/project/package-json.ts
```diff
@@ -31,6 +33,7 @@ export const template = `{
   "repository": "{{{repository}}}",
   "scripts": {
     "lint": "eslint --ext '.js,.ts' **/*.ts",
+    "fix-lint": "eslint --quiet --fix --ext '.js,.ts' **/*.ts",
```
tainguyenbui, line 36: great stuff, what do you think about renaming these scripts to `lint:check` and `lint:fix`? those changes may also require refactoring some pipeline steps

glammers1: @tainguyenbui I can do that if you think those are better names or at least more common ones (I think so). However, take into account that this is the template to generate a project, it's not the Booster project itself, so what do you mean with `may also require refactoring some pipelines steps`? Do you mean to make the same change in the Booster project? Let me know, and I'll rename those actions only here or on both sides.

tainguyenbui: @glammers1, in that case, it shouldn't break anything. Yes indeed, the use of `:` is a more standard way, and also makes it tidier

glammers1: Done f1a7c7fac56ed6844f76ec602c2732498ea28739. I also opened an issue #96 to make the same change in Booster.

PR #121, opened 2020-06-01
Use the new "waitForIt" function in test to allow polling results with longer timeouts

alvaroloes:
## Description
Using fixed timeouts in integration tests were given inconsistent results. Some times the tests passed and sometimes we get a timeout and tests didn't pass because they were running slower than expected, so we needed to increase the time even more.

The function `waitForIt` executes a function every second until a condition is met. It keeps doing that for 1 minute, after that, it throws an exception (these times are configurable).

Inside the function passed to waitForIt should be a query that gets the result you want to test. In the condition, you should put the expectation over that result.

## Checks
- [x] Project Builds


alvaroloes:
bot: integration

packages/framework-integration-tests/integration/aws/utils.ts
```diff
@@ -199,6 +199,11 @@ export async function graphQLClient(authToken?: string): Promise<ApolloClient<No
   return new ApolloClient({
     cache: cache,
     link: link,
+    defaultOptions: {
+      query: {
+        fetchPolicy: 'no-cache',
+      },
+    },
```
NickSeagull, line 206: Is this intended?

packages/framework-integration-tests/integration/aws/utils.ts
```diff
@@ -258,3 +263,37 @@ export async function countSnapshotItems(): Promise<number> {
 export async function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms))
 }
+
+//
+export async function waitForIt<TResult>(
+  tryFunction: () => Promise<TResult>,
+  checkResult: (result: TResult) => boolean,
+  tryEveryMs = 1000,
```
NickSeagull, line 271: I'd call this `millisInterval`, from this name I infer it is a `bool`

alvaroloes:
bot: integration

packages/framework-integration-tests/integration/aws/utils.ts
alvaroloes: Yes, if we have the cache enabled in tests, they will fail (we do some retries)

packages/framework-integration-tests/integration/aws/utils.ts
alvaroloes: The idea was to have a name like: "Try every X ms"

PR #128, opened 2020-06-02
Fix: Await for command handlers to finish

alvaroloes:
## Description
This PR fixes a bug that happened always when the command handler does an asynchronous operation (uses `await` for example). Booster called the command handler without `await`, so the provider lambda finished before giving a chance to the command handler to finish.

## Changes
* The return type of command handlers is now `Promise<void>` as it should have been
* Added the corresponding await
* Do not handle events after the command handler finished if there were no events registered
* Added new tests for these new cases.

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-core/src/booster-command-dispatcher.ts
+++ b/packages/framework-core/src/booster-command-dispatcher.ts
@@ -36,7 +36,7 @@ export class BoosterCommandDispatcher {
     // the command inputted by the user.
     const register = new Register(commandEnvelope.requestID, commandEnvelope.currentUser)
     this.logger.debug('Calling "handle" method on command: ', command)
-    command.handle(register)
+    await command.handle(register)
     this.logger.debug('Command dispatched with register: ', register)
     await RegisterHandler.handle(this.config, this.logger, register)
   }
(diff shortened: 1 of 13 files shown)


packages/framework-core/src/booster-command-dispatcher.ts
```diff
@@ -36,7 +36,7 @@ export class BoosterCommandDispatcher {
     // the command inputted by the user.
     const register = new Register(commandEnvelope.requestID, commandEnvelope.currentUser)
     this.logger.debug('Calling "handle" method on command: ', command)
-    command.handle(register)
+    await command.handle(register)
```
alvaroloes, line 39: This was the culprit!

packages/framework-core/src/booster-register-handler.ts
```diff
@@ -10,6 +10,9 @@ import {
 
 export class RegisterHandler {
   public static async handle(config: BoosterConfig, logger: Logger, register: Register): Promise<void> {
+    if (register.eventList.length == 0) {
+      return
+    }
```
alvaroloes, line 15: Without this check, the provider throws an error when we try to store an empty array of events

alvaroloes:
bot: integration

PR #151, opened 2020-06-03
Your first booster app in 10 minutes

itrion:
## Description
This PR adds a section with the whole `Your First Booster App in 10 Minutes` guide

## Changes
Additionally, I did some fixes on the previous text but not big changes

## Checks
- [x] Project Builds
- [x] Updated documentation accordingly at [the docs repo](https://github.com/boostercloud/docs)
- [x] Added a link pointing to the documentation PR in this PR

## Additional information
I'll be adding tickets for:
- Ensuring there are no broken links in the docs
- Add links to the list of features in the `More Functionalities` section


docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+    readonly author: string
+  ) {}
+
+  public entityID(): UUID {
+    return this.postId;
+  }
+}
+```
+
+Now that we have an event, we can make the `CreatePost` command to emit it. Let's change
```
rdiaz82, line 290: I'm not sure about the use of `make` here. What do you think?

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+`Post`. To do so, we will create an `entity`.
+
+```bash
+boost new:entity Post --fields title:string content:string author:string --reduces PostCreated
+```
+
+This time Booster has created a file called `Post.ts` in the `src/entities` directory.
+An entity is a reduction function applied to a series of events. For this example, we won't
+take into account the previous state of the `Post`, so we will create a new one
+from the Event's data.
```
rdiaz82, line 312: Although it is not directly related with this PR, do we have in mind create other examples using the previous states?  I think that they will help a lot especially for more complex use cases.

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+}
+```
+
+#### 6. Deployment
+
+We have everything we need to deploy our application to the cloud. It is as simple as
+running this command:
+
+```bash
+boost deploy -e production
```
rdiaz82, line 384: what does mean the environment? Are there any differences if I deploy the booster app using development, testing or whatever value in that option?

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+{
+  "data": {
+    "CreatePost": true
+  }
+}
+```
+
+Note:
+
+> The IDs are generated on the client-side. For this example, we are not validating ID
```
rdiaz82, line 454: I don't know why but the first time that I read the sentence, it seems to me that there is also the possibility to configure the server-side ID generation

docs/README.md
itrion: Thank you @rdiaz82 for pointing this out. Yep, we have already captured an examples section 👍

docs/README.md
itrion: I'm not really sure because [make is super-polysemic](https://www.merriam-webster.com/dictionary/make). In that sentence it sounds to me like _now we can make this thing to happen_ but I'll change it by `edit` because it keeps the sense and we avoid using make.

docs/README.md
itrion: Excellent questions, the answer is I don't know :dumb: but I wouldn't explain the meaning and options in this 10m guide. Those will appear in the deploying section, and surely every provider will also need an explanation of what is the difference in using every environment.

Maybe @alvaroloes or @tainguyenbui can answer this better.

docs/README.md
itrion: You are right @rdiaz82, thank you for pointing this out. I'll rephrase this to make the point clearer

itrion: what do you think about this:

> With Booster, IDs are generated on the client-side. When running production applications
> consider adding validation for ID uniqueness. In this example, we have used [a UUID generator](https://www.uuidgenerator.net/version4)

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+    return new Post(event.postId, event.title, event.content, event.author)
+  }
+}
+```
+
+#### 5. First read model
+
+So far, we have been adding data to our blog. Now, we will retrieve our data using read
+models. A read model allows us to query our entire entities or just a subset of their
+attributes. Let's project our `Post` Entity:
```
javiertoledo, line 331:
```suggestion
In a real application, we rarely want to publish our entire domain model (entities) including all their fields, and we need to define which roles can access which subset of the data. We use ReadModels for that, to define the resources that client applications can query or subscribe to. We build our Read Models by projecting entities, let's project our `Post` Entity to produce a `PostReadModel`:
```

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+There are two things to do when creating a read model:
+
+1. Define who is authorized to query it
+1. and filter out the entity fields that are not needed in the read model
+
+Read models and commands constitute the public API of a Booster application. With the
+`CreatePost` command, we authorized `all` to execute it. This time we will be doing the
+same for the read model. Furthermore, We will exclude the `content` field from the `Post`
+entity so it won't be visible when querying this read-model. To do so, edit the class to
+look like this:
```
javiertoledo, line 356:
```suggestion
Read models and commands compound the public API of a Booster application. We allowed
anyone to submit the `CreatePost` command. We will allow anyone to query our read model too. 

Just for learning, we will show you how to exclude the `content` field from the `Post` entity so it won't be visible
when querying this read-model, you can add it later if you want to keep working in this application later. 

To authorize anyone to query the read model and exclude the `content` field, edit the class to look like this:
```

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+      return new PostReadModel(entity.id, entity.title, entity.author);
+  }
+
+}
+```
+
+#### 6. Deployment
+
+We have everything we need to deploy our application to the cloud. It is as simple as
+running this command:
```
javiertoledo, line 381:
```suggestion
At this point, we've learned how to create a publicly accessible command, emit an event to store the data, declare how the event is reduced into an entity, that represents our system internal state, and projected it into a read model that is publicly accessible, so you already know the basics to build event-driven applications with Booster. 

Let's deploy our application to the cloud to see it working. It is as simple as running the deploy command:
```

docs/README.md
javiertoledo: Booster allows defining more than one environment, similar to other frameworks like Rails. We introduced this feature when we started working in the local runtime, to be able to run the project locally using the `development` environment or publish it to the cloud using the `production` environment. There are no limitations on how you use this feature, so you could also configure staging environments or have personal environments for the developers using different accounts or parameterizing the app name.

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+
+Let's get started testing the project. We will perform three actions:
+
+- Add a couple of posts
+- Retrieve all posts
+- Retrieve a specific post
+
+Booster applications provide you with a GraphQL API out of the box. Commands are mapped
+to mutations and read models to queries. To perform calls to the GraphQL API, you can use
+any HTTP client you want; we recommend you to use [Postwoman](https://postwoman.io/graphql).
```
javiertoledo, line 409:
```suggestion
any HTTP client you want; we recommend you to use [Postwoman](https://postwoman.io/graphql), which is free and includes great support for GraphQL.
```

docs/README.md
javiertoledo: Well, this will change soon, we've planned to return the `entityID`s of all events emitted, instead of just that unmeaningful "true"

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+query {
+  PostReadModels {
+    id
+    title
+    author
+  }
+}
+```
+
+It should response something like:
```
javiertoledo, line 472:
```suggestion
It should respond with something like:
```

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+    "PostReadModel": {
+      "id": "95ddb544-4a60-439f-a0e4-c57e806f2f6e",
+      "title": "Build a blog in 10 minutes with Booster",
+      "author": "Boosted developer"
+    }
+  }
+}
+```
+
+#### 8. Removing stack
```
javiertoledo, line 521:
```suggestion
#### 8. Removing the stack
```

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+
+Now, let's destroy all the infrastructure created for running the application in the cloud.
```
javiertoledo, line 523:
```suggestion
It is convenient to destroy all the infrastructure created after you stop using it to avoid generating cloud resource costs.
```

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+Execute the following command from the root of the project. For security reasons, you have
```
javiertoledo, line 524: I think that security implies other more dramatic meanings, like avoiding data leaks or something like that, safety sounds softer to me...

```suggestion
Execute the following command from the root of the project. For safety reasons, you have
```

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+#### 9. More functionalities
+
+The are many other options for your serverless backend built with Booster Framework:
+
+- Authorize commands and read models based on different roles
+- Use GraphQL subscriptions
+- Make events to trigger other events
+- Serving static content
+- Reading entities within command handlers to apply domain-driven decisions
+- and much more...
```
javiertoledo, line 547:
```suggestion
- Build more complex authorization schemas for commands and read models based on user roles
- Use GraphQL subscriptions to get updates in real-time
- Make events trigger event handlers that can run code or trigger other events
- Deploy static content
- and much more...
```

We're still discussing this => Reading entities within command handlers to apply domain-driven decisions

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+
+Please continue reading to get more details about Booster and its capabilities.
```
javiertoledo, line 549:
```suggestion
Continue reading to dig more, you've just scratched the surface of all the Booster capabilities!
```

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+    register.events(/* YOUR EVENT HERE */)
+  }
+}
+
+```
+
+#### 3. First event
+
+It makes sense to emit an event after executing the command `CreatePost`, something like
+`PostCreated`. Let's do it by running this command:
```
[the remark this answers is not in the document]
itrion: I'll rewrite this suggestion to make it a bit simpler. I think it is not necessary to compare the approaches (table vs. event-sourced) at this point. Also, we are not updating the posts in the tutorial, so the `Later, when we edit the post...` part is not covered.

docs/README.md
```diff
@@ -121,5 +127,421 @@ something like
...
+
+> Booster's new: commands follows this structure
+>
+> - `boost` is the Booster CLI
+> - `new:<resource>` new is a CLI command, :project tells booster the kind of resource
+> - `boosted-blog` is a parameter for the `new:project` command
+> project name
+
+The `new:project` command generates some scaffolding for you. The project name will be the
+project's root so `cd` to it:
```
NickSeagull, line 177:
```suggestion
project's root so `cd` into it:
```

docs/README.md
```diff
@@ -123,3 +130,420 @@ something like
...
+
+Now open the project in your most preferred IDE, e.g. [Visual Studio Code](https://code.visualstudio.com/).
+
+#### 2. First command
+
+We will now define our first [command](#commands), it will allow us to create posts in our
+blog. Write this command in the project's root directory:
+
+```bash
+boost new:command CreatePost --fields postId:UUID title:string content:string author:string
```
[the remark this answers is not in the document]
NickSeagull: Perhaps a bank account app?

docs/README.md
```diff
@@ -121,5 +127,421 @@ something like
...
+boosted-blog
+└── src
+    └── commands
+        └── CreatePost.ts
+```
+
+We still need to define two more things:
+
+1. Who is authorized to run this command.
+1. And what events are being triggered when the command is executed.
```
NickSeagull, line 224:
```suggestion
2. And what events are being triggered when the command is executed.
```

docs/README.md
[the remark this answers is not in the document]
NickSeagull: Yep, this is "a taste" of Booster, i'd make it very simple. In the architecture section the user can read more

docs/README.md
```diff
@@ -121,5 +127,421 @@ something like
...
+}
+```
+
+```graphql
+mutation {
+  CreatePost(
+    input: {
+      postId: "05670e55-fd31-490e-b585-3a0096db0412"
+      title: "Booster framework rocks"
+      content: "I am so excited for writting the second post"
```
NickSeagull, line 433:
```suggestion
      content: "I am so excited for writing the second post"
```

docs/README.md
NickSeagull: I'd rewrite this part in your suggestion

```diff
- - Make events trigger event handlers that can run code or trigger other events
+ - Make events trigger code or other events
```

docs/README.md
itrion: For MD it doesn't matter what number you use because the parser will handle the ordered list automatically as soon as they start with `<number>.`

https://www.markdownguide.org/basic-syntax#ordered-lists

I think it is actually a good practice to just use `1.` because if you move the item you don't have to care about the new order

docs/README.md
itrion: I finally wrote: `- Make events trigger other events`

PR #161, opened 2020-06-08
BOOST-728: fix end to end tests

MYF95:
## Description
Adds AWS configuration to end-to-end tests and update scripts to include the environment variable needed for AWS `AWS_SDK_LOAD_CONFIG=true`.
Added the flag `FULL_INTEGRATION_TEST`, which makes the AWS setup config to not nuke the application stack, allowing the end-to-end tests to reuse it and nuke it when it finishes, reducing the amount of times we have to deploy/nuke the test application stack.

## Changes
- Update integration scripts in `package.json`
- Add AWS setup in end-to-end `setup.ts`

## Checks
- [X] Project Builds

### DIFF
--- a/packages/framework-integration-tests/integration/end-to-end/setup.ts
+++ b/packages/framework-integration-tests/integration/end-to-end/setup.ts
@@ -1,0 +1,43 @@
+import util = require('util')
+const exec = util.promisify(require('child_process').exec)
+import { deploy, nuke } from '../../src/deploy'
+import { config } from 'aws-sdk'
+
+// TODO: remove the AWS config with BOOST-729 to make end-to-end tests provider unaware
+
+before(async () => {
+  await setEnv()
+  await checkConfigAnd(deploy)
+})
+
+after(async () => {
+  await checkConfigAnd(nuke)
+})
+
+async function setEnv(): Promise<void> {
+  if (!process.env.BOOSTER_APP_SUFFIX) {
+    // If the user doesn't set an app name suffix, use the current git commit hash
+    // to build a unique suffix for the application name in AWS to avoid collisions
+    // between tests from different branches.
+    const { stdout } = await exec('git rev-parse HEAD')
+    process.env['BOOSTER_APP_SUFFIX'] = stdout.trim().substring(0, 6)
+    console.log('setting BOOSTER_APP_SUFFIX=' + process.env.BOOSTER_APP_SUFFIX)
+  }
+}
+
+async function checkConfigAnd(action: () => Promise<void>): Promise<void> {
+  console.log('Checking AWS configuration...')
+  if (!config.credentials?.accessKeyId || !config.credentials?.secretAccessKey) {
+    throw new Error(
+      "AWS credentials were not properly loaded by the AWS SDK and are required to run the integration tests. Check that you've set them in your `~/.aws/credentials` file or environment variables. Refer to AWS documentation for more details https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-credentials-node.html"
+    )
+  }
+  if (!config.region) {
+    throw new Error(
+      "AWS region was not properly loaded by the AWS SDK and is required to run the integration tests. Check that you've set it in your `~/.aws/config` file or AWS_REGION environment variable. Refer to AWS documentation for more details https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-region.html#setting-region-order-of-precedence"
+    )
+  } else {
+    console.log('AWS Region set to ' + config.region)
+  }
+  await action()
+}
--- a/packages/framework-integration-tests/package.json
+++ b/packages/framework-integration-tests/package.json
@@ -42,11 +42,11 @@
     "fix-lint": "eslint --quiet --fix --ext '.js,.ts' **/*.ts",
     "compile": "tsc -b tsconfig.json",
     "clean": "rimraf ./dist tsconfig.tsbuildinfo",
-    "integration": "yarn integration/aws && yarn integration/local && yarn integration/cli && yarn integration/end-to-end",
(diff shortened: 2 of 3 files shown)


glammers1:
bot: integration

glammers1:
Screenshot 2020-06-08 at 19 19 21

Tests are passing.

tainguyenbui:
bot: integration

glammers1:
bot: integration

packages/framework-integration-tests/integration/end-to-end/setup.ts
```diff
@@ -0,0 +1,43 @@
+import util = require('util')
+const exec = util.promisify(require('child_process').exec)
+import { deploy, nuke } from '../../src/deploy'
+import { config } from 'aws-sdk'
+
+// TODO: remove the AWS config with BOOST-729 to make end-to-end tests provider unaware
```
alvaroloes, line 6: 👍

javiertoledo: Can end-to-end tests actually be provider-unaware? Passing these tests in AWS doesn't mean that they'll pass in Azure...

packages/framework-integration-tests/package.json
```diff
@@ -42,11 +42,11 @@
     "fix-lint": "eslint --quiet --fix --ext '.js,.ts' **/*.ts",
     "compile": "tsc -b tsconfig.json",
     "clean": "rimraf ./dist tsconfig.tsbuildinfo",
-    "integration": "yarn integration/aws && yarn integration/local && yarn integration/cli && yarn integration/end-to-end",
-    "integration/aws": "mocha --forbid-only --config \"integration/providers/aws/.mocharc.yml\" \"integration/providers/aws/**/*.integration.ts\"",
+    "integration": "FULL_INTEGRATION_TEST=true yarn integration/aws && yarn integration/local && yarn integration/cli && yarn integration/end-to-end",
+    "integration/aws": "AWS_SDK_LOAD_CONFIG=true mocha --forbid-only --config \"integration/providers/aws/.mocharc.yml\" \"integration/providers/aws/**/*.integration.ts\"",
```
javiertoledo, line 46: If I understand this correctly, the goal for introducing this `FULL_INTEGRATION_TEST` is that you avoid nuking the AWS environment after `integration/aws` to reuse it in `integration/end-to-end`.

Some thoughts about this:
* This behavior is a bit fragile because this `package.json` file is becoming quite complex and it will quickly become a mess if we keep adding environment variables and other stuff. 
* End-to-end tests will always depend on a provider, or at least, I see no way to make them provider agnostic. 
* We're seeing the need to manage actual deployments in a more fine-grained way.

Some ideas to address these isues:

1. Create separate `integration/aws/deploy` and `integration/aws/nuke` scripts.
2. `integration/aws` don't perform deploy or nuke operations, just run the tests assuming that an application with a name that matches the current branch exists. With these two changes, we can manually manage deployments and tests as we debug them, making the process easier, and we can prepare more complex scripts.
3. Likewise, `integration/end-to-end`, which might actually be renamed as `end-to-end/aws`, doesn't perform deploy or nuke operations.
4. Consider unifying the integration and end-to-end `setup.ts` files, as all they'll do if we remove the deployment from there is setting up the environment.
5. Create a new ¿bash? script file with all the steps and conditions to run all the integration tests using these building blocks, making the `integration` script just call that file. This way we can easily manage the increasingly complex full-integration-tests procedure (which will soon include azure and kubernetes tests too)

MYF95: I think the idea of making them provider-unaware is because the end-to-end tests, at least the ones we have right now, are based on testing GraphQL requests, so the only reason they need a provider is for the endpoints URLs.

We will definitely need some better structure for other providers, and I really like the idea of having more defined folders for the providers since we will need custom configuration for each provider.

To avoid dragging the failing end-to-end tests I would rather tackle this issue in a separate ticket, but if it makes sense to work on it now, I'll start looking into it @javiertoledo

tainguyenbui: Whilst I agree with most of you have said @javiertoledo, I think that making End-to-End provider agnostic is still possible. At the end of the day, all an End-to-End test needs is an endpoint. If we make the retrieval of an endpoint and provisioning of infrastructure abstract enough so we just return an endpoint, it would become provider agnostic.

packages/framework-integration-tests/integration/end-to-end/setup.ts
alvaroloes: Oh, nono, of course, we need to pass these tests using each provider, but the tests themselves are the same (I mean the same file, the same code), we don't need to repeat them. That's why we want these tests to only use the public API of Booster.

This way, we have the following benefits:
1.- We get tests for free for all the providers
2.- The most important: we will ensure all providers behave the same way regardless the provider used 💥

packages/framework-integration-tests/package.json
alvaroloes: Linking here my response to the other comment because it is relevant regarding why provider-unaware end-to-end tests are the "bomb": https://github.com/boostercloud/booster/pull/161#discussion_r437360597

Also @javiertoledo I think missed some context about the reason of this PR:
- Check the description of this ticket: https://agilemonkeys.atlassian.net/browse/BOOST-728
- This other ticket is its continuation: https://agilemonkeys.atlassian.net/browse/BOOST-729

In brief: we wanted, right now, to just fix the end to end test as simply as possible, because they were failing in master. Then, in the next sprint, we will refactor them to leave them in much better shape. Doing this refactor in this sprint was a bit too much and the idea is to focus on _creating_ integration tests.

javiertoledo: Gotcha! I answered on the ticket. Anyway, please consider my other suggestions about how to keep this `package.json` manageable 😜

PR #184, opened 2020-06-10
Project generators integration tests

glammers1:
## Description
Add tests for project generators and checking that the generating files can be compiled.

## Changes
* CLI integration tests are added

## Checks
- [X] Project Builds

## Additional information
I've been struggling with providing stdin inputs for the CLI prompts. Since these are integration tests, I didn't want to mock the prompt operation (provided by the library called [inquirer](https://www.npmjs.com/package/inquirer)), so I finally came up with an approach that is able to write in the writable stdin. I had to use an interval to send the answers, though. I've not found any other approach that worked fine so far. I found several articles to mock the prompt operation ([this is a good one by the way](https://glebbahmutov.com/blog/unit-testing-cli-programs/)), but I was not able to find a better approach for our use case. Feel free to discuss it!

I'll open an issue on GitHub to troubleshoot this issue `// TODO remove skip when '-h' flag works fine, now it's throwing an error`.

### DIFF
--- a/packages/framework-integration-tests/integration/cli/cli.project.integration.ts
+++ b/packages/framework-integration-tests/integration/cli/cli.project.integration.ts
@@ -40,18 +40,18 @@ describe('Project', () => {
   )
 
   const sendToStdin = (childProcess: ChildProcess, promptAnswers: Array<string>, delay: number) => {
-    let currentRepetitions = 0;
+    let currentRepetitions = 0
     const totalRepetitions = promptAnswers.length
     let answers: Array<string> = promptAnswers.slice()
 
-    let intervalID = setInterval(() => {
+    const intervalID = setInterval(() => {
       childProcess.stdin?.write(answers[0])
       answers = answers.slice(1)
 
       if (++currentRepetitions === totalRepetitions) {
-        clearInterval(intervalID);
+        clearInterval(intervalID)
       }
-    }, delay);
+    }, delay)
   }
 
   const execNewProject = (projectName: string, promptAnswers: Array<string> = [], flags: Array<string> = []) => {
@@ -64,6 +64,20 @@ describe('Project', () => {
 
       if (promptAnswers.length > 0) {
         sendToStdin(childProcess, promptAnswers, 1000)
+      }
+    })
+  }
+
+  const packageJsonAssertions = (expectedJson: string, jsonContent: string, objectsToCompareJustKeys: Array<string>, checkKeysAndValues: boolean = true) => {
+    const expectedJsonObj = JSON.parse(expectedJson)
+    const jsonContentObj = JSON.parse(jsonContent)
+
+    Object.entries(expectedJsonObj).forEach(([key, value]) => {
+      if (objectsToCompareJustKeys.includes(key)) {
+        expect(jsonContentObj.hasOwnProperty(key)).true
+        return packageJsonAssertions(JSON.stringify(expectedJsonObj[key]), JSON.stringify(jsonContentObj[key]), objectsToCompareJustKeys, false)
+      } else {
+        checkKeysAndValues ? expect(jsonContentObj[key]).to.deep.equals(expectedJsonObj[key]) : expect(jsonContentObj.hasOwnProperty(key)).true
       }
     })
   }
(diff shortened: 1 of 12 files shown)


packages/framework-integration-tests/integration/cli/cli.project.integration.ts
```diff
@@ -0,0 +1,205 @@
...
+
+      if (++currentRepetitions === totalRepetitions) {
+        clearInterval(intervalID);
+      }
+    }, delay);
+  }
+
+  const execNewProject = (projectName: string, promptAnswers: Array<string> = [], flags: Array<string> = []) => {
+    return new Promise<string>((resolve): void => {
+      const childProcess = require('child_process').exec(`${cliPath} new:project ${projectName} ${flags.join(' ')}`,
```
tainguyenbui, line 59: you could simplify this line if you add the code below to the top of the file
```
import util = require('util')
const exec = util.promisify(require('child_process').exec)
```

```suggestion
      const childProcess = exec(`${cliPath} new:project ${projectName} ${flags.join(' ')}`,
```

packages/framework-integration-tests/integration/fixtures/cart-demo/package.json
```diff
@@ -0,0 +1,43 @@
+{
```
tainguyenbui, line 1: I have mixed feelings with this fixture. On one hand it its great that it checks all the dependencies and the versions. On the other hand, it will cause many integration failures since dependencies will be update often, as well as the `@boostercloud` packages on every release. Is there any way that we could compare this one using regex?

tainguyenbui:
great stuff, I left a couple of comments

packages/framework-integration-tests/integration/cli/cli.project.integration.ts
glammers1: The code already has that exec const at the top of the [file](https://github.com/boostercloud/booster/pull/184/files#diff-7aa788437077ac27a72747346874c349R8) and I used it in some tests, but this case is different. I don't want to use the `promisify` version here. Using that approach I was not able to write in `stdin` to provide answers to CLI's questions since I don't have access to the `stdin` until the Promise is resolved, but the Promise never resolves because it gets stuck waiting for a response that never comes. I got a timeout error in that scenario. That's why I create a new Promise manually here where I execute the bare `exec` and I am able to send stuff to `stdin`. Once the exec process finishes and its callback is executed, I can resolve the Promise with the stdout value. That was the tricky part I spent most of the time coding these tests.

packages/framework-integration-tests/integration/fixtures/cart-demo/package.json
glammers1: That's an excellent question. Thanks for pointing that out, I am going to think about it. I agree with you.

packages/framework-integration-tests/integration/cli/cli.project.integration.ts
tainguyenbui: makes sense 👌

packages/framework-integration-tests/integration/cli/setup.ts
```diff
@@ -29,6 +34,7 @@ before(async () => {
 
   try {
     await Promise.all(removeFiles(testFiles))
+    await Promise.all(removeFolders(testFolders))
```
javiertoledo, line 37: This code might be slightly faster (probably not really significative, but... 🤷) if you concat the two arrays and call `Promise.all` only one time:

```suggestion
    await Promise.all([
      ...removeFiles(testFiles),
      ...removeFolders(testFolders),
    ])
```

packages/framework-integration-tests/integration/fixtures/cart-demo/package.json
javiertoledo: Well, at least this test will fail if things change...

glammers1: I was thinking about solving this issue using regex but I haven't found a smooth approach to do that without maintaining a complex regex. Instead, I just created a new method where I check that the key-values from the generated package.json are the same from the expected one, but the `dependencies` and `devDependencies` section where I only check the keys and not the values. Here is the commit f3d2eea4a6e1b4a36498e3c08fd11f5a24e896ba

glammers1: Take into the account that is the generated package.json has an invalid version of a library for any reason (let's say a typo in the Booster templates), these tests will fail for that reason anyway, so I think is not worth trying to check whether or not a version meets a correct npm version pattern.

glammers1:
bot: integration

packages/framework-integration-tests/integration/fixtures/cart-demo/package.json
tainguyenbui: sounds good to me. That just made me find out that we should probably not ally * for the framework-aws-provider

glammers1:
bot: integration

glammers1:
bot: integration

PR #200, opened 2020-06-12
Fix: error when static site bucket not found

tainguyenbui:
## Description
Handles the error returned when there isn't any static site bucket to be deleted

## Changes
- Create new function in `s3utils.ts` to check whether there is a bucket available
- Adds new logic to avoid deleting an inexistent bucket 

## Checks
- [x] Project Builds


tainguyenbui:
bot: integration

PR #210, opened 2020-06-15
BOOST-467 event reduction into entities

tainguyenbui:
## Description
Adds a test to verify that a reduction in an Entity is applied due to an event

## Changes
- Refactor the createAdmin function to become more flexible with user roles
- Add test to check entity reducers

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
+++ b/packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
@@ -1,7 +1,7 @@
 import { ApolloClient } from 'apollo-client'
 import { NormalizedCacheObject } from 'apollo-cache-inmemory'
-import { graphQLClient, waitForIt } from '../providers/aws/utils'
-import { random } from 'faker'
+import { createUser, getAuthToken, graphQLClient, waitForIt } from '../providers/aws/utils'
+import { random, commerce, finance, lorem, internet } from 'faker'
 import { expect } from 'chai'
 import gql from 'graphql-tag'
 import { CartItem } from '../../src/common/cart-item'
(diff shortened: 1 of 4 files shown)


packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -162,6 +162,170 @@ describe('Cart end-to-end tests', () => {
...
   })
 
+  describe('Entities', () => {
+    let userAuthToken: string
+    let userEmail: string
+    const mockPassword = 'Enable_G0d_Mode3e!'
+
+    before(async () => {
+      userEmail = internet.email()
+      // TODO: Make retrieval of auth token cloud agnostic
```
alvaroloes, line 172: 👍

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -162,6 +162,170 @@ describe('Cart end-to-end tests', () => {
...
+                  }
+                }
+              `,
+            })
+          },
+          (result) => result?.data?.ProductReadModels?.some((product: any) => product.sku === mockSku)
+        )
+
+        const product = products.data.ProductReadModels.find((product: any) => product.sku === mockSku)
+        productId = product.id
```
alvaroloes, line 247: I guess you use the ID returned in the "expected" result just because you are doing a deep comparison (as if not, it would fail), correct?

packages/framework-integration-tests/integration/providers/aws/utils.ts
```diff
@@ -74,7 +74,7 @@ export async function userPoolPhysicalResourceId(): Promise<string> {
   }
 }
 
-export async function createAdmin(username: string, password: string): Promise<void> {
+export async function createUser(username: string, password: string, role: string = 'User'): Promise<void> {
```
alvaroloes, line 77: Good change!

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
tainguyenbui: that is correct @alvaroloes, I usually like checking it just in case 😅

tainguyenbui:
bot: integration

PR #211, opened 2020-06-15
Add the booster architecture introduction and diagram

itrion:
## Description
After some discussion with @javiertoledo, we have updated the diagram, and now it illustrates the key Booster concepts for both the data structures perspective and the functional perspective.

## Changes
Note:
> if you look at the readme file when reviewing, you'll see the old diagram rendered. This PR is also updating the diagram so the new one is this https://github.com/boostercloud/booster/blob/8621a7b5cbf9016acc105304df2b113012c58317/docs/img/booster-arch.png

With this change, the documentation is more explicit and more coherent. It is because it illustrates the role of `Reductions - Entities` and `Projections - ReadModels`, which was confusing for newcomers. Additionally, it avoids loops, making emphasis on the decoupling between reads and writes and the inversion of control around events and entities. Finally, it is more coherent because we are now using the concepts the developer sees when dealing with the `boost` CLI, e.g. `boost new:command | event | entity....` and also the code `@Entity, @Reduces, @ReadModel, @Projects, @Commands, handle()...`

## Checks
- [ ] Project Builds


docs/README.md
```diff
@@ -641,4 +641,20 @@ The are many other options for your serverless backend built with Booster Framew
 - Reading entities within command handlers to apply domain-driven decisions
 - and much more...
 
-Continue reading to dig more; you've just scratched the surface of all the Booster capabilities!
+Continue reading to dig more; you've just scratched the surface of all the Booster
+capabilities!
+
+## Booster architecture
+
+Two patterns influence the Booster's event-driven architecture: Command-Query Responsibility Segregation ([CQRS](https://www.martinfowler.com/bliki/CQRS.html)) and [Event Sourcing](https://martinfowler.com/eaaDev/EventSourcing.html). Booster shines by making them play very well together. You get the benefits of joining these patterns without dealing with their pains.
```
javiertoledo, line 649:
```suggestion
Two patterns influence the Booster's event-driven architecture: Command-Query Responsibility Segregation ([CQRS](https://www.martinfowler.com/bliki/CQRS.html)) and [Event Sourcing](https://martinfowler.com/eaaDev/EventSourcing.html). They're complex techniques to implement from scratch with lower-level frameworks, but Booster makes them feel natural and very easy to use.
```

docs/README.md
```diff
@@ -641,4 +641,20 @@ The are many other options for your serverless backend built with Booster Framew
...
+
+![architecture][booster-arch-png]
+
+The public interface of a Booster application is just `Commands` and `ReadModels`. Booster proposes an entirely different approach to the Model-View-* and CRUD frameworks. With Booster, the clients submit commands, query the read models, or subscribe to them for receiving real-time updates thanks to the out of the box [GraphQL API][booster-graphql-api].
+
+Booster applications are event-driven and event-sourced so, **the source of truth is the whole history of events**. When a client submits a command, the `CommandHandler` _wakes up_ and executes its logic. Optionally, it can *register* as many `Events` as needed. The framework caches the current state by automatically *reducing* all the registered events into `Entities`.  Interested parties can *react* to events via `EventHandlers`, and finally, the *projection* functions transform the entities into `ReadModels`.
```
glammers1, line 655:
```suggestion
Booster applications are event-driven and event-sourced so, **the source of truth is the whole history of events**. When a client submits a command, the `CommandHandler` _wakes up_ and executes its logic. Optionally, it can *register* as many `Events` as needed. The framework caches the current state by automatically *reducing* all the registered events into `Entities`. Interested parties can *react* to events via `EventHandlers`, and finally, the *projection* functions transform the entities into `ReadModels`.
```

glammers1: I just remove an extra whitespace between the dot and the word `Interested`.

PR #212, opened 2020-06-16
Integration test AWS: Entities projection into read models

glammers1:
## Description
Add an integration test that checks that a read model is correctly projecting an entity and generating the corresponding objects in the read model database **(AWS)**.

## Changes
- A new integration test was added.
- New helper methods were added to query the read model tables.

## Checks
- [X] Project Builds
~- [ ] Updated documentation accordingly~ NOT NEEDED

## Additional information

The test executes two mutations to update the same read model on purpose. I want to check that the same DynamoDB item is updated and no new rows are created.

### DIFF
--- a/packages/framework-integration-tests/integration/providers/aws/entities.integration.ts
+++ b/packages/framework-integration-tests/integration/providers/aws/entities.integration.ts
@@ -1,0 +1,78 @@
+import { countReadModelItems, graphQLClient, queryReadModels, waitForIt } from './utils'
+import { expect } from 'chai'
+import gql from 'graphql-tag'
+import { ApolloClient } from 'apollo-client'
+import { NormalizedCacheObject } from 'apollo-cache-inmemory'
+import { random } from 'faker'
+
+const CART_READ_MODEL_NAME = 'CartReadModel'
+
+describe('entities', async () => {
+  let client: ApolloClient<NormalizedCacheObject>
+
+  before(async () => {
+    client = await graphQLClient()
+  })
+
+  it('should be projected into a read model', async () => {
+    const readModelItemsCount = await countReadModelItems(CART_READ_MODEL_NAME)
+
+    const commandsPromises: Promise<any>[] = []
+
+    const mockCartId = random.uuid()
+    const mockProductId = random.uuid()
+    const mockQuantity = random.number({ min: 1 })
+    const mockPaymentId: string = random.uuid()
+    const mockConfirmationToken: string = random.alphaNumeric(10)
+
+    commandsPromises.push(
+      client.mutate({
+        variables: {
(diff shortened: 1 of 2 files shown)


packages/framework-integration-tests/integration/providers/aws/entities.integration.ts
```diff
@@ -0,0 +1,78 @@
...
+            ConfirmPayment(input: { paymentId: $paymentId, cartId: $cartId, confirmationToken: $confirmationToken })
+          }
+        `,
+      })
+    )
+
+    const commandsPromisesResult = await Promise.all(commandsPromises)
+
+    expect(commandsPromisesResult[0]).not.to.be.null
+    expect(commandsPromisesResult[0].data?.ChangeCartItem).to.be.true
```
tainguyenbui, line 58: will this expect already satisfy the one in line 57?

packages/framework-integration-tests/integration/providers/aws/entities.integration.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const expectedReadModelItemsCount = readModelItemsCount + 1
+    await waitForIt(
+      () => countReadModelItems(CART_READ_MODEL_NAME),
+      (newReadModelItemsCount) => newReadModelItemsCount === expectedReadModelItemsCount
+    )
+
+    const latestReadModelItem = await queryReadModels(mockCartId, CART_READ_MODEL_NAME)
+    expect(latestReadModelItem).not.to.be.null
+    expect(latestReadModelItem[0].id).to.be.equal(mockCartId)
```
tainguyenbui, line 70: I guess you could also perform a `to.be.deep.equal` and compare an object instead

packages/framework-integration-tests/integration/providers/aws/utils.ts
```diff
@@ -248,6 +248,38 @@ export async function queryEvents(primaryKey: string, latestFirst = true): Promi
...
+  const stackName = appStackName()
+
+  return `${stackName}-${readModelName}`
+}
+
+export async function queryReadModels(primaryKey: string, readModelName: string, latestFirst = true): Promise<any> {
+  const output: QueryOutput = await documentClient
+    .query({
+      TableName: await readModelTableName(readModelName),
+      KeyConditionExpression: 'id = :v',
```
tainguyenbui, line 263: i would give `:v` a more meaningful name. I understand that it means `value` but since this expression will be always targeting the primaryKey, what about having something like `:primaryKey`?

tainguyenbui:
bot: integration

tainguyenbui:
@glammers1 merging this one, as it got 2 approvals plus integration tests pass

PR #215, opened 2020-06-18
Add headings for the booster architecture section

itrion:
## Description
This PR adds the headings for the Booster architecture sections. It also inlines a reference because using references vs using links is something that may need an agreement as references are a bit harder to handle.

## Checks
- [x] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -657,9 +657,37 @@ Booster applications are event-driven and event-sourced so, **the source of trut
 In this chapter you'll walk through these concepts and its details.
 
 [booster-graphql-api]:#graphql-api
-[booster-arch-png]:./img/booster-arch.png
+[booster-arch-png**:./img/booster-arch.png
 
-### Event handlers
+### 1. Command and command handlers
+
+#### Commands naming convention
+
+#### Creating a command
+
+#### The command handler function
+
+##### Validating data
+
+##### Reading entities
+
+##### Registering events
+
+#### Authorizing a command
+
+#### Submitting a command
+
+### 2. Events
+
+#### Events naming convention
+
+#### Creating events
+
+#### The event store
+
+#### Events ordering
+
+### 3. Event handlers
 In event-driven architectures we have different parts of our application that react to events, one of them is the `@Entity`, in charge of reducing the event. But we also have event handlers, a class with the `@EventHandler` decorator. The event handlers also react to events, and are used when you want to trigger new events based on the original one.
 
 An event handler would look like this:
@@ -725,3 +753,30 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
   ...
 }
 ```
+
+### 4. Entities and reducers
+
+#### Entities naming convention
+
+#### Creating entities
+
+#### The reducer function
+
+#### Aggregate data using entities
+
+#### Eventual consistency
+
+### 5. Read models and projections
+
+#### Read models naming convention
+
+#### Creating a read model
+
+#### The projection function
+
+#### Authorizing read models
+
+#### Querying a read model
+
+#### Getting real-time updates for a read model
+
--- a/docs/README.md
+++ b/docs/README.md
@@ -648,16 +648,13 @@ capabilities!
 
 Two patterns influence the Booster's event-driven architecture: Command-Query Responsibility Segregation ([CQRS](https://www.martinfowler.com/bliki/CQRS.html)) and [Event Sourcing](https://martinfowler.com/eaaDev/EventSourcing.html). They're complex techniques to implement from scratch with lower-level frameworks, but Booster makes them feel natural and very easy to use.
 
-![architecture][booster-arch-png]
+![architecture](./img/booster-arch.png)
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -648,18 +648,43 @@ capabilities!
...
 Booster applications are event-driven and event-sourced so, **the source of truth is the whole history of events**. When a client submits a command, the `CommandHandler` _wakes up_ and executes its logic. Optionally, it can *register* as many `Events` as needed. The framework caches the current state by automatically *reducing* all the registered events into `Entities`. Interested parties can *react* to events via `EventHandlers`, and finally, the *projection* functions transform the entities into `ReadModels`.
 
 In this chapter you'll walk through these concepts and its details.
 
-[booster-graphql-api]:#graphql-api
-[booster-arch-png]:./img/booster-arch.png
+### 1. Command and command handlers
 
-### Event handlers
+#### Commands naming convention
```
tainguyenbui, line 661: @javiertoledo commented on my PR that maybe setting naming convention is not necessary

docs/README.md
```diff
@@ -725,3 +750,30 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+
+#### Creating entities
+
+#### The reducer function
+
+#### Aggregate data using entities
+
+#### Eventual consistency
+
+### 5. Read models and projections
```
tainguyenbui, line 766: will we talk here about how we could project several entities into a read model?

itrion: I think Javi said that aggregating in read-models isn't the right approach, and instead, we should have `Aggregated Entities` and then read model for those. 

@javiertoledo, Is that correct?

With that in mind, I added this heading [`Aggregate data using entities`](https://github.com/boostercloud/booster/pull/215/files#diff-e70100059cc3a7376740bbe9751a082eR762) under the entities section.

We can advise the users three things:

1. To aggregate data only in entities
1. do it only in read models
1. or, aggregate data in both

With the proposed approach, we go with the first choice

docs/README.md
itrion: I think we should propose one way of doing things and then let people use the framework however they want, or they think it makes sense.

@tainguyenbui @javiertoledo, think about [shuhari](https://en.wikipedia.org/wiki/Shuhari), the Japanese martial art concept which describes the stages of learning to mastery.

- **shu** -> you obey what you have been told because you have no idea of anything
- ha -> you digress the tradition and breaks the rules because you have enough understanding
- ri -> you are already a master, so you don't need rules and tradition because you create them

At this point, we are addressing people in the first stage (everyone, except some of us, is in this stage because Booster is something new for most people). They might need our guidance on how to do it well, and then they can start using the framework breaking the rules.

Maybe we don't need a naming convention for every section, but in commands, it makes sense because the tradition up to this moment is to create five methods in a controller class, and with Booster you have tons of choices.

javiertoledo: For commands, we could say that we put them in the `commands` folder and use names that are imperative verbs + the object of that verb like `CreateOrder` or `AddItemToCart`. I didn't like the name convention proposed for event handlers because they're already in a folder named `event-handlers` and annotated with the `@EventHandler` decorator, so adding the prefix `Handle` seemed redundant to me. The problem with event handlers is that we initially designed them in Haskell, where you can define many functions with the same name that then you can match by the actual parameter type:

```haskell
handle :: SomeEvent -> [SomeOtherEvent]
handle event = ...
```

and other handler as:
```haskell
handle :: YetAnotherEvent -> [EvenAnotherEvent]
handle event = ...
```

With this design, event handlers don't need a wrapping class or even a name, they're just transformations when an event of kind A appears, I'll emit a list of events of kind B. You don't even need to define which handler you use because the language does the match for you.

With TypeScript, we're forced to wrap handlers in a class to use decorators, and classes require names. I don't know if we can come up with a convention that makes sense for all situations, so I'd let the developer decide. Maybe I'm wrong and it makes a lot of sense to prefix all handlers with `Handle*`, as @tainguyenbui suggested, but I'm not 100% sure.

docs/README.md
javiertoledo: There's not a firm decision on how we're going to design access from handlers to read models or if that makes sense at all. There's some discussion about this topic in [an internal Confluence page](https://agilemonkeys.atlassian.net/wiki/spaces/B/pages/1199341571/Fixing+Entity+ReadModel+and+GraphQL+overlap?focusedCommentId=1207173121#comment-1207173121)

Basically, Read Models are designed to optimize reads from the API side, but they're all public by default, so there are some use cases where we want to query the state from a handler but we can't publish the queried data as a read model in the public API. On the other side, it's important that we store the current state in a database that we can query to answer questions that require obtaining a list of entities and things like that. There are many alternatives:

1. Read models can be private, and we let handlers to read any read model
2. Read models are only accessible by the API and we implicitly store all Entities in tables and create a querying mechanism for entity tables (I bet we can reuse the current mechanism for Read Models). The good thing about this is that "it just works", bad thing is that we generate a lot of unnecessary lambda activity.
3. Read models are only for the API, but we add a cached=true parameter to the @Entity decorator that makes Booster generate a table for this entity and keep it up to date. Cached entities can be queried.

We have to evaluate pros and cons of each option to make sure that we pick a solution that makes sense, don't introduce weird extra config parameters to the developer and is cost-efficient

itrion: Hopefully, we will have that clear when we reach that point in the docs, if not we can postpone the decision and document it later

PR #220, opened 2020-06-19
Booster architecture commands

itrion:
## Description
This PR adds some parts of the documentation for the `Command and command handlers` section.

I'll open it as a draft because it is not complete but it can be reviewed/discussed. Next week I'll be on vacations so if some of you can help with the missing parts that would be awesome:

## Checks
- [ ] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -887,7 +887,7 @@ Booster generates this GraphQL mutation:
 mutation CreateProduct($input: CreateProductInput!): Boolean
 ```
 
-where `CreateProductInput` is 
+where the schema for `CreateProductInput` is 
 
 ```text
 {
--- a/docs/README.md
+++ b/docs/README.md
@@ -722,7 +722,7 @@ Note:
 
 #### The command handler function
 
-Booster generates the command handler function as a method of the command class. That function is called by the framework every time its command is submitted. It is the place to run validations, return errors, query entities to make decisions, and register relevant domain events.
+Booster generates the command handler function as a method of the command class. That function is called by the framework every time that one instance of this command is submitted. This is the right place to run validations, return errors, query entities to make decisions and register relevant domain events.
 
 ##### Validating data
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -767,19 +767,13 @@ mutation($input: CreateProductInput!) {
 }
 ```
 
-You'll get this response:
+You'll get something like this response:
 
 ```grapqhl
 {
   "errors": [
     {
       "message": "price must be below 10, and it was 19.99",
-      "locations": [
-        {
-          "line": 5,
-          "column": 3
-        }
-      ],
       "path": [
         "CreateProduct"
       ]
--- a/docs/README.md
+++ b/docs/README.md
@@ -667,9 +667,11 @@ There is an architectural splitting between commands and command handlers though
   authorize: 'all' | Array<RoleClass>
 })
 export class CommandName {
-  public constructor(readonly fieldA: SomeType, 
-                     readonly fieldB: SomeOtherType,
-                     /* as many fields as needed */) {}
+  public constructor(
+    readonly fieldA: SomeType,
+    readonly fieldB: SomeOtherType,
+    /* as many fields as needed */
+  ) {}
 
   public async handle(register: Register): Promise<void> {
     // Validate inputs
--- a/docs/README.md
+++ b/docs/README.md
@@ -700,7 +700,7 @@ Despite you can place commands, and other Booster files, in any directory, we st
 ```text
 project-root
 ├── src
-│   ├── commands <------ They must be here
+│   ├── commands <------ They should be here
 │   ├── common
 │   ├── config
 │   ├── entities
--- a/docs/README.md
+++ b/docs/README.md
@@ -715,7 +715,7 @@ The preferred way to create a command is by using the generator, e.g.
 boost new:command CreateProduct --fields sku:SKU displayName:string description:string price:Money
 ```
 
-The generator will automatically create a file called `CreateProduct.ts` with a TypeScript class of the same name under the commands directory. You can still create the command manually. Since the generator is not doing any *magic*, all you need is a class decorated as `@Command`. Anyway, we recommend you to use the generator as much as you can because it handles the boilerplate code for you.
+The generator will automatically create a file called `CreateProduct.ts` with a TypeScript class of the same name under the commands directory. You can still create the command manually. Since the generator is not doing any *magic*, all you need is a class decorated as `@Command`. Anyway, we recommend you always to use the generator, because it handles the boilerplate code for you.
 
 Note:
 > Running the command generator with a `CommandName` already existing, will override the content of the current one. Soon, we will display a warning before overwriting anything. Meantime, if you missed a field, just add it to the class because in Booster, all the infrastructure and data structures are inferred from your code.
--- a/docs/README.md
+++ b/docs/README.md
@@ -683,7 +683,7 @@ Every time you submit a command through the GraphQL API, Booster calls the comma
 
 #### Commands naming convention
 
-Semantic is very important in Booster as it will play an essential role in designing a coherent system. Your application should reflect your domain concepts, and commands are not an exception. Although you can name commands whatever you want, we strongly recommend you to name them starting with verbs in imperative plus the object being affected. If we were designing an e-commerce application, some commands would be:
+Semantic is very important in Booster as it will play an essential role in designing a coherent system. Your application should reflect your domain concepts, and commands are not an exception. Although you can name commands in any way you want, we strongly recommend you to name them starting with verbs in imperative plus the object being affected. If we were designing an e-commerce application, some commands would be:
 
 - CreateProduct
 - DeleteProduct
@@ -945,4 +945,3 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
 #### Querying a read model
 
 #### Getting real-time updates for a read model
-
--- a/docs/README.md
+++ b/docs/README.md
@@ -851,7 +851,7 @@ export class MoveStock {
 
 ##### Registering events
 
-After a command handler execution, it is possible to register domain events. The command handler function receives the `register` argument, so within the handler, it is possible to call `register.events(...)` with a list of events. For more details about events and the register, see the [`Events`](#2-events) section.
+Within the command handler execution, it is possible to register domain events. The command handler function receives the `register` argument, so within the handler, it is possible to call `register.events(...)` with a list of events. For more details about events and the register parameter, see the [`Events`](#2-events) section.
 
 #### Authorizing a command
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -726,7 +726,7 @@ Booster generates the command handler function as a method of the command class.
 
 ##### Validating data
 
-Booster uses the typed nature of GraphQL for ensuring that types are correct before reaching the handler, so you don't have to validate types. There are still business rules to be checked before proceeding with a command. For example, a given number must be between a threshold or a string must match a regular expression. In that case, it is enough just to throw an error in the handler, and then Booster will use the error's message as the response to make it descriptive, e.g.
+Booster uses the typed nature of GraphQL to make sure that types are correct before reaching the handler, so you don't have to validate types. There are still business rules to be checked before proceeding with a command. For example, a given number must be between a threshold or a string must match a regular expression. In that case, it is enough just to throw an error in the handler, and then Booster will use the error's message as the response to make it descriptive, e.g.
 
 Given this command:
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -658,9 +658,9 @@ In this chapter you'll walk through these concepts and its details.
 
 ### 1. Command and command handlers
 
-Booster is different than MVC frameworks in which you typically implement controller classes with the five CRUD methods per model. Instead of that, you define the user actions when interacting with an application, this approach fits very well with Domain-Driven Design. Depending on your application's domain, some examples of commands would be: `RemoveItemFromCart`, `RatePhoto`, `AddCommentToPost`, etc. Although, you can still have `Create*`, `Delete*`, or `Update*` commands when they make sense.
+Booster is different than MVC frameworks in which you typically implement controller classes with the five CRUD methods per model. Instead of that, you define the user actions when interacting with an application. This approach fits very well with Domain-Driven Design. Depending on your application's domain, some examples of commands would be: `RemoveItemFromCart`, `RatePhoto`, `AddCommentToPost`, etc. Although, you can still have `Create*`, `Delete*`, or `Update*` commands when they make sense.
 
-There is an splitting between commands and command handlers though they *live* under the same file. The command is the class with the `@Command` decorator, and the generated handle function is the command handler. That is because Booster adopts several concepts from functional programming, the separation between data structures and data transformations is one of them. In Booster a command looks like this:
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -658,22 +658,191 @@ In this chapter you'll walk through these concepts and its details.
...
+    // register.events([event1,...])
+  }
+}
+```
+
+Every time you submit a command through the GraphQL API, Booster calls the command handler function for the given command. The Commands are part of the public API so that you can define authorization policies for them. They are also the place for validating input data before registering events into the event store because they are immutable once there.
+
 #### Commands naming convention
 
+Semantic is very important in Booster as it will play an essential role in designing a coherent system. Your application should reflect your domain concepts, and commands are not an exception. Although you can name commands whatever you want, we strongly recommend you to name them starting with verbs in imperative plus the object being affected. If we were designing an e-commerce application, some commands would be:
```
javiertoledo, line 686:
```suggestion
Semantic is very important in Booster as it will play an essential role in designing a coherent system. Your application should reflect your domain concepts, and commands are not an exception. Although you can name commands in any way you want, we strongly recommend you to name them starting with verbs in imperative plus the object being affected. If we were designing an e-commerce application, some commands would be:
```

docs/README.md
```diff
@@ -658,22 +658,191 @@ In this chapter you'll walk through these concepts and its details.
...
+
 #### Creating a command
 
+The preferred way to create a command is by using the generator, e.g.
+
+```shell
+boost new:command CreateProduct --fields sku:SKU displayName:string description:string price:Money
+```
+
+The generator will automatically create a file called `CreateProduct.ts` with a TypeScript class of the same name under the commands directory. You can still create the command manually. Since the generator is not doing any *magic*, all you need is a class decorated as `@Command`. Anyway, we recommend you to use the generator as much as you can because it handles the boilerplate code for you.
```
javiertoledo, line 718:
```suggestion
The generator will automatically create a file called `CreateProduct.ts` with a TypeScript class of the same name under the commands directory. You can still create the command manually. Since the generator is not doing any *magic*, all you need is a class decorated as `@Command`. Anyway, we recommend you always to use the generator, because it handles the boilerplate code for you.
```

docs/README.md
```diff
@@ -658,22 +658,191 @@ In this chapter you'll walk through these concepts and its details.
...
+
+Note:
+> Running the command generator with a `CommandName` already existing, will override the content of the current one. Soon, we will display a warning before overwriting anything. Meantime, if you missed a field, just add it to the class because in Booster, all the infrastructure and data structures are inferred from your code.
```
javiertoledo, line 721:
```suggestion
> Running the command generator with a `CommandName` already existing, will overwrite the content of the current one. In future releases, we will display a warning before overwriting anything. Meantime, if you missed a field, just add it to the class because, in Booster, all the infrastructure and data structures are inferred from your code.
```

docs/README.md
```diff
@@ -658,22 +658,191 @@ In this chapter you'll walk through these concepts and its details.
...
+
 #### The command handler function
 
+Booster generates the command handler function as a method of the command class. That function is called by the framework every time its command is submitted. It is the place to run validations, return errors, query entities to make decisions, and register relevant domain events.
```
javiertoledo, line 725:
```suggestion
Booster generates the command handler function as a method of the command class. That function is called by the framework every time that one instance of this command is submitted. This is the right place to run validations, return errors, query entities to make decisions and register relevant domain events.
```

docs/README.md
```diff
@@ -658,22 +658,191 @@ In this chapter you'll walk through these concepts and its details.
...
+
 ##### Validating data
 
+Booster uses the typed nature of GraphQL for ensuring that types are correct before reaching the handler, so you don't have to validate types. There are still business rules to be checked before proceeding with a command. For example, a given number must be between a threshold or a string must match a regular expression. In that case, it is enough just to throw an error in the handler, and then Booster will use the error's message as the response to make it descriptive, e.g.
```
javiertoledo, line 729:
```suggestion
Booster uses the typed nature of GraphQL to make sure that types are correct before reaching the handler, so you don't have to validate types. There are still business rules to be checked before proceeding with a command. For example, a given number must be between a threshold or a string must match a regular expression. In that case, it is enough just to throw an error in the handler, and then Booster will use the error's message as the response to make it descriptive, e.g.
```

docs/README.md
```diff
@@ -658,22 +658,191 @@ In this chapter you'll walk through these concepts and its details.
...
+```grapqhl
+{
+  "errors": [
+    {
+      "message": "price must be below 10, and it was 19.99",
+      "locations": [
+        {
+          "line": 5,
+          "column": 3
+        }
```
javiertoledo, line 781: It's a bit weird to send this location info back to the client without context

docs/README.md
```diff
@@ -658,22 +658,191 @@ In this chapter you'll walk through these concepts and its details.
...
+      ],
+      "path": [
+        "CreateProduct"
+      ]
+    }
+  ]
+}
+```
+
+There could be situations in which you want to register an event representing an error. (TODO: Example here, I've seen the example of the error event in the integration tests, but I'm not 100% it works here. I'm thinking in a stock error related to an item that should not be there initially)
```
javiertoledo, line 791: Should we put this TODO here? I think it'll be rendered in the docs... You can use HTML comments to avoid showing it (or create an issue):

```suggestion
There could be situations in which you want to register an event representing an error. <!-- TODO: Example here, I've seen the example of the error event in the integration tests, but I'm not 100% it works here. I'm thinking in a stock error related to an item that should not be there initially -->
```

docs/README.md
itrion: Than you @javiertoledo, the `line` and `column` here are relative to the mutation itself, not Booster code. But yes, I agree with you that it makes no sense to write that. It depends on how the GraphQL client pretty prints the queries so I'll remove the location from the example

docs/README.md
itrion: No, this PR is actually a draft because of those TODOs, I'll continue working on this.

docs/README.md
javiertoledo: Yes, I mean, if you copied that from a real request, the @boostercloud/booster-core team would want to revisit it to build a nicer interface or at least create an issue to solve it in the future 😜

docs/README.md
javiertoledo: Oh, sorry, I didn't realize it was a draft until now :trollface:

docs/README.md
```diff
@@ -658,22 +658,244 @@ In this chapter you'll walk through these concepts and its details.
...
+  "errors": [
+    {
+      "message": "price must be below 10, and it was 19.99",
+      "path": [
+        "CreateProduct"
+      ]
+    }
+  ]
+}
+```
```
NickSeagull, line 787: That moment when you didn't know about this and yet you coded part of the exception handling 😂

docs/README.md
```diff
@@ -658,22 +658,244 @@ In this chapter you'll walk through these concepts and its details.
...
+    if (!this.enoughStock(this.productID, this.origin, this.quantity)) {
+      register.events(new ErrorEvent(`There is not enough stock for ${this.productID} at ${this.origin}`))
+    } else {
+      register.events(new StockMoved(/*...*/))
+    }
+  }
+
+  private enoughStock(productID: string, origin: string, quantity: number): boolean {
+    /* ... */
+  }
```
NickSeagull, line 815: I know that this is just an example, but probably the user will want to write a `fetchEntitySnapshot` call here, which returns a promise.

Therefore the call on the `handle` method should be `await`ed.

Feel free to ignore this 😂

docs/README.md
```diff
@@ -658,22 +658,244 @@ In this chapter you'll walk through these concepts and its details.
...
+
+  public async handle(register: Register): Promise<void> {
+    const stock = await Booster.fetchEntitySnapshot(Stock, this.productID)
+    if (!this.enoughStock(this.origin, this.quantity, stock)) {
+      register.events(new ErrorEvent(`There is not enough stock for ${this.productID} at ${this.origin}`))
+    }
+  }
+
+  private enoughStock(origin: string, quantity: number, stock?: Stock): boolean {
+    return stock !== undefined && stock.countByLocation[origin] >= quantity
```
NickSeagull, line 845: `countByLocation[origin]` could also be `undefined`, perhaps something like this?

```suggestion
  const count = stock?.countByLocation[origin]
  if (count) {
    return count >= quantity
  }
  return false
```

docs/README.md
```diff
@@ -658,22 +658,244 @@ In this chapter you'll walk through these concepts and its details.
...
+  }
+}
+```
+
 ##### Registering events
 
+After a command handler execution, it is possible to register domain events. The command handler function receives the `register` argument, so within the handler, it is possible to call `register.events(...)` with a list of events. For more details about events and the register, see the [`Events`](#2-events) section.
```
NickSeagull, line 852:
```suggestion
After a command handler execution, it is possible to register domain events. The command handler function receives the `register` argument, so within the handler, it is possible to call `register.events(...)` with a list of events. For more details about events and the register parameter, see the [`Events`](#2-events) section.
```

docs/README.md
```diff
@@ -658,22 +658,244 @@ In this chapter you'll walk through these concepts and its details.
...
+}
+```
+
+Booster generates this GraphQL mutation:
+
+```text
+mutation CreateProduct($input: CreateProductInput!): Boolean
+```
+
+where `CreateProductInput` is
```
NickSeagull, line 888:
```suggestion
where the schema for `CreateProductInput` is 
```

docs/README.md
```diff
@@ -658,22 +658,244 @@ In this chapter you'll walk through these concepts and its details.
...
+```
+
+The generator will automatically create a file called `CreateProduct.ts` with a TypeScript class of the same name under the commands directory. You can still create the command manually. Since the generator is not doing any *magic*, all you need is a class decorated as `@Command`. Anyway, we recommend you always to use the generator, because it handles the boilerplate code for you.
+
+Note:
+> Running the command generator with a `CommandName` already existing, will override the content of the current one. Soon, we will display a warning before overwriting anything. Meantime, if you missed a field, just add it to the class because in Booster, all the infrastructure and data structures are inferred from your code.
+
 #### The command handler function
 
+Booster generates the command handler function as a method of the command class. That function is called by the framework every time that one instance of this command is submitted. This is the right place to run validations, return errors, query entities to make decisions and register relevant domain events.
```
tainguyenbui, line 725:
```suggestion
Booster generates the command handler function as a method of the command class. This function is called by the framework every time that one instance of this command is submitted. The command handler is the right place to run validations, return errors, query entities to make decisions and register relevant domain events.
```

docs/README.md
tainguyenbui: to the Nick's suggestion, I would add `count > 1` because if count was -1 the condition would still be true

tainguyenbui: or event something like

`const hasStock = stock?.countByLocation[origin] > 0`

docs/README.md
```diff
@@ -658,22 +658,244 @@ In this chapter you'll walk through these concepts and its details.
 
 ### 1. Command and command handlers
 
+Booster is different than MVC frameworks in which you typically implement controller classes with the five CRUD methods per model. Instead of that, you define commands, which are the user actions when interacting with an application. This approach fits very well with Domain-Driven Design. Depending on your application's domain, some examples of commands would be: `RemoveItemFromCart`, `RatePhoto`, `AddCommentToPost`, etc. Although, you can still have `Create*`, `Delete*`, or `Update*` commands when they make sense.
```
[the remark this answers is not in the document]
itrion: @NickSeagull I think the sentence is correct, we could also say `Depending on the domain of your application, some examples...` but I do prefer that's longer

docs/README.md
```diff
@@ -658,22 +658,244 @@ In this chapter you'll walk through these concepts and its details.
...
+There is an architectural splitting between commands and command handlers though they *live* under the same file. The command is the class with the `@Command` decorator, and the generated method called `handle` is the command handler. That is because Booster adopts several concepts from functional programming; the separation between data structures and data transformations is one of them. In Booster a command looks like this:
+
+```typescript
+@Command({
+  authorize: 'all' | Array<RoleClass>
+})
+export class CommandName {
+  public constructor(readonly fieldA: SomeType, 
+                     readonly fieldB: SomeOtherType,
+                     /* as many fields as needed */) {}
```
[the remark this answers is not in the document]
itrion: Sure!!

docs/README.md
```diff
@@ -658,22 +658,244 @@ In this chapter you'll walk through these concepts and its details.
...
+
+  public async handle(register: Register): Promise<void> {
+    // Validate inputs
+    // Run domain logic
+    // register.events([event1,...])
+  }
+}
+```
+
+Every time you submit a command through the GraphQL API, Booster calls the command handler function for the given command. The Commands are part of the public API so that you can define authorization policies for them. They are also the place for validating input data before registering events into the event store because they are immutable once there.
```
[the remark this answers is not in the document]
itrion: Thank you, my friend, may I ask what the part you liked the most is?

docs/README.md
```diff
@@ -658,22 +658,244 @@ In this chapter you'll walk through these concepts and its details.
...
+
+- CreateProduct
+- DeleteProduct
+- UpdateProduct
+- ChangeCartItems
+- ConfirmPayment
+- MoveStock
+- UpdateCartShippingAddress
+
+Additionally, commands must be within the commands directory of the project source: `project-root/src/commands`.
```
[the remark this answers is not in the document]
itrion: Changed, let me know your thoughts now

docs/README.md
itrion: Sure @NickSeagull @tainguyenbui thank you for the feedback, we can use the `!!` idiom to keep the snippet short

```typescript
  private enoughStock(origin: string, quantity: number, stock?: Stock): boolean {
    const count = stock?.countByLocation[origin]
    return !!count && count >= quantity
  }
```

Take into account how we calculate the stock is not that important here as how we fetch and use entities within a command handler. It is important to pay attention to these details but it is also to save extra efforts to our readers.

Let me know your thoughts 👍

docs/README.md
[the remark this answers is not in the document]
NickSeagull: Nevermind, got confused by the use of the apostrophe

docs/README.md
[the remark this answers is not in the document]
NickSeagull: I like how you walk the user from the entry point (GraphQL) to the use case validation/registration

docs/README.md
NickSeagull: What does `!!` do? Is there a documentation page for it?

docs/README.md
```diff
@@ -658,22 +658,246 @@ In this chapter you'll walk through these concepts and its details.
...
+- ConfirmPayment
+- MoveStock
+- UpdateCartShippingAddress
+
+Despite you can place commands, and other Booster files, in any directory, we strongly recommend you to put them here `project-root/src/commands`. Having all the commands in one place will help you to understand your application's capabilities at a glance.
+
+```text
+project-root
+├── src
+│   ├── commands <------ They must be here
```
NickSeagull, line 703: Given that we've changed the text, we should probably change it here

```suggestion
│   ├── commands <------ They should be here
```

docs/README.md
itrion: LOL :) that's a typical ruby pattern to make a boolean expression from something falsy or truly, and it seems it is being used in TypeScript too https://stackoverflow.com/a/49792111

PR #225, opened 2020-06-26
Booster architecture entities

NickSeagull:


docs/README.md
```diff
@@ -753,16 +753,151 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+Instead of polluting the `Profile` entity with the events of the `Cart`, you create a third entity: `ProfileFavorites`.
+
+Then create two event handlers:
+
+* One that's listening for `ProfileCreated` and emits a `FavoritesCreated` event
+* Another one that listens for `OrderPlaced` and emits a `NewItemsPurchased`
+
+Then, the reducers of the `ProfileFavorites` entity will be subscribed to
+these newly emitted events. Aggregating the required data into a single entity.
+
```
NickSeagull, line 890: I'm absolutely unsure about this section, would need help to rephrase it better

docs/README.md
```diff
@@ -753,16 +753,151 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+Note:
+
+> Running the entity generator with an EntityName that already exists, it will override the content of the current one. In future releases, we will display a warning before overwriting anything. Meantime, if you missed a field, just add it to the class because in Booster, all the infrastructure and data structures are inferred from your code.
+
 #### The reducer function
 
+Booster generates the reducer function as a static method of the entity class. That function is called by the framework every time that an event of the
+specified type is emitted. **This function must be pure**, which means that no side effects should happen.
+
```
itrion, line 832: @NickSeagull, could you please, define what a side effect is? Or rephrase this to say what to avoid explicitly?

docs/README.md
```diff
@@ -753,16 +753,151 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+The reducer has two parameters by default:
+
+* `event` - The event object that has triggered this reducer
+* `currentEntity?` - A possible previous entity generated from another call to a reducer. Note that **this parameter is optional**, and can be `undefined` if this reducer call is the first one, like when you just deployed your app.
```
itrion, line 836:
```suggestion
* `currentEntity?` - A possible entity generated from a previous call to the reducer. **This parameter is optional**, and will be `undefined` when an event occurred for the very first time.
```

docs/README.md
```diff
@@ -753,16 +753,151 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+    event: ProductRemoved,
+    currentCart?: Cart
+  ): Cart {
+    const newItems = removeFromCart(event.item, currentCart)
+    return new Cart(event.cartID, newItems)
+  }
+}
+```
+
+We can visualize reduction like this:
```
itrion, line 870:
```suggestion
You can visualize reduction like this:
```

docs/README.md
```diff
@@ -753,16 +753,151 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
 #### Aggregate data using entities
 
+You can create entities to aggregate data that would belong to different
+entities.
+
+As an example, let's say you have already an app that has a `Cart` entity and a `Profile` entity and you want to add a new feature to have a list of the users favorite items.
+
+Instead of polluting the `Profile` entity with the events of the `Cart`, you create a third entity: `ProfileFavorites`.
+
+Then create two event handlers:
```
itrion, line 883:
```suggestion
Then create two (event handlers)[#3-event-handlers]:
```

docs/README.md
itrion: I think we can reshape the example to make it clearer. I get the point but I have two concerns:

- I'm not sure the proposed approach is cleaner than polluting the Profile entity with Cart events. 
  Does it makes sense to react to `ProfileCreated` and `OrderPlaced` directly in `ProfileFavorites`?

- Why are the purchased items the favorite ones? 
  for me that relationship is not clear, wouldn't it be better to talk about a `PurchasesHistory` entity?

NickSeagull: Yep, we can make this easier to see with your second point.

Regarding the first one, I think Booster has a limitation: different entities cannot subscribe to the same events. But I have yet to confirm this, as I'm not sure.

docs/README.md
NickSeagull: Done https://github.com/boostercloud/booster/pull/225/commits/4d592402bd59261283a79378ed0f969ee8710b79 thanks!

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
 
 ### 4. Entities and reducers
 
+The source of truth of your Booster app are the events, but events make sense in the context of a domain entity.
+For example, in a banking app, there might be two events: `MoneyDeposited` and `MoneyWithdrawn`. But these events
+only make sense in the context of a `BankAccount`.
+
+Entities are created on the fly, by _reducing_ the whole event stream. They aren't saved anywhere, although Booster
```
javiertoledo, line 761:
```suggestion
Entities are created on the fly, by _reducing_ the whole event stream. You shouldn't assume that they are stored anywhere, but Booster
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+does create automatic snapshots in order to make the reduction process efficient. You are the one that defines the
```
javiertoledo, line 762:
```suggestion
does create automatic snapshots to make the reduction process efficient. You are the one in charge of writing the
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+reducer function.
+
+An entity is defined as a class with the `@Entity` decorator. Inside of it there is a static method with
```
javiertoledo, line 765:
```suggestion
An entity is defined as a class with the `@Entity` decorator. Inside of it, you can write one or more static methods with
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+the `@Reduces` decorator that also specifies the event the reducer is subscribed to. In Booster an entity looks like this:
```
javiertoledo, line 766:
```suggestion
the `@Reduces` decorator that defines the name of the event class that the reducer is subscribed to. The reducer method is called once every time that one event of that type is registered in the event store, and Booster expects you to return how the entity ends up after the event has been applied. An entity class looks like this:
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+  public constructor(readonly fieldA: SomeType, readonly fieldB: SomeOtherType /* as many fields as needed */) {}
+
+  @Reduces(SomeEvent)
+  public static reduceSomeEvent(event: SomeEvent, previousState?: EntityName): EntityName {
+    /* Return a new entity based on the previous one */
+  }
+}
+```
+
+Each time an event is generated, all reducers that are listening to it are triggered. Note that event ordering is
```
javiertoledo, line 780: In the current implementation, an event can be only reduced by its entity once.
```suggestion
Each time an event is registered, the reducer of its entity is triggered. Note that event ordering is
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+preserved per entity instance. This means that if two events arrive at the same time at a reducer, the first one to
```
javiertoledo, line 781:
```suggestion
preserved per entity instance. This means that one entity will receive just **one event each time**, and all other events of any kind that belong to the same entity will be waiting in a queue until the previous reducer has finished. This is important to make sure that entities state can be easily predicted even when events are being generated concurrently all over the place. It's also one of the reasons why we recommend to keep reducer functions simple and pure: with no side effects or external data gathering.
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+be picked will be the one that was generated first.
```
javiertoledo, line 782:
```suggestion
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+
 #### Entities naming convention
 
+Entities model your data, so name them as closely to your domain as possible. Typical entity names are nouns that
```
javiertoledo, line 786:
```suggestion
Entities are a representation of your application state in a specific moment, so name them as closely to your domain objects as possible. Typical entity names are nouns that
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+
+```text
+boost new:entity Product --fields displayName:string description:string price:Money
+```
+
+The generator will automatically create a file called `Product.ts` with a TypeScript class of the same name under the entities directory. You can still create the entity manually. Since the generator is not doing any magic, all you need is a class decorated as `@Command`. Anyway, we recommend you always to use the generator, because it handles the boilerplate code for you.
+
+Note:
+
+> Running the entity generator with an EntityName that already exists, it will override the content of the current one. In future releases, we will display a warning before overwriting anything. Meantime, if you missed a field, just add it to the class because in Booster, all the infrastructure and data structures are inferred from your code.
```
javiertoledo, line 823:
```suggestion
> Running the entity generator with an EntityName that already exists will overwrite the content of the current one. In future releases, we will display a warning before overwriting anything. Meantime, if you missed a field, just add it to the class because, in Booster, all the infrastructure and data structures are inferred from your code.
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+
 #### The reducer function
 
+Booster generates the reducer function as a static method of the entity class. That function is called by the framework every time that an event of the
+specified type is emitted. **This function must be pure**, which means that you shouldn't be calling third party services, reading or writing to a database, or changing something somewhere.
```
javiertoledo, line 828:
```suggestion
specified type is emitted. It's highly recommended to **keep your reducer functions pure**, which means that you should be able to produce the new entity version by just looking at the event and the previous entity state (which are both injected via parameter by the framework). You should avoid calling third party services, reading or writing to a database, or changing any external state.
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+
+The reducer has two parameters by default:
```
javiertoledo, line 830:
```suggestion
Booster injects two parameters to the reducer functions:
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+
+- `event` - The event object that has triggered this reducer
```
javiertoledo, line 832:
```suggestion
- `event` - The event object that triggered the reducer
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+- `currentEntity?` - A possible entity generated from a previous call to the reducer. **This parameter is optional**, and will be `undefined` when an event occurred for the very first time.
```
javiertoledo, line 833:
```suggestion
- `currentEntity?` - The current state of the entity instance that the event belongs to if it exists. **This parameter is optional** and will be `undefined` if the entity don't exist yet (For example, when you process a `NoteCreated` event that will generate the first version of a `Note` entity).
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+
+This is the place to write business logic for your data generation.
```
javiertoledo, line 835:
```suggestion
This is how you define how events change your application state.
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+
+Then, the reducers of the `ProfileFavorites` entity will be subscribed to
+these newly emitted events. Aggregating the required data into a single entity.
+
 #### Eventual consistency
 
+Due to the event driven and async nature of Booster, your data might
+not be instantly updated. Booster will consume the commands,
+generate events, and _eventually_ generate the entities. Most of the
+time this is not perceivable, but under huge load this might appear.
```
javiertoledo, line 884:
```suggestion
time this is not perceivable, but under huge loads, it could be noticed.
```

docs/README.md
```diff
@@ -753,16 +754,138 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+
+This property is called [Eventual Consistency](https://en.wikipedia.org/wiki/Eventual_consistency), and it is a trade-off of high availability for
```
javiertoledo, line 886:
```suggestion
This property is called [Eventual Consistency](https://en.wikipedia.org/wiki/Eventual_consistency), and it is a trade-off to have high availability for
```

docs/README.md
NickSeagull: I don't understand why removing this part of the sentence

docs/README.md
NickSeagull: Will apply this, but changing the domain to the ecommerce one, to keep with the tone

NickSeagull: Done https://github.com/boostercloud/booster/pull/225/commits/b2e1928eb2a349878647c4edde045f6f1d3c3552

NickSeagull:
We are finally moving the Aggregating data with entities to the _Going deeper with Booster_ section

docs/README.md
```diff
@@ -977,16 +978,121 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
 
 ### 4. Entities and reducers
 
+The source of truth of your Booster app are the events, but events make sense in the context of a domain entity.
+For example, in a banking app, there might be two events: `MoneyDeposited` and `MoneyWithdrawn`. But these events
+only make sense in the context of a `BankAccount`.
+
+Entities are created on the fly, by _reducing_ the whole event stream. You shouldn't assume that they are stored anywhere, but Booster
```
itrion, line 985:
```suggestion
Entities are created on the fly, by _reducing_ the whole event stream. You shouldn't assume that they are stored anywhere. However, Booster
```

docs/README.md
itrion: For me, the sentence is redundant after reading the paragraph, and it sounds like a sentence from the Master Yoda 😂  I'd also remove it.

docs/README.md
```diff
@@ -977,16 +978,121 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
+
+Entities live within the entities directory of the project source: `project-root/src/entities`.
+
+```text
+project-root
+├── src
+│   ├── commands
+│   ├── common
+│   ├── config
+│   ├── entities <------ They must be here
```
itrion, line 1029:
```suggestion
│   ├── entities <------ put them here
```

We have agreed that it is not something mandatory and I've updated some of them here: https://github.com/boostercloud/booster/pull/226/commits/ed28013afe8c683b15a83f63f94651df1b5a27e7

docs/README.md
```diff
@@ -977,16 +978,121 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
-#### Aggregate data using entities
+Booster generates the reducer function as a static method of the entity class. That function is called by the framework every time that an event of the
+specified type is emitted. It's highly recommended to **keep your reducer functions pure**, which means that you should be able to produce the new entity version by just looking at the event and the previous entity state (which are both injected via parameter by the framework). You should avoid calling third party services, reading or writing to a database, or changing any external state.
+
+Booster injects two parameters to the reducer functions:
+
+- `event` - The event object that triggered the reducer
+- `currentEntity?` - The current state of the entity instance that the event belongs to if it exists. **This parameter is optional** and will be `undefined` if the entity don't exist yet (For example, when you process a `ProductCreated` event that will generate the first version of a `Product` entity).
+
+This is how you define how events change your application state.
```
itrion, line 1059:
```suggestion
This is how events change your application state.
```

docs/README.md
javiertoledo: Because it's not needed after rephrasing the previous line

PR #226, opened 2020-06-29
Document the events section

itrion:
## Description
This PR adds the documentation for the events section

## Checks
- [ ] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -964,7 +964,7 @@ where the schema for `CreateProductInput` is
 
 ### 2. Events
 
-Events are **immutable facts** within your application's domain. They are the cornerstone of Booster because of its event-driven and event-sourced nature. Booster events are TypeScript classes decorated with `@Event`. An event class may looks like this:
+Events are **immutable records of facts** within your application's domain. They are the cornerstone of Booster because of its event-driven and event-sourced nature. Booster events are TypeScript classes decorated with `@Event`. An event class may look like this:
 
 ```typescript
 @Event
--- a/docs/README.md
+++ b/docs/README.md
@@ -983,7 +983,7 @@ All the event classes must implement an `entityID` method, returning the ID of t
 
 #### Events naming convention
 
-As with commands, you can name events in any way you want, depending on your application's domain. Though, we recommend you to chose short sentences written in past tense because events are facts that have happened and can't be changed. Some event names would be:
+As with commands, you can name events in any way you want, depending on your application's domain. Though, we recommend you to choose short sentences written in past tense because events are facts that have happened and can't be changed. Some event names would be:
 
 - ProductCreated
 - ProductUpdated
--- a/docs/README.md
+++ b/docs/README.md
@@ -30,7 +30,7 @@
   * [2. Events](#2-events)
     + [Events naming convention](#events-naming-convention)
     + [Creating events](#creating-events)
-    + [The event store](#the-event-store)
+    + [Registering events in the event store](#registering-events-in-the-event-store)
     + [Events ordering](#events-ordering)
   * [3. Event handlers](#3-event-handlers)
     + [Creating an event handler](#creating-an-event-handler)
--- a/docs/README.md
+++ b/docs/README.md
@@ -1047,9 +1047,9 @@ That will generate a file called `StockMoved.ts` under the proper `project-root/
 
 Note:
 
-> Running the event generator for an existing EventName, will override the content of the current one. Soon, we will display a warning before overwriting anything. Meanwhile, if you missed a field, just add it to the class because, in Booster, all the infrastructure and data structures are inferred from your code.
+> Running the event generator for an existing EventName, will overwrite the content of the current one. Soon, we will display a warning before overwriting anything. Meanwhile, if you missed a field, just add it to the class because, in Booster, there is no hidden magic, all the infrastructure and data structures are inferred from your code.
 
-#### Registering events and the event store
+#### Registering events in the event store
 
 Creating an event file is different than storing an event instance in the event store. Booster calls the last `registering` an event. As said before, Booster applications are event-sourced, which means that all the events are stored forever. Think that all events go to a kind of event store, the store is a log queried by the [reducer functions](#4-entities-and-reducers) to recreate the application's current state.
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -764,7 +764,7 @@ Despite you can place commands, and other Booster files, in any directory, we st
 ```text
 project-root
 ├── src
-│   ├── commands <------ They should be here
+│   ├── commands <------ put them here
 │   ├── common
 │   ├── config
 │   ├── entities
@@ -1030,7 +1030,7 @@ project-root
 │   ├── common
 │   ├── config
 │   ├── entities
-│   ├── events <------ They must be here
+│   ├── events <------ put them here
 │   ├── index.ts
 │   └── read-models
 ```
@@ -1140,8 +1140,7 @@ project-root
 │   ├── config
 │   ├── entities
 │   ├── events
-│   ├── event-handlers <------ They must be here
-│   ├── index.ts
+│   ├── event-handlers <------ put them here
 │   └── read-models
 ```
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -1051,9 +1051,9 @@ Note:
 
 #### Registering events in the event store
 
-Creating an event file is different than storing an event instance in the event store. Booster calls the last `registering` an event. As said before, Booster applications are event-sourced, which means that all the events are stored forever. Think that all events go to a kind of event store, the store is a log queried by the [reducer functions](#4-entities-and-reducers) to recreate the application's current state.
+Creating an event file is different than storing an event instance in the event store. In Booster terminology, the latter receives the name of `registering` an event. As said before, Booster applications are event-sourced, which means that all the events are stored forever. Imagine this store as an infinite log used by the [reducer functions](#4-entities-and-reducers) to recreate the application's current state.
 
-Booster injects the register as a parameter in the `handle` method of both the command and the event handlers. Then, registering an event is as simple as calling this function `register.events([event1, event2, ...])`, e.g.
+Booster injects the register as a parameter in the `handle` method of both the command and the event handlers. Then you can register events by calling the `register.events(...)` method as many times as you want, e.g.
 
 ##### Registering events from command handlers
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -989,8 +989,9 @@ export class CartPaid {
     readonly paymentID: UUID) {}
 
   public entityID(): UUID {
-    // returning cartID or paymentID depends on the business logic
-    return this.paymentID
+    // returns cartID because we want to associate
+    // (and reduce) it within the Cart entity
+    return this.cartID
   }
 }
 ```
@@ -1006,6 +1007,8 @@ export class ProductCreated {
   ) {}
 
   public entityID(): UUID {
+    // returns a new UUID because the Product entity
+    // does not exist yet
     return UUID.generate()
   }
 }
--- a/docs/README.md
+++ b/docs/README.md
@@ -979,7 +979,7 @@ export class EventName {
 }
 ```
 
-Events and [entities](#4-entities-and-reducers) are closely related because entities represent the application's state after reducing the stream of events. All the events associated with the same entity are reduced together in the order they have occurred. If an event is related to an entity or not, it is domain-dependent. That's why Booster requires you to implement the `entityID` method. Entities and events are somehow coupled, but that coupling is limited to data. Take a look at this event:
+Events and [entities](#4-entities-and-reducers) are intimately related. All events belong to one entity through the `entityID` method, and entities represent the application's state after reducing the stream of events. Indeed, an entity is just an aggregated representation of the same data present in its events. It is possible to rebuild entities from events at any time. Booster guarantees that all the events associated with an entity, will be reduced in the same order they were stored. Take a look at this event:
 
 ```typescript
 @Event
@@ -995,11 +995,21 @@ export class CartPaid {
 }
 ```
 
-Note:
+An event must be able to find the entity ID value within the event attributes. Either among a parameter injected in the constructor or because it is nested in one of them. In the `CartPaid` example, the entity ID (`paymentID`) is injected directly. However, there are scenarios where the entity ID is not known, or it doesn't exist. For example, in events like `ProductCreated` you can return a completely new ID:
 
-> An event must be able to find the entity ID value within the arguments. Or well because it got injected directly, or because it is present as a nested attribute. In the case of the `CartPaid` example, the entity ID is injected directly.
+```typescript
+@Event
+export class ProductCreated {
+  public constructor(
+    readonly displayName: string,
+    readonly price: Money
+  ) {}
 
-In most situations, you can reduce the event stream to an entity like a Cart or Payment. However, in some use cases, the event stream relates to a specific entity without reducing it. For example, a record of sensor values in a weather station; in that case, the station has no particular value to be reduced. You can implement the semantics that best suits your needs.
+  public entityID(): UUID {
+    return UUID.generate()
+  }
+}
+```
 
 #### Events naming convention
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -964,14 +964,14 @@ where the schema for `CreateProductInput` is
 
 ### 2. Events
 
-Events are **immutable facts** within your application's domain. They are the cornerstone of Booster because of its event-driven and event-sourced nature. Booster events are TypeScript classes decorated with `@Event`:
+Events are **immutable facts** within your application's domain. They are the cornerstone of Booster because of its event-driven and event-sourced nature. Booster events are TypeScript classes decorated with `@Event`. An event class may looks like this:
 
 ```typescript
 @Event
 export class EventName {
   public constructor(readonly field1: SomeType,
                      readonly field2: SomeOtherType,
-                     /*as many fields as you needed*/) {}
+                     /* This event's entity ID must be present */) {}
 
   public entityID(): UUID {
     return /* the associated entity ID */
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -676,15 +676,112 @@ In this chapter you'll walk through these concepts and its details.
 
 ### 2. Events
 
+Events are **immutable facts** within your application's domain. They are the cornerstone in Booster because of its event-driven and event-sourced nature. Booster events are TypeScript classes decorated with `@Event `:
```
tainguyenbui, line 679:
```suggestion
Events are **immutable facts** within your application's domain. They are the cornerstone of Booster because of its event-driven and event-sourced nature. Booster events are TypeScript classes decorated with `@Event`:
```

docs/README.md
```diff
@@ -676,15 +676,112 @@ In this chapter you'll walk through these concepts and its details.
...
+
+```typescript
+@Event
+export class EventName {
+  public constructor(readonly field1: SomeType,
```
tainguyenbui, line 684: one of these fields must be the id of the entity, it could be of type UUID or string, is it worth doing something in relation to that?

docs/README.md
```diff
@@ -676,15 +676,112 @@ In this chapter you'll walk through these concepts and its details.
...
+    return /* the associated entity ID */
+  }
+}
+```
+
+All the event classes must implement an `entityID` method, returning the ID of the entity they are tied to. This id will be used later for building up the application's state via [entities](#4-entities-and-reducers). In most situations, you can reduce the event stream to an entity like a Cart. However, in some use cases, the event stream relates to a specific entity without reducing it. For example, a register of sensor values in a weather station; in that case, the station has no specific value to be reduced. You can implement the semantics that best suits your needs.
+
 #### Events naming convention
 
+As with commands, you can name events in any way you want, depending on your application's domain. Though, we recommend you to chose short sentences written in past tense because events are facts that have happened and can't be changed. Some event names would be:
```
tainguyenbui, line 698:
```suggestion
As with commands, you can name events in any way you want, depending on your application's domain. Though, we recommend you to choose short sentences written in past tense because events are facts that have happened and can't be changed. Some event names would be:
```

docs/README.md
```diff
@@ -676,15 +676,112 @@ In this chapter you'll walk through these concepts and its details.
...
+
+- ProductCreated
+- ProductUpdated
+- ProductDeleted
+- CartItemChanged
+- StockMoved
+
+As with other Booster files, events have their directory:
```
tainguyenbui, line 706:
```suggestion
As with other Booster files, events have their own directory:
```

docs/README.md
```diff
@@ -676,15 +676,112 @@ In this chapter you'll walk through these concepts and its details.
...
+
+```shell
+boost new:event StockMoved --fields productID:string origin:string destination:string quantity:number
+```
+
+That will generate a file called `StockMoved.ts` under the proper `project-root/src/events` directory. You can also create the file manually, but we recommend using the generator and avoid dealing manually with boilerplate code.
+
+Note:
+
+> Running the event generator with an EventName already existing, will override the content of the current one. Soon, we will display a warning before overwriting anything. Meantime, if you missed a field, just add it to the class because, in Booster, all the infrastructure and data structures are inferred from your code.
```
tainguyenbui, line 732:
```suggestion
> Running the event generator for an existing EventName will override the content of the current one. Soon, we will display a warning before overwriting anything. Meanwhile, if you missed a field, just add it to the class because, in Booster, all the infrastructure and data structures are inferred from your code.
```

docs/README.md
```diff
@@ -676,15 +676,112 @@ In this chapter you'll walk through these concepts and its details.
...
+    } else if (event.destination == 'customer') {
+      register.events(new ProductAvailabilityChanged(event.productID, -event.quantity))
+    }
+  }
+}
+```
 
 #### Events ordering
 
+<!-- TODO: several people have asked about how Booster ensures event ordering. I think it makes sense to explain that here  -->
```
tainguyenbui, line 781: Will this bit be added as a future task or in this PR?

docs/README.md
itrion: Yes @tainguyenbui, good catch I'll change it

docs/README.md
itrion: I had it that way before passing the text for Grammarly. They said it was redundant but I'll add it because it sounds better for me too 👍

docs/README.md
itrion: @tainguyenbui, regarding that restriction, is it actually needed to inject the entity ID in the event constructor directly? Or is it ok to just be able to find the entity ID within one of the arguments?

I mean, can we:

```typescript
class SomeType {
  constructor(readonly entityID: UUID, someField: string){}
}

//

@Event
export class EventName {
  public constructor(readonly field1: SomeType,
                     readonly field2: SomeOtherType,
                     /*as many fields as you needed*/) {}

  public entityID(): UUID {
    return field1.entityID
  }
}


```

tainguyenbui: this id is really up to you. I wouldn't name the property entityID though.

Let's say that there is a class that has 4 properties, two of them UUID, you will then choose which one to use. That ID is important because for instance, when using read models, it will be the ID you have to pass if you want to find an item in the database

docs/README.md
itrion: In a different PR, I'll create a ticket because I don't know how we ensure events ordering

docs/README.md
itrion: ok, so it has nothing to do with the constructor parameters but with being able to return an existing entity ID from the `entityID` method right?

tainguyenbui: nothing to do with the constructor, that is right

docs/README.md
itrion: @tainguyenbui, I applied the changes except the `Warning:`, we are not warning yet because that's something we could achieve better with a nice style. There are several notes looking like this

Note:

> Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean nec augue est. In hac habitasse platea dictumst. Integer tincidunt dignissim velit, nec tincidunt lacus fringilla non. Pellentesque at porta risus. Integer pretium augue ut luctus sagittis.

docs/README.md
```diff
@@ -900,15 +900,132 @@ where the schema for `CreateProductInput` is
 
 ### 2. Events
 
+Events are **immutable facts** within your application's domain. They are the cornerstone of Booster because of its event-driven and event-sourced nature. Booster events are TypeScript classes decorated with `@Event`. An event class may looks like this:
```
javiertoledo, line 903:
```suggestion
Events are **immutable records of facts** within your application's domain. They are the cornerstone of Booster because of its event-driven and event-sourced nature. Booster events are TypeScript classes decorated with `@Event`. An event class may look like this:
```

docs/README.md
javiertoledo: `entityID` is a method to allow people to find their entityID in any way they want, it's not a requirement that it is a field in the event itself.

docs/README.md
```diff
@@ -900,15 +900,132 @@ where the schema for `CreateProductInput` is
...
+                     readonly field2: SomeOtherType,
+                     /* This event's entity ID must be present */) {}
+
+  public entityID(): UUID {
+    return /* the associated entity ID */
+  }
+}
+```
+
+Events and [entities](#4-entities-and-reducers) are closely related because entities represent the application's state after reducing the stream of events. All the events associated with the same entity are reduced together in the order they have occurred. If an event is related to an entity or not, it is domain-dependent. That's why Booster requires you to implement the `entityID` method. Entities and events are somehow coupled, but that coupling is limited to data. Take a look at this event:
```
javiertoledo, line 918: In the current implementation, it's a requirement that an event is related to one entity, and entities can only reduce their own events. I've changed the paragraph a little bit to reflect that:

```suggestion
Events and [entities](#4-entities-and-reducers) are intimately related. All events belong to one entity through the `entityID` field, and entities represent the application's state after reducing the stream of events. Indeed, an entity could be deleted and rebuilt from the events anytime, so an entity is just an aggregate representation of the same data that is stored in its events. It's guaranteed by Booster that all the events associated with an entity are reduced in the same order in which they have been stored. Take a look at this event:
```

docs/README.md
```diff
@@ -900,15 +900,132 @@ where the schema for `CreateProductInput` is
...
+  public entityID(): UUID {
+    // returning cartID or paymentID depends on the business logic
+    return this.paymentID
+  }
+}
+```
+
+Note:
+
+> An event must be able to find the entity ID value within the arguments. Or well because it got injected directly, or because it is present as a nested attribute. In the case of the `CartPaid` example, the entity ID is injected directly.
```
javiertoledo, line 936:
```suggestion
> An event must be able to find the entity ID value within the event attributes, either among the parameters injected in the constructor or because it is present in a nested attribute of some of the event attributes. In the case of the `CartPaid` example, the entity ID (`paymentID`) is injected directly.
```

docs/README.md
```diff
@@ -900,15 +900,132 @@ where the schema for `CreateProductInput` is
...
+
+In most situations, you can reduce the event stream to an entity like a Cart or Payment. However, in some use cases, the event stream relates to a specific entity without reducing it. For example, a record of sensor values in a weather station; in that case, the station has no particular value to be reduced. You can implement the semantics that best suits your needs.
```
javiertoledo, line 938: In the current implementation, it is required that all events are reduced by their entity. We'll probably change this later, but this is how it currently works.
```suggestion
```

docs/README.md
```diff
@@ -900,15 +900,132 @@ where the schema for `CreateProductInput` is
...
+
+```shell
+boost new:event StockMoved --fields productID:string origin:string destination:string quantity:number
+```
+
+That will generate a file called `StockMoved.ts` under the proper `project-root/src/events` directory. You can also create the file manually, but we recommend using the generator and avoid dealing manually with boilerplate code.
+
+Note:
+
+> Running the event generator for an existing EventName, will override the content of the current one. Soon, we will display a warning before overwriting anything. Meanwhile, if you missed a field, just add it to the class because, in Booster, all the infrastructure and data structures are inferred from your code.
```
javiertoledo, line 976:
```suggestion
> Running the event generator for an existing EventName, will overwrite the content of the current one. Soon, we will display a warning before overwriting anything. Meanwhile, if you missed a field, just add it to the class because, in Booster there's no hidden magic, all the infrastructure and data structures are inferred from your code.
```

docs/README.md
```diff
@@ -900,15 +900,132 @@ where the schema for `CreateProductInput` is
...
+
+#### Registering events and the event store
```
javiertoledo, line 978:
```suggestion
#### Registering events in the event store
```

docs/README.md
javiertoledo: It's ensured by sharding events by entity name and ID, but definitely, create a ticket so any of us can elaborate!

docs/README.md
alvaroloes: To elaborate even more: 
There are events that represent the creation of the entity, they are the first event for a specific entity. For example, one called `CartCreated` or `ProductCreated`.

In those cases, the user could even do the following:

```typescript
@Event
export class EventName {
  public constructor(readonly field1: SomeType,
                     readonly field2: SomeOtherType,
                     /*as many fields as you needed*/) {}

  public entityID(): UUID {
    return UUID.generate()
  }
}
```

So in this case, the entityID is not even taken from any place, it is generated in place.

Because of all this variability, we made the `entityID()` member a function, instead of a field.

And also this is the reason why Events are not generated with an `ID` field by default when using generators

docs/README.md
alvaroloes: Agree with your change, Javi, but I would change the following: 
> All events belong to one entity through the `entityID` field...

"All events belong to one entity through the `entityID` **method**"

> Indeed, an entity could be deleted and rebuilt from the events anytime, so an entity is just an aggregate representation of the same data that is stored in its events.

Do you think it is a good idea to say "an entity could be _deleted_"? I think this could mislead people to think entities are stored someplace, instead of just a reduced stream of events (thinking this way will help them better to develop in Booster)

I possible rephrase could be:

"Indeed, an entity is just an aggregate representation of the same data that is stored in its events, so it can be rebuilt from the events anytime"

docs/README.md
itrion: Awesome feedback @alvaroloes I think that was the missing piece here 👍

docs/README.md
itrion: 👍  I edited this paragraph from the old docs

docs/README.md
```diff
@@ -900,15 +900,132 @@ where the schema for `CreateProductInput` is
...
+
+Creating an event file is different than storing an event instance in the event store. Booster calls the last `registering` an event. As said before, Booster applications are event-sourced, which means that all the events are stored forever. Think that all events go to a kind of event store, the store is a log queried by the [reducer functions](#4-entities-and-reducers) to recreate the application's current state.
+
+Booster injects the register as a parameter in the `handle` method of both the command and the event handlers. Then, registering an event is as simple as calling this function `register.events([event1, event2, ...])`, e.g.
```
alvaroloes, line 992:
```suggestion
Booster injects the register as a parameter in the `handle` method of both the command and the event handlers. Then, registering an event is as simple as calling this function `register.events([event1, event2, ...])` (you can also do several calls if you prefer), e.g.
```

docs/README.md
```diff
@@ -900,15 +900,132 @@ where the schema for `CreateProductInput` is
...
+
+```typescript
+@Event
+export class CartPaid {
+  public constructor(
+    readonly cartID: UUID,
+    readonly paymentID: UUID) {}
+
+  public entityID(): UUID {
+    // returning cartID or paymentID depends on the business logic
```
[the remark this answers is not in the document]
itrion: I thought the same but in [the example](https://github.com/boostercloud/booster/blob/master/packages/framework-integration-tests/src/events/CartPaid.ts#L9) of the integration test package, we are returning the paymentID. Actually, the confusion around this makes me write [an RFC](https://agilemonkeys.atlassian.net/wiki/spaces/B/pages/1267269684/Wrap+UUID+with+a+value+object+called+EntitynameID+when+generating+entities+to+avoid+ambiguity) about improving how we handle UUIDs to avoid ambiguity

docs/README.md
```diff
@@ -900,15 +900,132 @@ where the schema for `CreateProductInput` is
...
+As with other Booster files, events have their own directory:
+
+```text
+project-root
+├── src
+│   ├── commands
+│   ├── common
+│   ├── config
+│   ├── entities
+│   ├── events <------ They must be here
```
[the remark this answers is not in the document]
itrion: I see your point, and I have one concern. Changing the project structure would affect the DX because the generator wouldn't be that useful but annoying

what do you think?

I agree we have to change the must, but I couldn't encourage users to change the directories

docs/README.md
itrion: Good point @alvaroloes, I'll open a ticket to address it, as the same note is repeated in other parts too.

docs/README.md
```diff
@@ -900,15 +900,145 @@ where the schema for `CreateProductInput` is
...
+                     readonly field2: SomeOtherType,
+                     /* This event's entity ID must be present */) {}
+
+  public entityID(): UUID {
+    return /* the associated entity ID */
+  }
+}
+```
+
+Events and [entities](#4-entities-and-reducers) are intimately related. All events belong to one entity through the `entityID` method, and entities represent the application's state after reducing the stream of events. Indeed, an entity is just an aggregated representation of the same data present in its events. It is possible to rebuild entities from events at any time. Booster guarantees that all the events associated with an entity, will be reduced in the same order they were stored. Take a look at this event:
```
alvaroloes, line 918:
```suggestion
Events and [entities](#4-entities-and-reducers) are intimately related. All events belong to one entity through the `entityID` method, and entities represent the application's state after reducing the stream of events. Indeed, an entity is just an aggregated representation of the same data present in its events. It is possible to rebuild entities from events at any time. Booster guarantees that all the events associated with an entity will be reduced in the same order they were stored. Take a look at this event:
```

alvaroloes: (just removed a comma in `associated with an entity[,] will be...`)

docs/README.md
```diff
@@ -900,15 +900,145 @@ where the schema for `CreateProductInput` is
...
+
+  public entityID(): UUID {
+    // returns cartID because we want to associate
+    // (and reduce) it within the Cart entity
+    return this.cartID
+  }
+}
+```
+
+An event must be able to find the entity ID value within the event attributes. Either among a parameter injected in the constructor or because it is nested in one of them. In the `CartPaid` example, the entity ID (`paymentID`) is injected directly. However, there are scenarios where the entity ID is not known, or it doesn't exist. For example, in events like `ProductCreated` you can return a completely new ID:
```
alvaroloes, line 935: I see that you left this as it was. Didn't you agree with my suggestion? ([this one](https://github.com/boostercloud/booster/pull/226#discussion_r449045139)

docs/README.md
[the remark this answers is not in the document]
alvaroloes: Yep, but in the future, we could add a `--output | -o` parameter in the generator to tell Booster where you want it to put the file.

docs/README.md
itrion: I'm sorry, I think I forgot to rephrase this, I will. In fact, I included the new UUID example for the case of the `Created*` like events because that helps a lot. The singleton entity use case is also interesting.

docs/README.md
javiertoledo: I wouldn't generate the ID value in the `entityID` function, because in that way the event can't be replayed. I rather generate the IDs before registering them in the command or event handler.

alvaroloes: @itrion 
@javiertoledo is right. And not only that, this would make the current Booster fail, as when we deserialize the event, we call several times to `entityID()` and we expect it to be always the same, and it would not if we generate a random string inside the function. So this is something we not only should not mention in the documentation, but we should discourage it!

itrion: That makes total sense and it worth it to mention that the `entityID` method must also be pure then

PR #228, opened 2020-06-30
BOOST-786: update new docs

MYF95:
## Description

You can see the rendered file [here](https://github.com/boostercloud/booster/blob/808c6e73573bd707e84615bb263057ee35d0741e/docs/README.md)

Filled in documentation with pieces from the old public docs.
Tried mixing the code we previously had on the right with the text and moved some paragraphs into places where they might belong better.

## Changes
Updated README

## Checks
- [X] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1003,6 +1003,281 @@ _accessToken_ | The access token you get in the sign-in call.
 You will get a HTTP status code different from 2XX and a body with a message telling you the reason of the error.
 
 ### GraphQL API
+This is the main API of your application, as it allows you to:
+
+ - _Modify_ data by **sending commands**
+ - _Read_ data by **querying read models**
+ - _Receive data in real time_ by **subscribing to read models** 
+ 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -764,16 +764,565 @@ public static async handle(event: StockMoved, register: Register): Promise<void>
...
 
 #### The projection function
+A `Projection` is a method decorated with the `@Projection` decorator that, given a new entity value and (optionally) a previous read model state, generate a new read model value.
+
+Read models can be projected from multiple [entities](#4-entities-and-reducers) as soon as they share some common key called `joinKey`.
 
 #### Authorizing read models
 
 #### Querying a read model
+Defining a read models enables a new REST Read endpoint that you can use to query or poll the read model records [see the API documentation](#GraphQL-API).
```
NickSeagull, line 806:
```suggestion
You can use the GraphQL endpoint to query or subscribe to the read model records: [see the API documentation](#GraphQL-API).
```

MYF95: Committing the suggestion directly, thanks!

PR #237, opened 2020-07-02
BOOST-794: update deployment section

MYF95:
## Description
Updated deployment section to be inside the features block

## Changes
<Changes made>

## Checks
- [X] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -59,11 +59,8 @@
     + [Sending commands](#sending-commands)
     + [Reading read models](#reading-read-models)
     + [Subscribing to read models](#subscribing-to-read-models)
-- [Deploying](#deploying)
-  * [AWS](#aws)
-    + [Configure your provider credentials](#configure-your-provider-credentials)
-  * [Deploy your project](#deploy-your-project)
-  * [Deleting your cloud stack](#deleting-your-cloud-stack)
+  * [Cloud native](#cloud-native)
+    + [Deploying to AWS](#deploying-to-aws)
 - [Frequently Asked Questions](#frequently-asked-questions)
 
 <!-- tocstop -->
@@ -1801,7 +1798,7 @@ mutation {
 }
 ```
 
-### Serverless ready
+### Cloud native
 
 One of the goals of Booster is to become provider agnostic so you can deploy your application to any serverless provider like AWS, Google Cloud, Azure, etc...
 
@@ -1811,6 +1808,7 @@ all cloud providers. (**Contributions are welcome!** 😜)
 #### Deploying to AWS
 
 ##### Configure your provider credentials
+
 > Creating a plain text file manually named `~/.aws/credentials` with the following content will be enough:
 
 ```text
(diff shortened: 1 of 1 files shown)


MYF95:
Didn't change any of the main text because I think we're moving forward to simply having the docs and will revisit/rewrite it later on. Left a comment on the only part of the text which kind of might not make sense, but let me know in any other case if we should change anything else @itrion

docs/README.md
```diff
@@ -1496,16 +1496,16 @@ mutation {
 }
 ```
 
-## Deploying
+### Serverless ready
 
 One of the goals of Booster is to become provider agnostic so you can deploy your application to any serverless provider like AWS, Google Cloud, Azure, etc...
 
 So far, in the current version, only AWS is supported, but given the high level of abstraction, it will eventually support
```
MYF95, line 1503: Not sure if this paragraph is appropiate to be here, since we're currently having more versions (like Kubernetes now) almost ready.

docs/README.md
```diff
@@ -1496,16 +1496,16 @@ mutation {
 }
 ```
 
-## Deploying
+### Serverless ready
```
javiertoledo, line 1499: Why renaming this section? For me "Deploying" was clearer, maybe "Deploying your application"?

docs/README.md
```diff
@@ -1496,16 +1496,16 @@ mutation {
...
 
 One of the goals of Booster is to become provider agnostic so you can deploy your application to any serverless provider like AWS, Google Cloud, Azure, etc...
 
 So far, in the current version, only AWS is supported, but given the high level of abstraction, it will eventually support
 all cloud providers. (**Contributions are welcome!** 😜)
 
-### AWS
+#### Deploying in AWS
```
javiertoledo, line 1506:
```suggestion
#### Deploying to AWS
```

docs/README.md
```diff
@@ -1531,7 +1531,7 @@ It will take a while, but you should have your project deployed to your cloud pr
 
 If you make changes to your code, you can run `boost deploy -e production` again to update your project in the cloud.
 
-### Deleting your cloud stack
+##### Deleting your cloud stack
```
javiertoledo, line 1534: To be consistent with other titles in this section:
```suggestion
##### Delete your cloud stack
```

docs/README.md
NickSeagull: Agree

itrion: @NickSeagull @javiertoledo, the idea is to have this structure:

- Features
  - Serverless ready (is a feature)
    - Deploying to AWS
    - Deploying to Azure
    - Deploying to a Kubernetes cluster
    - Deploying to Pivotal Cloud Foundry (or whatever cloud supporting functions)

That's the reason for this change. Take into account that the table of content shows headings up to level 3, so the Deploying to AWS section is still quickly visible.

**Having said that, do you still want to change the headings structure?** it is OK for me, I just want to be sure you understand the motivation because for me **the Booster capacity to deploy in any cloud is a big win and a great feature.**

NickSeagull: I agree with your explanation then, but I'd call the section "Cloud Native" rather than "Serverless ready"

PR #239, opened 2020-07-03
Add a table of content at the beginning of the docs

itrion:
## Description
In addition to the toc itself, this PR adds a development dependency and a script for updating the table of content just by running `npm run update-documentation-toc` in the root directory. That is one step behind automatizing this flow with a GitHub action.

There were 2 subsections with a wrong indentation level, I fixed that (the TOC make the error pretty clear :P )

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "eslint-config-prettier": "^6.10.0",
     "eslint-plugin-prettier": "^3.1.2",
     "lerna": "^3.20.2",
+    "markdown-toc": "1.2.0",
     "mocha": "^7.1.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,6 +5,9 @@
   "homepage": "https://booster.cloud",
   "bugs": "https://github.com/boostercloud/booster/issues",
   "private": true,
+  "scripts": {
+    "update-documentation-toc": "markdown-toc -i --maxdepth 4 docs/README.md"
+  },
```
javiertoledo, line 10: It'd be nice to call this script from a Github action

PR #242, opened 2020-07-08
[BOOST-473] local provider commands storing events in db

tainguyenbui:
## Description
Adds first integration test for local provider to verify that an event is written to the DB when a command is fired

## Changes
- Adds tests
- Fix issue with local provider `start`

## Checks
- [x] Project Builds
- [n/a] Updated documentation accordingly


tainguyenbui:
bot: integration

PR #244, opened 2020-07-08
Wait after integration tests deployment and before tests begin

alvaroloes:
## Description
After a whole day of chasing the random errors we were getting in integration tests, it turned out that we needed to wait for some time after the deployment of the AWS stack. 

This explains why the integration tests always worked in stacks that had been already deployed before, and not in new stacks that had been deployed right before the tests start.

I also took the opportunity to do some other reactors to increase the reliability of tests

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-integration-tests/integration/providers/aws/setup.ts
+++ b/packages/framework-integration-tests/integration/providers/aws/setup.ts
@@ -2,10 +2,14 @@ import util = require('util')
 const exec = util.promisify(require('child_process').exec)
 import { deploy, nuke } from '../../../src/deploy'
 import { config } from 'aws-sdk'
+import { sleep } from '../helpers'
 
 before(async () => {
   await setEnv()
   await checkConfigAnd(deploy)
+  console.log('Waiting 30 seconds after deployment to let the stack finish its initialization...')
+  await sleep(30000)
+  console.log('...sleep finished. Let the tests begin 🔥!')
 })
 
 after(async () => {
--- a/packages/framework-integration-tests/integration/providers/aws/subscriptions.integration.ts
+++ b/packages/framework-integration-tests/integration/providers/aws/subscriptions.integration.ts
@@ -37,28 +37,33 @@ describe('subscriptions', () => {
 
   describe('the "terminate" operation', () => {
     it('should delete al subscription of the connectionID when socket is disconnected', async () => {
-      const originalSubscriptionsCount = await countSubscriptionsItems()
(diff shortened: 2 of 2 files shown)


alvaroloes:
bot: integration

packages/framework-integration-tests/integration/providers/aws/subscriptions.integration.ts
```diff
@@ -17,12 +17,11 @@ describe('subscriptions', () => {
     })
 
     it('should delete a subscription when the client calls "unsubscribe"', async () => {
-      const mockCartId = random.uuid()
       const originalSubscriptionsCount = await countSubscriptionsItems()
 
       // Let's create two subscriptions to the same read model
-      cartSubscription(client, mockCartId).subscribe(() => {})
-      const subscriptionObservable = cartSubscription(client, mockCartId).subscribe(() => {})
+      cartSubscription(client, random.uuid()).subscribe(() => {})
```
tainguyenbui, line 23: is there any difference between subscribing to an existing cart id vs a non-existing one?

packages/framework-integration-tests/integration/providers/aws/setup.ts
```diff
@@ -2,10 +2,14 @@ import util = require('util')
 const exec = util.promisify(require('child_process').exec)
 import { deploy, nuke } from '../../../src/deploy'
 import { config } from 'aws-sdk'
+import { sleep } from '../helpers'
 
 before(async () => {
   await setEnv()
   await checkConfigAnd(deploy)
+  console.log('Waiting 30 seconds after deployment to let the stack finish its initialization...')
```
javiertoledo, line 10: Did anything change? It worked without waiting so much before... 🤔

packages/framework-integration-tests/integration/providers/aws/subscriptions.integration.ts
alvaroloes: It should not. But this was 80% out of my desperation and 20% a suspicion that ApolloClient could be caching this (it should not)

packages/framework-integration-tests/integration/providers/aws/setup.ts
alvaroloes: Yes, the WebSockets. We are adding more and more subscriptions tests and they use sockets, whose AWS resource is the main source of inconsistencies right after the stack is deployed. 

We might need to revisit this in the future to be really sure.

PR #245, opened 2020-07-08
Added a suggestion to place role classes

javiertoledo:
When creating your first project in Booster, it's common to get confused about where to put your role classes. Booster doesn't require to put them in any specific place, but it's a good idea to give a suggestion.


javiertoledo:
Not yet, [we have a ticket for that](https://agilemonkeys.atlassian.net/browse/BOOST-834).

PR #248, opened 2020-07-14
Suggestions and changes to "Commands" and "Events" architecture sections

alvaroloes:
## Description
As I read these sections, I was doing small corrections and clarifications

## Checks
- [x] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -711,7 +711,7 @@ Two patterns influence the Booster's event-driven architecture: Command-Query Re
 
 ![architecture](./img/booster-arch.png)
 
-The public interface of a Booster application is just `Commands` and `ReadModels`. Booster proposes an entirely different approach to the Model-View-\* and CRUD frameworks. With Booster, the clients submit commands, query the read models, or subscribe to them for receiving real-time updates thanks to the out of the box [GraphQL API](#graphql-api)
+The public interface of a Booster application is just `Commands` and `ReadModels`. Booster proposes an entirely different approach to the Model-View-\* and CRUD frameworks. With Booster, clients submit commands, query the read models, or subscribe to them for receiving real-time updates thanks to the out of the box [GraphQL API](#graphql-api)
 
 Booster applications are event-driven and event-sourced so, **the source of truth is the whole history of events**. When a client submits a command, the `CommandHandler` _wakes up_ and executes its logic. Optionally, it can _register_ as many `Events` as needed. The framework caches the current state by automatically _reducing_ all the registered events into `Entities`. Interested parties can _react_ to events via `EventHandlers`, and finally, the _projection_ functions transform the entities into `ReadModels`.
 
@@ -719,9 +719,9 @@ In this chapter you'll walk through these concepts and its details.
 
 ### 1. Command and command handlers
 
-Booster is different than MVC frameworks in which you typically implement controller classes with the five CRUD methods per model. Instead of that, you define commands, which are the user actions when interacting with an application. This approach fits very well with Domain-Driven Design. Depending on your application's domain, some examples of commands would be: `RemoveItemFromCart`, `RatePhoto`, `AddCommentToPost`, etc. Although, you can still have `Create*`, `Delete*`, or `Update*` commands when they make sense.
+Booster is different than MVC frameworks in which you typically implement controller classes with the four CRUD methods per model. Instead of that, you define commands, which are the user actions when interacting with an application. This approach fits very well with Domain-Driven Design. Depending on your application's domain, some examples of commands would be: `RemoveItemFromCart`, `RatePhoto`, `AddCommentToPost`, etc. Although, you can still have `Create*`, `Delete*`, or `Update*` commands when they make sense.
 
-There is an architectural splitting between commands and command handlers though they *live* under the same file. The command is the class with the `@Command` decorator, and the generated method called `handle` is the command handler. That is because Booster adopts several concepts from functional programming; the separation between data structures and data transformations is one of them. In Booster a command looks like this:
+There is an architectural splitting between commands and command handlers though they *live* under the same file. The command is the class with the `@Command` decorator, and the generated method called `handle` is the command handler. That is because Booster adopts several concepts from functional programming; the separation between data structures and data transformations is one of them. In Booster, a command looks like this:
 
 ```typescript
 @Command({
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -711,17 +711,17 @@ Two patterns influence the Booster's event-driven architecture: Command-Query Re
...
 
 In this chapter you'll walk through these concepts and its details.
 
 ### 1. Command and command handlers
 
-Booster is different than MVC frameworks in which you typically implement controller classes with the five CRUD methods per model. Instead of that, you define commands, which are the user actions when interacting with an application. This approach fits very well with Domain-Driven Design. Depending on your application's domain, some examples of commands would be: `RemoveItemFromCart`, `RatePhoto`, `AddCommentToPost`, etc. Although, you can still have `Create*`, `Delete*`, or `Update*` commands when they make sense.
+Booster is different than MVC frameworks in which you typically implement controller classes with the four CRUD methods per model. Instead of that, you define commands, which are the user actions when interacting with an application. This approach fits very well with Domain-Driven Design. Depending on your application's domain, some examples of commands would be: `RemoveItemFromCart`, `RatePhoto`, `AddCommentToPost`, etc. Although, you can still have `Create*`, `Delete*`, or `Update*` commands when they make sense.
```
alvaroloes, line 722: Not sure why this is marked as changed 😳

docs/README.md
```diff
@@ -742,11 +742,11 @@ export class CommandName {
...
 
 #### Commands naming convention
 
-Semantic is very important in Booster as it will play an essential role in designing a coherent system. Your application should reflect your domain concepts, and commands are not an exception. Although you can name commands in any way you want, we strongly recommend you to name them starting with verbs in imperative plus the object being affected. If we were designing an e-commerce application, some commands would be:
+Semantics is very important in Booster as it will play an essential role in designing a coherent system. Your application should reflect your domain concepts, and commands are not an exception. Although you can name commands in any way you want, we strongly recommend you to name them starting with verbs in imperative plus the object being affected. If we were designing an e-commerce application, some commands would be:
```
alvaroloes, line 749: Apparently the noun goes always in the plural form

docs/README.md
juanjoman, line 722:
```suggestion
Booster is different than MVC frameworks in which you typically implement controller classes with CRUD methods per model. Instead of that, you define commands, which are the user actions when interacting with an application. This approach fits very well with Domain-Driven Design. Depending on your application's domain, some examples of commands would be: `RemoveItemFromCart`, `RatePhoto`, `AddCommentToPost`, etc. Although, you can still have `Create*`, `Delete*`, or `Update*` commands when they make sense.
```

Just curious about this, doesn't "CRUD" includes specifically 4 methods?

docs/README.md
```diff
@@ -778,22 +778,22 @@ The preferred way to create a command is by using the generator, e.g.
...
 
 Note:
-> Running the command generator with a `CommandName` already existing, will override the content of the current one. Soon, we will display a warning before overwriting anything. Meantime, if you missed a field, just add it to the class because in Booster, all the infrastructure and data structures are inferred from your code.
+> Generating a command with the same name as an already existing one will override its content. Soon, we will display a warning before overwriting anything.
 
 #### The command handler function
 
-Booster generates the command handler function as a method of the command class. This function is called by the framework every time that one instance of this command is submitted. The command handler the right place to run validations, return errors, query entities to make decisions and register relevant domain events.
+Each command class must have a method called `handle`. This function is the command handler, and it will be called by the framework every time one instance of this command is submitted. Inside it you can run validations, return errors, query entities to make decisions and register relevant domain events.
```
tainguyenbui, line 788:
```suggestion
Each command class must have a method called `handle`. This function is the command handler, and it will be called by the framework every time one instance of this command is submitted. Inside the handler you can run validations, return errors, query entities to make decisions and register relevant domain events.
```

docs/README.md
```diff
@@ -1335,7 +1334,7 @@ You can use the GraphQL endpoint to query or subscribe to the read model records
 
 ## Features
 
-### IAM - Authentication and Authorization
+### Authentication and Authorization
```
tainguyenbui, line 1337: This might require a regeneration of the table of content

docs/README.md
juanjoman, line 788:
```suggestion
Each command class must have a method called `handle`. This function is the command handler, and it will be called by the framework every time one instance of this command is submitted. From command handlers, you can run validations, return errors, query entities to make decisions, and register relevant domain events.
```

docs/README.md
javiertoledo: Yes, but Read is done through the Query API

docs/README.md
```diff
@@ -778,22 +778,22 @@ The preferred way to create a command is by using the generator, e.g.
 boost new:command CreateProduct --fields sku:SKU displayName:string description:string price:Money
 ```
 
-The generator will automatically create a file called `CreateProduct.ts` with a TypeScript class of the same name under the commands directory. You can still create the command manually. Since the generator is not doing any *magic*, all you need is a class decorated as `@Command`. Anyway, we recommend you always to use the generator, because it handles the boilerplate code for you.
+The above generator will automatically create a file called `CreateProduct.ts` with a TypeScript class of the same name under the `commands` directory. You can still create (or modify) the command manually. Since the generator is not doing any *magic*, all you need is a class decorated as `@Command`. Anyway, we recommend you always to use the generator, because it handles the boilerplate code for you.
```
itrion, line 781: @alvaroloes, something I've learned is to avoid explicit mentions to the position of the tables, snippets, graphics, etc. because if you move them, you don't have to change the texts with the reference.

```suggestion
The generator will automatically create a file called `CreateProduct.ts` with a TypeScript class of the same name under the `commands` directory. You can still create (or modify) the command manually. Since the generator is not doing any *magic*, all you need is a class decorated as `@Command`. Anyway, we recommend you always to use the generator, because it handles the boilerplate code for you.
```

I learned it using LaTex because it is in charge of placing the visual elements wherever they fit better, so you actually don't know where they'll be after the build. I know we are not using LaTex here but I still consider it a very good tip as we could be placing snippets in a different position after building the final docs

docs/README.md
```diff
@@ -711,17 +711,17 @@ Two patterns influence the Booster's event-driven architecture: Command-Query Re
...
 
-There is an architectural splitting between commands and command handlers though they *live* under the same file. The command is the class with the `@Command` decorator, and the generated method called `handle` is the command handler. That is because Booster adopts several concepts from functional programming; the separation between data structures and data transformations is one of them. In Booster a command looks like this:
+There is an architectural splitting between commands and command handlers though they *live* under the same file. The command is the class with the `@Command` decorator, and the generated method called `handle` is the command handler. That is because Booster adopts several concepts from functional programming; the separation between data structures and data transformations is one of them. In Booster, a command looks like this:
```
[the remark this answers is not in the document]
alvaroloes: Yeah, at least to be consistent with the other concepts

docs/README.md
alvaroloes: Oh! I changed "*five* CRUD methods..." by "*four* CRUD methods..."

Curious: Does anybody know why it was written "five"? @boostercloud/booster-core @itrion

docs/README.md
alvaroloes: Amazing tip Johan. 100% agree

docs/README.md
alvaroloes: Applied already a similar suggestion above. Thanks!

docs/README.md
alvaroloes: done! b40504a

docs/README.md
alvaroloes: Done b40504a

docs/README.md
alvaroloes: Included the extra comma: b40504a

PR #250, opened 2020-07-14
Suggestions and changes to "Event handlers", "Entities", and "Read models" architecture sections

alvaroloes:
## Description
I finish reading the whole docs before the "Application API" section and these are the final suggestions and fixes

## Checks
- [x] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -759,7 +759,7 @@ Semantic is very important in Booster as it will play an essential role in desig
 Despite you can place commands, and other Booster files, in any directory, we strongly recommend you to put them here `project-root/src/commands`. Having all the commands in one place will help you to understand your application's capabilities at a glance.
 
 ```text
-project-root
+<project-root>
 ├── src
 │   ├── commands <------ put them here
 │   ├── common
@@ -1026,7 +1026,7 @@ As with commands, you can name events in any way you want, depending on your app
 As with other Booster files, events have their own directory:
 
 ```text
-project-root
+<project-root>
 ├── src
 │   ├── commands
 │   ├── common
@@ -1045,7 +1045,7 @@ The preferred way to create event files is the `new:event` generator, e.g.
 boost new:event StockMoved --fields productID:string origin:string destination:string quantity:number
 ```
 
-That will generate a file called `StockMoved.ts` under the proper `project-root/src/events` directory. You can also create the file manually, but we recommend using the generator and avoid dealing manually with boilerplate code.
+That will generate a file called `StockMoved.ts` under the proper `<project-root>/src/events` directory. You can also create the file manually, but we recommend using the generator and avoid dealing manually with boilerplate code.
 
 Note:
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1124,18 +1124,16 @@ export class HandleAvailability {
 
 #### Creating an event handler
 
-Event handlers can be easily created using the Booster CLI. There are two compulsory arguments that will need to be provided following the `boost new:event-handler` command, the first one will be the event handler name, and the other will be the name of the event that it will react to. For instance:
+Event handlers can be easily created using the Booster CLI command `boost new:event-handler`. There are two compulsory arguments: the event handler name, and the name of the event it will react to. For instance:
 
 ```typescript
-boost new:event-handler HandleAvailability -e StockMoved
+boost new:event-handler HandleAvailability --event StockMoved
```
alvaroloes, line 1130: To keep the same style as in other command examples, I changed it by the full parameter name

docs/README.md
```diff
@@ -1160,9 +1158,9 @@ register.events(new ProductAvailabilityChanged(event.productID, -event.quantity)
 
 #### Reading entities from event handlers
 
-Event handlers are also a good place to retrieve entity information before triggering new events.
+Just as we do in command handlers, we can also retrieve entities information to make decisions based on its current state.
```
itrion, line 1161:
```suggestion
Just as we do in command handlers, we can also retrieve entities information to make decisions based on their current state.
```

docs/README.md
```diff
@@ -1235,16 +1235,16 @@ The preferred way to create an entity is by using the generator, e.g.
 boost new:entity Product --fields displayName:string description:string price:Money
 ```
 
-The generator will automatically create a file called `Product.ts` with a TypeScript class of the same name under the entities directory. You can still create the entity manually. Since the generator is not doing any magic, all you need is a class decorated as `@Command`. Anyway, we recommend you always to use the generator, because it handles the boilerplate code for you.
+The generator will automatically create a file called `Product.ts` with a TypeScript class of the same name under the `entities` directory. You can still create the entity manually by creating a class decorated as `@Entity`. Anyway, we recommend you always to use the generator because it handles the boilerplate code for you.
```
itrion, line 1238:
```suggestion
The generator will automatically create a file called `Product.ts` with a TypeScript class of the same name under the `entities` directory. You can still create the entity manually, writing a class decorated with `@Entity`. Anyway, we recommend you always to use the generator because it handles the boilerplate code for you.
```

To avoid using `create` twice in the same sentence

docs/README.md
```diff
@@ -1124,18 +1124,16 @@ export class HandleAvailability {
 
 #### Creating an event handler
 
-Event handlers can be easily created using the Booster CLI. There are two compulsory arguments that will need to be provided following the `boost new:event-handler` command, the first one will be the event handler name, and the other will be the name of the event that it will react to. For instance:
+Event handlers can be easily created using the Booster CLI command `boost new:event-handler`. There are two compulsory arguments: the event handler name, and the name of the event it will react to. For instance:
```
NickSeagull, line 1127: What does `compulsory` mean here?

javiertoledo: Maybe... ??
```suggestion
Event handlers can be easily created using the Booster CLI command `boost new:event-handler`. There are two mandatory arguments: the event handler name, and the name of the event it will react to. For instance:
```

alvaroloes: `compulsory` means "not optional", but I don't like the word too. I like the Javi's suggestion

PR #255, opened 2020-07-16
Local provider event reduction into entities

tainguyenbui:
## Description
Add logic to correctly reduce events into entities and create snapshots

## Changes
- Implement new functions to notify Booster Core that an event has been stored in the Database
- Add tests

## Checks
- [x] Project Builds


tainguyenbui:
bot: integration

PR #259, opened 2020-07-22
Auth tokens via graphql request body

tainguyenbui:
## Description
Allow user Authorization token to be sent as part of the GraphQL request body 

## Changes
- Handle requests where Authorization token is provided in the GraphQL request body
- Update tests to adapt to new logic
- Remove code in core related to the AuthorizerRequest and Lambda Authorizer attached to API Gateway

## Checks
- [x] Project Builds


docs/README.md
```diff
@@ -1856,6 +1856,38 @@ mutation {
 }
 ```
 
+It is also possible to use a GraphQL client to subscribe to changes in the application. However, it might be necessary to include an access token, provided when the user signs-in, in the request if the operation requires authorization. This token should be added to the GraphQL client as an operation option through a middleware. An example can be seen below:
```
tainguyenbui, line 1859: @alvaroloes @itrion I added some more information about how to include the auth token in subscriptions

tainguyenbui:
bot: integration

docs/README.md
alvaroloes: Thanks! In any case, I am writing and rephrasing the GraphQL documentation and I will include this new way of passing the token ;-)

tainguyenbui:
bot: integration

tainguyenbui:
@NickSeagull @javiertoledo I will merge this since it had two approvals already. But feel free to review it later on and create issues if needed /cc @alvaroloes

PR #263, opened 2020-07-27
[BOOST-870] Return value different from 0 for  compilation error

rdoria1:
## Description
When a compilation error happens, even if the error is logged the return value is actually 0. This is problematic when integrating booster applications with CD tools like Circle CI. As described in the ticket https://agilemonkeys.atlassian.net/browse/BOOST-870 the deploy step was marked successful because of the return value.

## Changes
Throw an exception instead of just logging the error.

## Checks
- [X] Project Builds

## Additional information
Also ran `lerna run integration/cli` to run the integration tests for the CLI package where all the changes happen.


rdoria1:
bot: integration

Issue #196, opened 2020-06-11
[CLI] help display for boost sometimes shows an error


glammers1:
Example of command without errors:

```
$ boost -h            
CLI of the Booster Cloud Framework, the next level of abstraction for cloud-native applications

VERSION
  @boostercloud/cli/0.4.1 darwin-x64 node-v12.16.3

USAGE
  $ boost [COMMAND]

COMMANDS
  deploy  Deploy the current application as configured in your `index.ts` file.
  help    display help for boost
  new     generate new resource, write 'boost new' to see options
  nuke    Remove all resources used by the current application as configured in your `index.ts` file.
  run     Run the current application locally, as configured in your configuration files.
```

Example of a command with errors:

```
$ boost new:command -h
generate new resource, write 'boost new' to see options

USAGE
  $ boost new:command [COMMANDNAME]

OPTIONS
  -f, --fields=fields  field that this command will contain
  -h, --help           show CLI help

ExitError: EEXIT: 0
    at Object.exit (/Users/glammers/.nvm/versions/node/v12.16.3/lib/node_modules/@boostercloud/cli/node_modules/@oclif/errors/lib/index.js:18:11)
    at Command.exit (/Users/glammers/.nvm/versions/node/v12.16.3/lib/node_modules/@boostercloud/cli/node_modules/@oclif/command/lib/command.js:54:23)
    at Command._help (/Users/glammers/.nvm/versions/node/v12.16.3/lib/node_modules/@boostercloud/cli/node_modules/@oclif/command/lib/command.js:127:21)
    at Object.parse (/Users/glammers/.nvm/versions/node/v12.16.3/lib/node_modules/@boostercloud/cli/node_modules/@oclif/command/lib/flags.js:37:17)
    at Parser._flags (/Users/glammers/.nvm/versions/node/v12.16.3/lib/node_modules/@boostercloud/cli/node_modules/@oclif/parser/lib/parse.js:149:42)
    at Parser.parse (/Users/glammers/.nvm/versions/node/v12.16.3/lib/node_modules/@boostercloud/cli/node_modules/@oclif/parser/lib/parse.js:117:28)
    at Object.parse (/Users/glammers/.nvm/versions/node/v12.16.3/lib/node_modules/@boostercloud/cli/node_modules/@oclif/parser/lib/index.js:27:27)
[...]

PR #273, opened 2020-08-01
fix(cli) -h flag exit error #196

dotalo:
## Description
This resolves the exit error that happens when appending the `-h` flag on any `booster new:` command.

## Changes
Instead of having a `runWithErrors()` function which is then wrapped with a catch, run all code inside the `public static run()` function, with a try-catch block and the `this.parse()` call outside the try block because it always [throws the ExitError when displaying help with -h](https://github.com/oclif/command/blob/b3efe1e5e052a7a1a0b98b25bf4d8a521b5850d5/src/command.ts#L200). This was applied to every command inside `booster/packages/cli/src/commands/new/`

## Checks
- [x] Project Builds
- [x] Updated documentation accordingly: no documentation needs to be changed.

### DIFF
--- a/packages/cli/src/common/script.ts
+++ b/packages/cli/src/common/script.ts
@@ -109,7 +109,7 @@ export class Script<TContext> {
     } catch (err) {
       const defaultHandler = (e: Error): string => e.stack || e.message
       const handler = this.errorHandlers[err.name] || defaultHandler
-      Script.logger.fail(handler(err))
+      throw new Error(handler(err))
     }
   }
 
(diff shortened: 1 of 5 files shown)


dotalo:
Will open again, as I mixed code. Sorry for any problems I have caused :/

javiertoledo:
Hey @dotalo, no problem at all, I had the chance to skim over the code in the mobile app during the weekend and your changes are definitely worth the wait, we often deprioritize these kind of details as we're too focused on the framework core, for now, thank you very much! If there's anything we can do to help just let us know!

PR #297, opened 2020-08-04
Enhancing README file with notes

davidverdu:
Detailing some steps to run the example better

## Description
enhancing the documentation

## Changes
README file

## Checks
- [x] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -622,7 +622,7 @@ In this section, we will be sending requests by hand using the online tool [Post
 which is free and includes great support for GraphQL. However, you can use any client you want. Your endpoint URL should look like this:
 
 ```text
-https://<some random number>.execute-api.us-east-1.amazonaws.com/production/graphql
+<httpURL>/graphql
 ```
 
 ##### 7.1 Creating posts
--- a/docs/README.md
+++ b/docs/README.md
@@ -480,6 +480,12 @@ public handle(register: Register): void {
 }
 ```
 
+Remember to import the event class correctly on the top part of the file:
+
+```typescript
+import { PostCreated } from '../events/PostCreated'
+```
+
 We can do any validation in the command handler before storing the event, for our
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -480,6 +480,12 @@ public handle(register: Register): void {
 }
 ```
 
+Remember to import the event class correctly on the top part of the file:
```
javiertoledo, line 483:
```suggestion
Remember to import the event class correctly on the top of the file:
```

docs/README.md
```diff
@@ -613,7 +619,12 @@ Booster applications provide you with a GraphQL API out of the box. You send com
 _mutations_ and get read models data using _queries_ or _subscriptions_. 
 
 In this section, we will be sending requests by hand using the online tool [Postwoman](https://postwoman.io/graphql), 
-which is free and includes great support for GraphQL. However, you can use any HTTP client you want.
+which is free and includes great support for GraphQL. However, you can use any HTTP client you want. In Postwoman
+online tool, you should enter the following URL as Endpoint:
```
javiertoledo, line 623:
```suggestion
which is free and includes great support for GraphQL. However, you can use any client you want. Your endpoint URL should look like this:
```

docs/README.md
```diff
@@ -613,7 +619,12 @@ Booster applications provide you with a GraphQL API out of the box. You send com
...
+
+```text
+https://<some random number>.execute-api.us-east-1.amazonaws.com/production/graphql
```
alvaroloes, line 626: This is specific to AWS. I think we should not use anything that is provider-specific if we can avoid it. 
We already talked about the `httpURL` output, so we can rewrite this like:
```suggestion
<httpURL>/graphql
```

davidverdu:
I agree with you. Your suggestions make sense.

docs/README.md
davidverdu: This make sense Javi

PR #310, opened 2020-08-04
[BOOST-925] Allow only single Role per User

rdoria1:
## Description
Allow only one role per user.

## Changes
After some conversations, we realized allowing a user to have multiple roles can be conflicting. Since roles have their one attributes, applying multiple roles to a single user with conflicting attributes values is misleading.

## Checks
- [X] Project Builds

## Additional information
Also ran integration tests.


rdoria1:
bot: integration

PR #333, opened 2020-08-07
Inject command object into handlers

NickSeagull:
**Note:** Changes should be actually

```diff
+31
-82
```

The rest are `yarn.lock` changes

## Description

In order to maintain API consistency with the rest of the user-implemented functions, we gonna make now the command handlers be a static function. 

## Changes

* Separated CommandInterface into CommandInterface and CommandClassInterface. 

* CommandInterface has no restrictions, so we making it `unknown`.

* CommandClassInterface restricts the static method to receive a Command

## Checks
- [X] Project Builds

### DIFF
--- a/packages/framework-core/src/booster-command-dispatcher.ts
+++ b/packages/framework-core/src/booster-command-dispatcher.ts
@@ -6,6 +6,7 @@ import {
   InvalidParameterError,
   NotAuthorizedError,
   NotFoundError,
+  CommandClassInterface,
 } from '@boostercloud/framework-types'
 import { BoosterAuth } from './booster-auth'
 import { RegisterHandler } from './booster-register-handler'
@@ -30,13 +31,13 @@ export class BoosterCommandDispatcher {
 
     const commandClass = commandMetadata.class
     this.logger.debug('Found the following command:', commandClass.name)
-    const command = new commandClass()
+    const command = commandClass as CommandClassInterface
     Object.assign(command, commandEnvelope.value)
     // TODO: Here we could call "command.validate()" so that the user can prevalidate
     // the command inputted by the user.
     const register = new Register(commandEnvelope.requestID, commandEnvelope.currentUser)
     this.logger.debug('Calling "handle" method on command: ', command)
-    await command.handle(register)
+    await command.handle(commandEnvelope.value, register)
     this.logger.debug('Command dispatched with register: ', register)
     await RegisterHandler.handle(this.config, this.logger, register)
   }
--- a/packages/framework-core/src/booster-command-dispatcher.ts
+++ b/packages/framework-core/src/booster-command-dispatcher.ts
@@ -32,12 +32,13 @@ export class BoosterCommandDispatcher {
     const commandClass = commandMetadata.class
     this.logger.debug('Found the following command:', commandClass.name)
     const command = commandClass as CommandClassInterface
-    Object.assign(command, commandEnvelope.value)
+    const commandInstance = new command()
+    Object.assign(commandInstance, commandEnvelope.value)
     // TODO: Here we could call "command.validate()" so that the user can prevalidate
     // the command inputted by the user.
     const register = new Register(commandEnvelope.requestID, commandEnvelope.currentUser)
     this.logger.debug('Calling "handle" method on command: ', command)
-    await command.handle(commandEnvelope.value, register)
+    await command.handle(commandInstance, register)
     this.logger.debug('Command dispatched with register: ', register)
     await RegisterHandler.handle(this.config, this.logger, register)
   }
--- a/packages/framework-core/src/booster-command-dispatcher.ts
+++ b/packages/framework-core/src/booster-command-dispatcher.ts
@@ -6,7 +6,7 @@ import {
   InvalidParameterError,
   NotAuthorizedError,
   NotFoundError,
-  CommandClassInterface,
+  CommandInterface,
 } from '@boostercloud/framework-types'
 import { BoosterAuth } from './booster-auth'
 import { RegisterHandler } from './booster-register-handler'
@@ -31,7 +31,7 @@ export class BoosterCommandDispatcher {
 
     const commandClass = commandMetadata.class
     this.logger.debug('Found the following command:', commandClass.name)
-    const command = commandClass as CommandClassInterface
+    const command = commandClass as CommandInterface
     const commandInstance = new command()
     Object.assign(commandInstance, commandEnvelope.value)
     // TODO: Here we could call "command.validate()" so that the user can prevalidate
--- a/packages/framework-core/src/decorators/command.ts
+++ b/packages/framework-core/src/decorators/command.ts
@@ -1,5 +1,5 @@
 import { Booster } from '../booster'
-import { CommandClassInterface, RoleAccess } from '@boostercloud/framework-types'
+import { CommandInterface, RoleAccess } from '@boostercloud/framework-types'
 import { getPropertiesMetadata } from './metadata'
 
 /**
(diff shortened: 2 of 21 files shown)


packages/framework-types/src/concepts/command.ts
```diff
@@ -2,16 +2,24 @@ import { Register } from './register'
...
+// The type correctness is ensured by the decorator, which ensures all of this.
+export interface CommandMetadata<TCommand = unknown> {
+  // For the class, we care that it has the static methods specified by CommandInterface
+  // and it has at least the properties of a class (like name, constructor, etc...)
+  // We don't care about the properties of the instance, so we set the type parameter of
+  // Class to unknown.
+  readonly class: CommandClassInterface<TCommand>
   readonly properties: Array<PropertyMetadata>
   readonly authorizedRoles: RoleAccess['authorize']
 }
```
javiertoledo, line 25: Great comments!

packages/framework-core/src/booster-command-dispatcher.ts
```diff
@@ -30,13 +31,13 @@ export class BoosterCommandDispatcher {
 
     const commandClass = commandMetadata.class
     this.logger.debug('Found the following command:', commandClass.name)
-    const command = new commandClass()
+    const command = commandClass as CommandClassInterface
     Object.assign(command, commandEnvelope.value)
```
alvaroloes, line 35: You are assigning the properties to the command class, instead of the command instance. So I don't think this will work as expected

packages/framework-core/src/booster-command-dispatcher.ts
```diff
@@ -30,13 +31,13 @@ export class BoosterCommandDispatcher {
...
     // TODO: Here we could call "command.validate()" so that the user can prevalidate
     // the command inputted by the user.
     const register = new Register(commandEnvelope.requestID, commandEnvelope.currentUser)
     this.logger.debug('Calling "handle" method on command: ', command)
-    await command.handle(register)
+    await command.handle(commandEnvelope.value, register)
```
alvaroloes, line 40: This won't work as expected, as we are not passing an instance of the command, but a plain object with its attributes, so if the user has some methods defined in its command class, it will fail. We should create the instance as we were doing, and pass it here

packages/framework-integration-tests/src/commands/MoveStock.ts
```diff
@@ -16,22 +16,26 @@ export class MoveStock {
     readonly quantity: number
   ) {}
 
-  public async handle(register: Register): Promise<void> {
-    const stock = await Booster.fetchEntitySnapshot(Stock, this.productID)
-    if (this.enoughStock(stock)) {
-      register.events(new StockMoved(this.productID, this.origin, this.destination, this.quantity))
+  public static async handle(command: MoveStock, register: Register): Promise<void> {
+    const stock = await Booster.fetchEntitySnapshot(Stock, command.productID)
+    if (MoveStock.enoughStock(command, stock)) {
```
alvaroloes, line 21: I think it is ok (and easier to read( to have the method defined in the instance of the command, as you did in the documentation.

packages/framework-types/src/concepts/command.ts
```diff
@@ -2,16 +2,24 @@ import { Register } from './register'
...
 
-/**
- * A `Command` is an imperative intent in your system.
- * All Command classes of your application must extend this class.
- */
-export interface CommandInterface {
-  handle(register: Register): Promise<void>
+export type CommandInterface = unknown
+
+export interface CommandClassInterface<TCommand = unknown> extends Class<TCommand> {
```
alvaroloes, line 7: I think you don't need this extra `CommandClassInterface` and you can leave the `CommandInterface` class as it was before (with the change of `handle` receiving the command as first argument). 
The key here is to apply the restriction of `CommandInterface` to the class, instead of the instance. See this comment LINK to see how it is applied.

packages/framework-core/src/decorators/command.ts
```diff
@@ -1,13 +1,13 @@
...
+import { CommandClassInterface, RoleAccess } from '@boostercloud/framework-types'
 import { getPropertiesMetadata } from './metadata'
 
 /**
  * Annotation to tell Booster which classes are your entities
  * @param attributes
  * @constructor
  */
-export function Command(attributes: RoleAccess): (commandClass: Class<CommandInterface>) => void {
+export function Command(attributes: RoleAccess): <TCommand>(commandClass: CommandClassInterface<TCommand>) => void {
```
alvaroloes, line 10: Related to this comment LINK

If you change this to: 
```suggestion
export function Command(attributes: RoleAccess): (commandClass: CommandInterface) => void {
```
It will work as expected and there is no need for the `CommandClassInterface`  😃

NickSeagull: I don't think so, because we are expecting that the class itself (not the instances) to have that interface, so the passed object must be of that type. Check out https://stackoverflow.com/questions/13955157/how-to-define-static-property-in-typescript-interface

packages/framework-types/src/concepts/command.ts
NickSeagull: I don't see any links @alvaroloes 🤔

packages/framework-core/src/booster-command-dispatcher.ts
NickSeagull: Fixed https://github.com/boostercloud/booster/pull/333/commits/0705b5cf55a9a88872dbaebfc866e4f149737501 , thanks!

packages/framework-integration-tests/src/commands/MoveStock.ts
NickSeagull: Fixed https://github.com/boostercloud/booster/pull/333/commits/a2c62c156f2da88ede722320c3a6b942bfa0685d thanks

packages/framework-types/src/concepts/command.ts
alvaroloes: Updated!

packages/framework-core/src/decorators/command.ts
alvaroloes: Have you tried it? Check this [playground](https://www.typescriptlang.org/play?#code/FASwdgLgpgTgZgQwMZQAQGEA2CDOOA8AKgEpRyZRLQAmAfKgN7CqphQDuqAFAHR8IwA5jgBcqBGACeAbQC6ASjEkyFKlGrAAvsGAB6XagC0xk6bPmLJnUmx5UAdRAQAFgGUICCCCQBZKC4B7akZmFlQABwBXACNMb1QcDy8kVGcJagouRVQANwCQYKYw4tQkALAcAIoeTADBLgByRM9vBvlQlm1tYBtcHAcnZwBJCo8wFD9Awo6ImLiUtLAMqCyxPIKQkrCyiqqoGrrG8GbxqDaZrp1waHhkNAAJdIoYTbDF5dXc-I1u-SNLAGA0w6OCRcZecqoSI4KA4dwtJCPJbPLjvZ5iJHLGDyV4sNGwHj4rJaEFgqggSHQ2FYPr2ZwBGEjE4oIY4TEo-EwMQ0vD4dmwWg4ophNicTlZQlPFbtbo6KlwpLefkwLiOFzw5KTenUdp-eUapVSlVq4ajCQTfzanEAEQCsLADQgpQCAFtwiAKHKYTgeTg6QyoEyxiy2UbVYMg+aoFqgrqDPLff7GWbTqzleH1YrfJbY6hbfbHc63R6oEA)

NickSeagull: Doh! I didn't reallize that I duplicated the responsibility of `CommandInterface` and renamed it. Now I understood, thanks for pointing it out. Fixed in https://github.com/boostercloud/booster/pull/333/commits/47bb8bfdefe0210d9e7efe72065084d714e5ab58

packages/framework-types/src/concepts/command.ts
NickSeagull: Fixed in https://github.com/boostercloud/booster/pull/333/commits/47bb8bfdefe0210d9e7efe72065084d714e5ab58

NickSeagull:
bot: integration

NickSeagull:
Integration tests pass locally, so I'll merge this :)

PR #373, opened 2020-08-14
Fix authorization token for websockets and apollo clients

alvaroloes:
## Description
After a deep investigation of the Apollo clients, we realized that the way we were using to send the authentication token through WebSockets was not standard and it didn't work in the Apollo clients of strongly typed languages like Java or iOS.

This PR uses the official way to send tokens, using the `connectionsParams` that are sent after the socket connection has been established.

These are the steps followed when the client needs an authenticated WebSocket connection (they are the ones dictated by the ["GraphQL over WebSocket protocol"](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md) plus some details of what it is done in this PR):

1. Client connects to the WebSocket using the Apollo Subscriptions Client. It specifies the `Authorization` field with the auth token in the `connectionsParams` fileld:
```typesctipt
new SubscriptionClient(`<websocket URL>`, {
  connectionParams: {
    Authorization: 'Bearer ' + token,
  }
})
```
2. When Apollo client establishes the WebSocket connection, it will send a message of type GQL_CONNECTION_INIT with those `connectionParams`
3. On the server-side, in the `GraphQLWebsocketHandler` class, we will get the associated user for that token, and store it in the new connections store.
4. Now, on every message we receive over that WebSocket connection, we will first get user data from the connections store and add it to the GraphQLRequestEnvelope. See `GraphQLWebsocketHandler.handleStart` method.
5. The message handling follows as usual, checking permissions, etc.
6. When the socket is disconnected or when we receive a GQL_TERMINATE message, we delete the user data associated to the terminating connection. 

## Checks
- [x] Project Builds
- [x] Updated documentation accordingly -> In progress, see draft PR: https://github.com/boostercloud/booster/pull/375


alvaroloes:
bot: integration

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -28,36 +29,38 @@ export class BoosterGraphQLDispatcher {
 
     this.graphQLSchema = new GraphQLGenerator(config, commandDispatcher, this.readModelDispatcher).generateSchema()
     this.websocketHandler = new GraphQLWebsocketHandler(
+      config,
       logger,
-      this.config.provider.readModels.notifySubscription.bind(null, config),
+      this.config.provider.auth,
+      this.config.provider.connections,
```
alvaroloes, line 35: Now we are passing the `auth` related functions to the GraphQLWebSocketHandler so that it is able to retrieve a user from a token. 

Also, we are passing the new `connections` related function to be able to store, fetch and delete connection data (user information)

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -28,36 +29,38 @@ export class BoosterGraphQLDispatcher {
...
+        onTerminate: this.handleDisconnect.bind(this),
       }
     )
   }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   public async dispatch(request: any): Promise<any> {
-    const envelope = await this.config.provider.graphQL.rawToEnvelope(request, this.logger)
-    this.logger.debug('Received the following GraphQL envelope: ', envelope)
+    const envelopeOrError = await this.config.provider.graphQL.rawToEnvelope(request, this.logger)
```
alvaroloes, line 46: I will explain this `envelope or error` thingy below, in the `graphql-atapter.ts` file

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -150,13 +156,15 @@ export class BoosterGraphQLDispatcher {
...
+  private async handleDisconnect(connectionID?: string): Promise<void> {
+    if (!connectionID) {
       // This should be impossible, but just in case
       this.logger.debug("Received a DISCONNECT message but field 'connectionID' is missing. Doing nothing")
       return
     }
-    return this.readModelDispatcher.unsubscribeAll(envelope.connectionID)
+    this.logger.debug('Deleting all subscriptions and connection data')
+    await this.config.provider.connections.deleteData(this.config, connectionID)
+    await this.readModelDispatcher.unsubscribeAll(connectionID)
```
alvaroloes, line 167: Now, when the socket disconnects, we delete bot the subscriptions and the connection data

packages/framework-core/src/booster-read-model-dispatcher.ts
```diff
@@ -13,9 +13,6 @@ import {
 import { BoosterAuth } from './booster-auth'
 import { Booster } from './booster'
 
-// TODO: Think if this should be configurable
-const subscriptionDurationSeconds = 24 * 60 * 60 // 24 hours
```
alvaroloes, line 17: This was finally moved to the Booster config, so it is configurable by the user now (it has sensible defaults)

packages/framework-core/src/services/graphql/websocket-protocol/graphql-websocket-protocol.ts
```diff
@@ -39,34 +47,48 @@ export class GraphQLWebsocketHandler {
...
       }
     } catch (e) {
       this.logger.error(e)
-      await this.sendToConnection(envelope.connectionID, new GraphQLInitError(e.message))
+      await this.connectionManager.sendMessage(this.config, envelope.connectionID, new GraphQLInitError(e.message))
     }
   }
 
-  private async handleInit(connectionID: string): Promise<void> {
+  private async handleInit(connectionID: string, clientMessage: GraphQLInit): Promise<void> {
```
alvaroloes, line 78: Here is where we manage the first message the Apollo clients sends with the `connectionParams` payload.

We check if there is an auth token, fetch the corresponding user, and store it in the connections store.

packages/framework-core/src/services/graphql/websocket-protocol/graphql-websocket-protocol.ts
```diff
@@ -39,34 +47,48 @@ export class GraphQLWebsocketHandler {
...
+      user: userEnvelope,
+    }
+    this.logger.debug('Storing connection data: ', connectionData)
+    await this.connectionManager.storeData(this.config, connectionID, connectionData)
     this.logger.debug('Sending ACK')
-    await this.sendToConnection(connectionID, new GraphQLInitAck())
+    await this.connectionManager.sendMessage(this.config, connectionID, new GraphQLInitAck())
   }
 
   private async handleStart(
```
alvaroloes, line 94: This method is called on every message in the socket. We need to get the user data from the connections store and set it in the envelope we are building to start the GraphQL operation. Check the `augmentEnvelope` method

packages/framework-core/src/services/graphql/websocket-protocol/graphql-websocket-protocol.ts
```diff
@@ -104,8 +121,25 @@ export class GraphQLWebsocketHandler {
...
+    await this.connectionManager.sendMessage(this.config, connectionID, new GraphQLData(message.id, result))
+    await this.connectionManager.sendMessage(this.config, connectionID, new GraphQLComplete(message.id))
+  }
+
+  private async augmentEnvelope(
+    connectionID: string,
+    envelope: GraphQLRequestEnvelope,
+    message: GraphQLStart
+  ): Promise<GraphQLRequestEnvelope> {
+    const connectionData = await this.connectionManager.fetchData(this.config, connectionID)
```
alvaroloes, line 133: Here we fetch the connection data associated to this socket connection. Then we set the user information, that is among the connection data, to the envelope

packages/framework-integration-tests/integration/providers/aws/utils.ts
```diff
@@ -11,9 +11,9 @@ import { internet } from 'faker'
 import { sleep } from '../helpers'
```
alvaroloes, line 11: I changed the way we are creating the Apollo clients so that we can now send the token as expected and, besides, we are able to refresh the token

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/graphql-stack.ts
```diff
@@ -151,4 +153,17 @@ export class GraphQLStack {
     })
     return table
   }
+
+  private buildConnectionsTable(): Table {
+    return new Table(this.stack, this.config.resourceNames.connectionsStore, {
```
alvaroloes, line 158: We just build an extra DynamoD table with the socket connection ID as key

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/static-website-stack.ts
```diff
@@ -1,34 +1,31 @@
-import {CfnOutput, RemovalPolicy, Stack} from '@aws-cdk/core'
-import {Bucket} from '@aws-cdk/aws-s3'
-import {BucketDeployment, Source} from '@aws-cdk/aws-s3-deployment'
+import { CfnOutput, RemovalPolicy, Stack } from '@aws-cdk/core'
```
alvaroloes, line 1: Forget this file, the linter auto-formated it, but I didn't change anything

packages/framework-provider-aws/src/index.ts
```diff
@@ -51,12 +55,19 @@ export const Provider: ProviderLibrary = {
...
   api: {
     requestSucceeded,
     requestFailed,
   },
+  connections: {
+    storeData: storeConnectionData.bind(null, dynamoDB),
+    fetchData: fetchConnectionData.bind(null, dynamoDB),
+    deleteData: deleteConnectionData.bind(null, dynamoDB),
+    sendMessage: sendMessageToConnection,
+  },
```
alvaroloes, line 70: This is the new namespace `connections` with functions to deal with... wair for it... connections!

packages/framework-provider-aws/src/library/graphql-adapter.ts
```diff
@@ -1,26 +1,36 @@
...
+import { GraphQLRequestEnvelope, GraphQLRequestEnvelopeError, Logger } from '@boostercloud/framework-types'
 import { CognitoIdentityServiceProvider } from 'aws-sdk'
+import { userEnvelopeFromAuthToken } from './auth-adapter'
 
 export async function rawGraphQLRequestToEnvelope(
   userPool: CognitoIdentityServiceProvider,
   request: APIGatewayProxyEvent,
   logger: Logger
-): Promise<GraphQLRequestEnvelope> {
+): Promise<GraphQLRequestEnvelope | GraphQLRequestEnvelopeError> {
```
alvaroloes, line 10: So what happened here:

This function is called by the core when a GraphQL request comes either through HTTP or WebSockets. 
When everything goes correctly, it returns the GraphQL request envelope with information to know the connection ID of the socket if it is the case. 
When something goes wrong like, for example, the token supplied by the user is expired and the function `userEnvelopeFromAuthToken` throws an error, we were just bubbling up the error. But that's wrong, as we also need to know the connectionID of the socket so that we can send the error back to the client, and that's what we are doing here now (we are also including more information)

The reason why returning a union type (Correct Result | Error result) it simplifies a lot the logic in the core. I tried to throw a custom error with the connectionID as a field, but the code in the core ended-up spaghettified (the reasons were that when you catch an error, you can't specify its type and that this error is happening before the core knows the structure of the request, complicating everything)

alvaroloes: This fixes several problems. Now, on the client side, you can properly know when the auth token is expired and refresh it, as you will get an expected error. Before, the client was getting a "network error" and the GraphQL client engine failed.

packages/framework-provider-aws/src/library/subscription-adapter.ts
```diff
@@ -64,21 +64,6 @@ export async function fetchSubscriptions(
   return result.Items as Array<SubscriptionEnvelope>
 }
 
-export async function notifySubscription(
```
alvaroloes, line 67: Just removed this function form here, as it now belongs to the `connections-adatper.ts` file

packages/framework-provider-aws/src/library/user-envelopes.ts
```diff
@@ -29,21 +28,3 @@ export class UserEnvelopeBuilder {
     return attributeMap
   }
 }
-
-type CouldHaveHeaders = { headers: { [name: string]: string } | null }
-export async function fetchUserFromRequest(
```
alvaroloes, line 34: This function was moved to `auth-adapter.ts`

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -28,36 +29,38 @@ export class BoosterGraphQLDispatcher {
...
         onStartOperation: this.runGraphQLOperation.bind(this),
         onStopOperation: this.readModelDispatcher.unsubscribe.bind(this.readModelDispatcher),
-        onTerminateOperations: this.readModelDispatcher.unsubscribeAll.bind(this.readModelDispatcher),
+        onTerminate: this.handleDisconnect.bind(this),
       }
     )
   }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   public async dispatch(request: any): Promise<any> {
```
NickSeagull, line 45: Maybe this is out of scope for this PR, but could you replace these `any` with `unknown`?

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -150,13 +156,15 @@ export class BoosterGraphQLDispatcher {
     return result
   }
 
-  private async handleDisconnect(envelope: GraphQLRequestEnvelope): Promise<void> {
-    if (!envelope.connectionID) {
+  private async handleDisconnect(connectionID?: string): Promise<void> {
+    if (!connectionID) {
       // This should be impossible, but just in case
       this.logger.debug("Received a DISCONNECT message but field 'connectionID' is missing. Doing nothing")
```
NickSeagull, line 162: Perhaps this should be `this.logger.error`?

packages/framework-core/src/services/graphql/websocket-protocol/graphql-websocket-protocol.ts
```diff
@@ -39,34 +47,48 @@ export class GraphQLWebsocketHandler {
...
     } catch (e) {
       this.logger.error(e)
-      await this.sendToConnection(envelope.connectionID, new GraphQLInitError(e.message))
+      await this.connectionManager.sendMessage(this.config, envelope.connectionID, new GraphQLInitError(e.message))
     }
   }
 
-  private async handleInit(connectionID: string): Promise<void> {
+  private async handleInit(connectionID: string, clientMessage: GraphQLInit): Promise<void> {
+    let userEnvelope: UserEnvelope | undefined
```
NickSeagull, line 79: If its a `let` no need to specify `undefined`

```suggestion
    let userEnvelope: UserEnvelope
```

packages/framework-core/test/services/graphql/websocket-protocol/graphql-websocket-protocol.test.ts
```diff
@@ -71,6 +89,30 @@ describe('the `GraphQLWebsocketHandler`', () => {
         envelope.connectionID = random.alphaNumeric(10)
       })
 
+      describe('with an error in the envelope', () => {
+        const errorMessage = lorem.sentences(1)
+        let erroneousEnvelope: GraphQLRequestEnvelopeError
+        beforeEach(() => {
+          erroneousEnvelope = {
```
NickSeagull, line 96: Perhaps

```suggestion
          envelopeWithError = {
```

packages/framework-integration-tests/integration/providers/aws/subscriptions.integration.ts
```diff
@@ -1,12 +1,19 @@
...
 import { random } from 'faker'
 import gql from 'graphql-tag'
+import { expect } from 'chai'
 import * as chai from 'chai'
 import { Observable } from 'apollo-client/util/Observable'
 
 chai.use(require('chai-as-promised'))
 
-describe('subscriptions', () => {
+describe.only('subscriptions', () => {
```
NickSeagull, line 16: What does this mean?

packages/framework-integration-tests/integration/providers/aws/subscriptions.integration.ts
```diff
@@ -63,6 +70,50 @@ describe('subscriptions', () => {
...
+      await waitForIt(countSubscriptionsItems, (newCount) => newCount == originalSubscriptionsCount + 2)
+      // Check we receive data when the read model is modified
+      await cartMutation(client, cartID)
+      await expect(Promise.all([promisify(observableOne), promisify(observableTwo)])).to.eventually.be.fulfilled
+
+      // Now reconnect and see if we keep the having the same subscription and receive data
+      await client.reconnect()
+      await waitForIt(countSubscriptionsItems, (newCount) => newCount == originalSubscriptionsCount + 2)
+      await cartMutation(client, cartID)
+      await expect(Promise.all([promisify(observableOne), promisify(observableTwo)])).to.eventually.be.fulfilled
```
NickSeagull, line 115: Might not mean much, but when I look at these tests I feel a lot of joy

packages/framework-integration-tests/integration/providers/aws/utils.ts
```diff
@@ -400,16 +415,12 @@ export async function queryEvents(primaryKey: string, latestFirst = true): Promi
...
-export async function subscriptionsTableName(): Promise<string> {
-  const stackName = appStackName()
-
-  return `${stackName}-subscriptions-store`
-}
-
 export async function countSubscriptionsItems(): Promise<number> {
-  const tableName = await subscriptionsTableName()
+  return countTableItems(`${appStackName()}-subscriptions-store`)
+}
```
NickSeagull, line 420: No need to calculate on each call, this can be a const
```suggestion
const countSubscriptionsItems = await countTableItems(`${appStackName()}-subscriptions-store`)
```

packages/framework-integration-tests/integration/providers/aws/utils.ts
```diff
@@ -400,16 +415,12 @@ export async function queryEvents(primaryKey: string, latestFirst = true): Promi
...
 
-  return countTableItems(tableName)
+export async function countConnectionsItems(): Promise<number> {
+  return countTableItems(`${appStackName()}-connections-store`)
```
NickSeagull, line 423: Same here

packages/framework-provider-azure/src/index.ts
```diff
@@ -50,15 +48,24 @@ export const Provider: ProviderLibrary = {
...
   // ProviderAPIHandling
   api: {
     requestSucceeded,
     requestFailed,
   },
+  connections: {
+    storeData: notImplemented as any,
+    fetchData: notImplemented as any,
+    deleteData: notImplemented as any,
+    sendMessage: notImplemented as any,
```
NickSeagull, line 62: I think its better to leave it as `undefined` instead of creating a new function

packages/framework-provider-kubernetes/src/index.ts
```diff
@@ -29,13 +28,22 @@ export const Provider: ProviderLibrary = {
...
   // ProviderAPIHandling
   api: {
     requestSucceeded: undefined as any,
     requestFailed: undefined as any,
   },
+  connections: {
+    storeData: notImplemented as any,
+    fetchData: notImplemented as any,
+    deleteData: notImplemented as any,
+    sendMessage: notImplemented as any,
```
NickSeagull, line 42: I think it's better to leave as undefined

packages/framework-provider-local/src/index.ts
```diff
@@ -57,13 +55,22 @@ export const Provider: ProviderLibrary = {
...
   // ProviderAPIHandling
   api: {
     requestSucceeded,
     requestFailed,
   },
+  connections: {
+    storeData: notImplemented as any,
+    fetchData: notImplemented as any,
+    deleteData: notImplemented as any,
+    sendMessage: notImplemented as any,
```
NickSeagull, line 69: Better leave as undefined

packages/framework-types/src/envelope.ts
```diff
@@ -49,6 +49,9 @@ export interface GraphQLRequestEnvelope extends Envelope {
   connectionID?: string
   value?: GraphQLOperation | GraphQLClientMessage
 }
+export type GraphQLRequestEnvelopeError = Pick<GraphQLRequestEnvelope, 'eventType' | 'connectionID' | 'requestID'> & {
+  error: Error
+}
```
NickSeagull, line 54: Lovely type level programming!

packages/framework-types/src/envelope.ts
```diff
@@ -64,6 +67,11 @@ export interface GraphQLOperation {
   variables?: Record<string, any>
 }
 
+export interface ConnectionDataEnvelope {
+  expirationTime: number // In Epoch format
```
NickSeagull, line 71: Epoch is not a format, but a day

```suggestion
  expirationTime: number // Since Unix epoch
```

packages/framework-core/src/booster-graphql-dispatcher.ts
alvaroloes: Yeah! this makes total sense because we do nothing with this type. Thanks, I'll see if I can change more any's by unknown

alvaroloes: I took the liberty and replaced a lot of any's by unknown in the core package 😎
Check 1c063b7602e4a54e5510d82ed5d531c2eed8fe32

alvaroloes: And this 418308b1b7e182cbc60aa08ee79fdb7b36ffae39

packages/framework-core/src/booster-graphql-dispatcher.ts
alvaroloes: Well, it is a very rare situation (I don't think it could happen) and if it happens, nothing wrong will occur. So I think it is ok to just print it. But! I think it makes sense to change it to `this.logger.info` so that it is always printed, not only in debug mode.

alvaroloes: Done here: c48ee3469cf108744a2abbf9faa0ed22910b3022

packages/framework-core/src/services/graphql/websocket-protocol/graphql-websocket-protocol.ts
alvaroloes: If I don't say that `userEnvelope` can be `undefined`, then I won't be able to assign undefined to it and the code will not compile.
To my knowledge, `let` has nothing to do with the type of the variable, but with its scope and mutability

packages/framework-core/test/services/graphql/websocket-protocol/graphql-websocket-protocol.test.ts
alvaroloes: Much better yep. Done: e20cc994db4d2e34abae96c9326061be98435056

packages/framework-integration-tests/integration/providers/aws/subscriptions.integration.ts
alvaroloes: Woah!! This passed unnoticed, I will remove it. 

If you put this, when you run the test suite, only this test will be executed. It is ideal for developing but really dangerous.

alvaroloes: Done: 170da5782af35d0fea0527c1b4cba14aee93bbbb

packages/framework-integration-tests/integration/providers/aws/subscriptions.integration.ts
alvaroloes: That's the joy of the elder developers that lies very deep in our hearts

packages/framework-integration-tests/integration/providers/aws/utils.ts
alvaroloes: @NickSeagull mmm I think there is something I'm missing. Each time I call the function `countSubscriptionItems` I need to do the call and get the new count.

I don't understand the suggestion either

packages/framework-integration-tests/integration/providers/aws/utils.ts
alvaroloes: check: https://github.com/boostercloud/booster/pull/373/files/d4211f84a9c2f7dc06a2dd3d4a1f482351518785#r473963791

packages/framework-provider-azure/src/index.ts
alvaroloes: I can't. These functions are used in some tests like this: `storeData.bind(this)`. If they are `undefined` it will fail

packages/framework-provider-kubernetes/src/index.ts
alvaroloes: Check https://github.com/boostercloud/booster/pull/373#discussion_r473965987

packages/framework-provider-local/src/index.ts
alvaroloes: Check https://github.com/boostercloud/booster/pull/373#discussion_r473965987

packages/framework-types/src/envelope.ts
alvaroloes: In the DynamoDB documentation, they refer to it as a format, [here](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-before-you-start.html#time-to-live-ttl-before-you-start-formatting)

Also, I think that when someone says `epoch format`, it is understood as "the number of seconds since epoch". What do you think?

packages/framework-core/src/services/graphql/websocket-protocol/graphql-websocket-protocol.ts
NickSeagull: Oh that's weird. In theory, at the type level `let` acts as an alias to a type guard that adds `undefined` to the type of the reference.

That is because when you define `let` at the first time, it will be `undefined`. BUT! If you are willing to assign `undefined` by hand it might not work I guess. Thanks!

packages/framework-provider-azure/src/index.ts
NickSeagull: Gah! :(

packages/framework-types/src/envelope.ts
NickSeagull: I see, Wikipedia says that people sometimes refer to it as "Epoch Time" which can be seen as a format. Leave it like this! :) 

Source https://en.wikipedia.org/wiki/Unix_time

NickSeagull (approved):
LGTM

packages/framework-provider-aws/src/library/auth-adapter.ts
```diff
@@ -1,11 +1,25 @@
...
+export async function userEnvelopeFromAuthToken(
+  userPool: CognitoIdentityServiceProvider,
+  token: string | undefined
+): Promise<UserEnvelope | undefined> {
+  if (!token) {
+    return undefined
+  }
+  const accessToken = token.replace('Bearer ', '')
+  const currentUserData = await userPool.getUser({ AccessToken: accessToken }).promise()
+  // TODO: Check the errors for when the token is expired, etc. and throw with corresponding types and then use HTTP status codes
```
itrion, line 19: @alvaroloes Is there a ticket for this TODO?

packages/framework-provider-aws/test/library/auth-adapter.test.ts
```diff
@@ -30,6 +36,52 @@ describe('the auth-adapter', () => {
...
+
+      replace(
+        userPool,
+        'getUser',
+        fake.returns({
+          promise: fake.resolves({
+            UserAttributes: userReturnedFromPool,
+          }),
+        })
+      )
```
itrion, line 76: I think we could also create a stub instance
```
var stub = createStubInstance(CognitoIdentityServiceProvider, {
    getUser: fake.returns({
        promise: fake.resolves({
            UserAttributes: userReturnedFromPool,
        }),
    })
});
```

javiertoledo:
bot:integration

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -28,36 +29,37 @@ export class BoosterGraphQLDispatcher {
...
       default:
         return this.config.provider.graphQL.handleResult({
-          errors: [new Error(`Unknown message type ${envelope.eventType}`)],
+          errors: [new Error(`Unknown message type ${envelopeOrError.eventType}`)],
         })
     }
   }
 
-  private async handleMessage(envelope: GraphQLRequestEnvelope): Promise<DispatchResult> {
+  private async handleMessage(envelope: GraphQLRequestEnvelope | GraphQLRequestEnvelopeError): Promise<DispatchResult> {
```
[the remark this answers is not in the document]
javiertoledo: I'd have used a promise.

packages/framework-provider-aws/src/library/read-model-adapter.ts
```diff
@@ -20,9 +20,10 @@ export async function fetchReadModel(
   readModelName: string,
   readModelID: UUID
 ): Promise<ReadModelInterface> {
-  const params = {
+  const params: DynamoDB.DocumentClient.GetItemInput = {
     TableName: config.resourceNames.forReadModel(readModelName),
     Key: { id: readModelID },
+    ConsistentRead: true,
```
javiertoledo, line 26: I don't think we should use consistent reads for read models. This makes reading the read models slower, and they're basically caches, so it should not matter if you read the previous version of it, everything else should be designed to allow read models to be eventually consistent

```suggestion
```

javiertoledo (approved):
Approving to be able to merge this PR, but we should make sure that we review the usages of consistent reads in DynamoDB (Created a new ticket for that https://github.com/boostercloud/booster/issues/397)

javiertoledo:
bot: integration

javiertoledo:
bot: integration

javiertoledo:
Integration tests are not passing, I'm investigating...

packages/framework-core/src/booster-graphql-dispatcher.ts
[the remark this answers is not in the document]
NickSeagull: @itrion Yes, Either is what fits better this place, specially because you can pattern match on it by using [fold](https://gcanti.github.io/fp-ts/modules/Either.ts#fold), but sadly, the team is not comfortable yet using `fp-ts`

@javiertoledo in my opinion, the Promise solution is not very good as it is encoding incorrectly the domain into the type level. A `Promise` denotes execution, and here we're only passing a pure value. Also by using a promise, we are undercovering the intent of the function, which is handing a regular envelope or an envelope containing an error.

packages/framework-core/test/services/graphql/websocket-protocol/graphql-websocket-protocol.test.ts
```diff
@@ -71,6 +90,30 @@ describe('the `GraphQLWebsocketHandler`', () => {
...
+        beforeEach(() => {
+          envelopeWithError = {
+            ...envelope,
+            error: new Error(errorMessage),
+          }
+        })
+
+        it('sends the error to the client', async () => {
+          resultPromise = websocketHandler.handle(envelopeWithError)
+          await resultPromise
```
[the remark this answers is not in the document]
NickSeagull: In fact this could've been

```ts
        it('sends the error to the client', () => websocketHandler.handle(envelopeWithError))
```

packages/framework-provider-aws/src/library/auth-adapter.ts
NickSeagull: Now there is 😃  https://github.com/boostercloud/booster/issues/401

javiertoledo:
bot: integration

javiertoledo:
bot: integration

javiertoledo:
bot: integration

javiertoledo:
I ran the integration tests locally. They took over half an hour and succeeded, but they're throwing a lot of debug messages that I don't remember to have seen before and I can't make them pass in Github actions. I wonder if they're timing out or there's some other kind of misconfiguration. I've created issue #402 to split the different module integration tests and try troubleshooting any issues, my feeling is that they're just "too big". We're even deploying the sample app twice to AWS!

I don't think we should keep this PR stuck until that other thing is resolved, so I'll just merge it, it seems that integration tests are failing to finish in all branches anyway, so I doubt that the problem is in this specific PR.

.github/workflows/integration-tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
         runs-on: [ubuntu-latest]
       fail-fast: false
 
-    runs-on: ${{ matrix.runs-on }}
+    runs-on: ${{ matrix.os }}
```
javiertoledo, line 14: The linter highlighted this as an error and suggested this change, have I broken anything?

packages/cli/src/services/config-service.ts
```diff
@@ -14,9 +13,7 @@ export async function compileProjectAndLoadConfig(): Promise<BoosterConfig> {
 
 async function compileProject(projectPath: string): Promise<void> {
   try {
-    await withinWorkingDirectory(projectPath, () => {
-      return exec('npm run compile')
-    })
+    await exec('npm run compile', { cwd: projectPath })
```
javiertoledo, line 16: We were doing some weird stuff to run things in a different folder...

packages/cli/src/templates/project/package-json.ts
```diff
@@ -1,15 +1,18 @@
...
   "dependencies": {
-    "@boostercloud/framework-core": "^0.5.0",
-    "@boostercloud/framework-types": "^0.5.0",
+    "@boostercloud/framework-core": "^${VERSION}",
+    "@boostercloud/framework-types": "^${VERSION}",
     "{{{providerPackageName}}}": "*"
   },
   "devDependencies": {
-    "@boostercloud/cli": "^0.5.0",
+    "@boostercloud/cli": "^${VERSION}",
```
javiertoledo, line 15: Not the versions in the project template are dynamically filled with the current version of the CLI.

packages/framework-integration-tests/integration/cli/cli.project.integration.ts
```diff
@@ -117,52 +128,54 @@ describe('Project', () => {
...
-    const expectedCartDemoTsConfigEslint = await readFileContent('integration/fixtures/cart-demo/tsconfig.eslint.json')
-    const cartDemoTsConfigEslintContent = await readFileContent(CART_DEMO_TS_CONFIG_ESLINT)
+    const expectedCartDemoTsConfigEslint = readFileContent('integration/fixtures/cart-demo/tsconfig.eslint.json')
+    const cartDemoTsConfigEslintContent = readFileContent(CART_DEMO_TS_CONFIG_ESLINT)
     expect(cartDemoTsConfigEslintContent).to.equal(expectedCartDemoTsConfigEslint)
 
-    const expectedCartDemoTsConfig = await readFileContent('integration/fixtures/cart-demo/tsconfig.json')
-    const cartDemoTsConfigContent = await readFileContent(CART_DEMO_TS_CONFIG)
+    const expectedCartDemoTsConfig = readFileContent('integration/fixtures/cart-demo/tsconfig.json')
+    const cartDemoTsConfigContent = readFileContent(CART_DEMO_TS_CONFIG)
```
javiertoledo, line 174: Lots of awaits in non-async methods, I guess it was just garbage from an old refactor.

packages/framework-integration-tests/integration/cli/cli.project.integration.ts
```diff
@@ -117,52 +128,54 @@ describe('Project', () => {
...
     expect(cartDemoTsConfigContent).to.equal(expectedCartDemoTsConfig)
 
-    process.chdir(projectName)
-    await exec('npm run compile')
-    await exec('npm run lint:check')
-    process.chdir('..')
+    await expect(exec('npm run compile', { cwd: projectName })).to.be.eventually.fulfilled
+    await expect(exec('npm run lint:check', { cwd: projectName })).to.be.eventually.fulfilled
```
javiertoledo, line 178: Fun thing here: when one of these execs failed, the `process.chdir('..')` was skipped, and it made the next tests to write more files in the wrong dir, making the execs fail again in a complete hell of recursive failures that ended up with a ton of new files spread all over the place, and making the final recursive rmdirs fail. Thanks God we have git or I would have been still deleting files by hand 🤣

packages/framework-integration-tests/integration/cli/setup.ts
```diff
@@ -22,19 +22,16 @@ const testFiles: Array<string> = [
...
 
 before(async () => {
   const integrationTestsPackageRoot = path.dirname(__dirname)
-  process.chdir(integrationTestsPackageRoot)
 
-  await exec('lerna bootstrap')
-  await exec('lerna clean --yes')
-  await exec('lerna run clean --stream')
-
-  process.chdir('..')
```
javiertoledo, line 37: `chdir` is evil

packages/framework-integration-tests/integration/helper/fileHelper.ts
```diff
@@ -1,23 +1,15 @@
...
-export const writeFileContent = (filePath: string, data: any): void => fs.writeFileSync(filePath, data)
+export const writeFileContent = (filePath: string, data: unknown): void => fs.writeFileSync(filePath, data)
 
-export const removeFiles = (filePaths: Array<string>): Array<Promise<void>> => {
-  return filePaths.map((file: string) => {
-    return new Promise((resolve) => {
-      fs.unlinkSync(file)
-      resolve()
-    })
-  })
```
javiertoledo, line 13: Promises might be confusing 🤣

packages/framework-integration-tests/package.json
```diff
@@ -48,7 +48,7 @@
     "fix-lint": "eslint --quiet --fix --ext '.js,.ts' **/*.ts",
     "compile": "tsc -b tsconfig.json",
     "clean": "rimraf ./dist tsconfig.tsbuildinfo",
-    "integration": "FULL_INTEGRATION_TEST=true yarn integration/aws && yarn integration/local && yarn integration/cli && yarn integration/end-to-end",
+    "integration": "FULL_INTEGRATION_TEST=true yarn integration/local && yarn integration/cli && yarn integration/aws && yarn integration/end-to-end",
```
javiertoledo, line 51: It's better to make the tests that run locally to fail earlier. We were passing the whole AWS suite just to hit the error in the CLI before, it sucked!

packages/framework-core/src/booster-graphql-dispatcher.ts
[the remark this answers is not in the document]
alvaroloes: Explanation of why I didn't use a promise here: https://github.com/boostercloud/booster/pull/373#discussion_r471482391

I tried a lot of things (even creating a specific error) but all of them were messier than this ;-)

packages/framework-core/test/services/graphql/websocket-protocol/graphql-websocket-protocol.test.ts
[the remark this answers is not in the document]
alvaroloes: @itrion Yes, the reason is that the `websocketHandler.handle` method _must never fail_, it must never throw an exception. Any error that happens inside it must be sent through the socket connection.

The best way to ensure this on every test was to have the variable `resultPromise` where every test must put the promise returned by `handle`. Then, in an `afterEach`  block it is checked that the promise is always fulfilled and never rejected.
This way, we avoid repeating that same expectation on every test and reduce the chances of someone creating a new test and not checking the resulting promise.

packages/framework-provider-aws/src/library/auth-adapter.ts
alvaroloes: Thanks!

packages/framework-provider-aws/src/library/read-model-adapter.ts
alvaroloes: I think you are right. I wrote this because I was having a lot of failures in integration tests. I don't remember if this was one of the reasons or not 😓

.github/workflows/integration-tests.yml
alvaroloes: No idea

packages/framework-integration-tests/integration/cli/cli.project.integration.ts
alvaroloes: WHAT!? 💥

packages/framework-provider-aws/src/library/read-model-adapter.ts
javiertoledo: Maybe we could configure it differently for integration and for production somehow?

PR #375, opened 2020-08-17
Docs/boost 564 and boost 795 graphql and apollo documentation

alvaroloes:
## Description
Fixes and finishes a lot of documentation about the GraphQL API, Apollo clients and the "GraphQL over WebSocket protocol"
It is in sync with the latest changes in PR #373 

## Checks
- [x] Project Builds


docs/README.md
```diff
@@ -1696,28 +1698,23 @@ Knowing this, you can infer the relationship between those operations and your B
...
 
 #### How to send GraphQL request
-GraphQL uses two existing protocols: 
-
-- _HTTP_ for `mutation` and `query` operations
-- _WebSocket_ for `subscription` operations
-
-The reason for the WebSocket protocol is that, in order for subscriptions to work, there must be a way for the server to send data
-to clients when it is changed. HTTP doesn't allow that, as it is the client the one which always initiates the request.
+GraphQL uses two existing protocols _HTTP_ and _WebSocket_. The reason for the WebSocket protocol is that, in order for subscriptions to work, there must be a way for the server to send data
```
rdiaz82, line 1701: The reason for the WebSocket protocol is to allow subscriptions to work. There must be a way for the server to send data
to clients when there is a change in a subscription. HTTP doesn't allow that, as it is always the client the one who initiates the request.

docs/README.md
```diff
@@ -1876,36 +1868,38 @@ Where:
...
 is modified, a new incoming message will appear in the socket with the updated version of the read model. This message
 will have exactly the same format as if you were done a query with the same parameters.
 
-Following the previous example, we now send a command (using a "mutation" operation) that adds
+Following with the previous example, we now send a command (using a "mutation" operation) that adds
 a new item with sku "ABC_02" to the `CartReadModel`. After it has been added, we receive the updated version of the read model through the
 socket.
 
-> Send the Command
+1. Send the following command:
```
rdiaz82, line 1902: What do you think about to specify here that we are creating a totally different call using the rest API to be more clear about what we are doing?

docs/README.md
```diff
@@ -1920,60 +1914,277 @@ mutation {
...
+#### Using Apollo Client
+One of the best clients to connect to a GraphQL API is the [Apollo](https://www.apollographql.com/) client.
+There will probably be a version for your client technology of choice. These are the main ones:
+- [For Javascript/Typescript](https://www.apollographql.com/docs/react/) ([Github](https://github.com/apollographql/apollo-client))
+- [For iOS](https://www.apollographql.com/docs/ios/) ([Github)](https://github.com/apollographql/apollo-ios))
+- [For Java/Kotlin/Android](https://www.apollographql.com/docs/android/) ([Github](https://github.com/apollographql/apollo-android))
+
+We recommend referring to the documentation of those clients to know how to 
+use them. Here is an example of how to fully instantiate the Javascript client so that it works for queries, mutations and subscriptions:
+```typescript
```
rdiaz82, line 1953: If all this code is available in the apollo client website I would remove it from here, what do you think?

docs/README.md
```diff
@@ -1920,60 +1914,277 @@ mutation {
...
+ { "type": "connection_init", "payload": { "Authorization": "<your token>" } }
 ```
 
-Once the client is setup, a subscription to read model GraphQL request through a GraphQL client could be:
+You normally won't be sending tokens in such a low-level way. GraphQL clients has easier ways to send these tokens. See ["Sending tokens with Apollo client"](#sending-tokens-with-apollo-clients) 
+
+##### Sending tokens with Apollo clients
+
+We recommend going to the specific documentation of the specific Apollo client you are using to know how to send tokens. However, the basics remains the same. Here is an example of how you would configure the Javascript/Typescript Apollo client to send the authorization token. The example is exactly the same as the one shown in section ["Using Apollo clients"](#using-apollo-client) but with the changes needed to send the token. Notice that the only things that change are the `HttpLink` and the `WebSocketLink`:
 ```typescript
```
rdiaz82, line 2063: The same with this code. If it is available on apollo website I would remove, unless we are here explaining something very tricky and uncovered in the official documentation.

docs/README.md
```diff
@@ -1920,60 +1914,277 @@ mutation {
...
+
+##### Refreshing tokens with Apollo clients
+
+Authorization tokens expire after a certain amount of time. When a token is expired, you will get an error and you will need to call the [refresh the token](#refresh-token) endpoint to get a new token. After you have done so, you need to use the new token in your GraphQL operations.
+
+There are several ways to do this. Here we show the simplest one for learning purposes. 
+
+First, we modify the example shown in the section ["Sending tokens with apollo clients](#sending-tokens-with-apollo-clients) so that the token is stored in a global variable and the Apollo links get the token from it. That variable will be updated when the user signs-in and the token is refreshed:
+
+```typescript
```
rdiaz82, line 2118: The same with this code, I would use 🪓 if it is included in apollo website.

rdiaz82 (approved):
Some minor totally ignorable comments but this is a LGTM

rdiaz82 (approved):
some totally ignorable comments but in general LGTM

docs/README.md
alvaroloes: @rdiaz82 Did you mean to change the text to what you wrote? If so, which part did you changed?

docs/README.md
alvaroloes: Make sense! Done here: b7df753be5bb1e66410e2b58ddd1638ccfb27c74

What do you think?

docs/README.md
alvaroloes: It is not exactly the same code, and it is scattered over some sections. If you follow along, you'll see that we will be improving this code to add tokens and the ability to refresh tokens, things that are not int he Apollo website.

Do you think it is not a good idea to put this code in the documentation?

docs/README.md
alvaroloes: Check https://github.com/boostercloud/booster/pull/375#discussion_r473982082

docs/README.md
alvaroloes: This part, especially, is not shown anywhere, and it is hard to figure it out, that's why I wanted to write it here. But! Maybe there are better ways to show it, with less code... not sure. What do you think?

Ideally, this would be shown in an example application, but we still don't have any that uses authentication with tokens as shown here (this is something new)

itrion: Is Appollo refreshing the token automatically, or is that something the devs should take care of?

If it is not Apollo, the important point is using a global variable, and in that case, I honestly don't like that approach.

I have several concerns about these explanations:
- They are too long.
- We are going deeper into something we should not (front end patterns)
- The global variable approach we are suggesting would make no sense in a reactive client application.
- We are mixing the experience required by the reader trying to teach them how to achieve this, assuming it is difficult for them to figure it out. But at the same time, the snippets have a lot of implicit knowledge, such as the `split` function from Apollo, the `isSubscriptionOperation` from us, the `links`, the `caches`...

I'd start by defining the target audience for this section and their motivations to read this, once clear, talk to them in their idiom. If we target expert readers, we can get rid of the low-level details explanation because they already know. If, on the contrary, we aim less-experienced readers, we should use a language and examples they can understand because they need that explanation.

IMO, this section's readers are here because they want to know how to integrate their clients with a Booster backend. It seems they are front end developers, or maybe backend developers making an integration. In both cases, they are familiar with calling APIs and refreshing tokens.

Maybe, the new thing here is the WebSocket stuff. But teaching that is no the responsibility of the Booster documentation. It is our responsibility to guide the readers, so we make their life's easier, and that includes making great use of their time and curating the links we use.

I totally, get your point in explaining this, I'm not saying we should not do it, we just must sure that we do it in the best possible way.

docs/README.md
```diff
@@ -1920,60 +1914,277 @@ mutation {
...
+
+For these reasons, in order to have an effective non-trivial communication through sockets, a sub-protocol is needed. It would be in charge of making both parts understand each other, share authentication tokens, matching response to the corresponding requests, etc.
+
+The Booster WebSocket communication uses the "GraphQL over WebSocket" protocol as subprotocol. It is in charge of all the low level stuff needed to properly send subscription operations to read models and receive the corresponding data.
+
+You don't need to know anything about this to develop usin Booster, neither in the backend side nor in the frontend side (as all the Apollo GraphQL clients uses this protocol), but it is good to know it is there to guarantee a proper communication. In case you are really curious, you can read about the protocol [here](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md)
```
itrion, line 2184: This `The GraphQL over WebSocket protocol` section is all that I meant in my previous comment. It is providing context for newcomers to web sockets, and experienced readers can omit the first two paragraphs. Then, an explanation about why that is relevant for understanding Booster, and finally, then readers, regardless of their expertise, can follow our link for more details about it.

I really loved ❤️  these four paragraphs

itrion (changes requested):

docs/README.md
```diff
@@ -1454,6 +1454,8 @@ Once a user has an access token, it can be included in any request made to your
 _Bearer_ token. It will be used to get the user information and
 authorize them to access protected resources.
 
+To know how to include the access token in your requests, check the section [Authorizing operations](#authorizing-operations)
```
javiertoledo, line 1457:
```suggestion
To learn how to include the access token in your requests, check the section [Authorizing operations](#authorizing-operations)
```

docs/README.md
```diff
@@ -1759,7 +1751,7 @@ mutation {
 }
 ```
 
-> Equivalent bare HTTP request:
+In case we are not using any GraphQL client, this would be the equivalent bare HTTP request:
```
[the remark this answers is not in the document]
javiertoledo: I think we should probably create an issue to do this in the whole documentation just to make sure we don't mix formats.

docs/README.md
```diff
@@ -1848,14 +1837,17 @@ METHOD: "POST"
...
 ```
 
+>**Note**: Remember that, in case you want to query a read model that is restricted to a specific set of roles, you must send the **access token** retrieved upon sign-in. Check ["Authorizing operations"](#authorizing-operations) to know how to do this.
+
 #### Subscribing to read models
 
-To subscribe to a specific read model, we need to use a "subscription" operation, and it must be _sent through the **websocketURL**_.
+To subscribe to a specific read model, we need to use a subscription operation, and it must be _sent through the **websocketURL**_ using the [_GraphQL over WebSocket_ protocol](#the-graphql-over-websocket-protocol).
 
+Doing this process manually is a bit cumbersome. _You will probably never need to do this_, as GraphQL clients like [Apollo](#using-apollo-client) abstract this process away. However, we will explain how to do it for learning purposes.
```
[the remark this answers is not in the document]
javiertoledo: Definitely an interesting article that we could publish, but it could make sense to keep a copy in the "going deeper" guide. I think it's worth having an easy to find explanation at hand in case someone wants to implement the protocol themselves...

javiertoledo (approved):
LGTM

PR #383, opened 2020-08-25
fix(cli) error messages on empty artifact names

charlietfe:
## Description
This PR fix how we were returning error messages, so instead of throwing the whole Error object we simply return the string error message.


NickSeagull (approved):
LGTM

PR #388, opened 2020-08-28
Enable CORS in the auth endpoints

### LINKED ISSUE #387: Enable CORS in the auth endpoints
There is a CORS issue in the AWS provider because the `auth` resource within the `httpUrl` API Gateway does not support [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), and it is making the Booster generated applications unusable from web clients, see #311 

image

**Acceptance criteria**
The generated `auth` resource must:
- Support the [preflight request](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) by adding an `OPTIONS` method returning these headers, e.g.

```http
OPTIONS /auth/sign-up

Access-Control-Request-Method: OPTIONS,POST
Access-Control-Request-Headers: *
Origin: *
```
- Include one `POST` methods under each `/auth/sign-up`, `/auth/sign-in` and `/auth/sign-out` resources with the `Access-Control-Allow-Origin: * ` header in the response.

Regarding the headers' values, we can default `Access-Control-Request-Method` to `OPTIONS,POST` because we are not generating more methods than those. And we can default `Access-Control-Request-Headers` to `Content-Type,X-Amz-Date,Authorization,X-Api-Key,X-Amz-Security-Token` because that's what AWS suggests.

**We still might need to answer:**
- What should be the value for the `Origin` header in the preflight request?
- What should be the value for the `Access-Control-Allow-Origin` header in the POST requests?

As a first iteration, and to avoid making the issue's scope too big, we can default them to `*`, but ideally, the user must be able to configure the allowed domains for CORS.

What do you think @javiertoledo @NickSeagull @alvaroloes ?

itrion:
## Description
This PR resolves #387 

## Changes
The Auth Stack generates the auth API with preflight support (accept OPTIONS requests) for the `auth/sign-up`, `auth/sign-in` ,and `auth/sign-out` endpoints. Also, the corresponding `POST` methods return the `Access-Control-Allow-Origin: *' header.

The integration test verifies the OPTIONS requests for almost all the used HTTP verbs. Regarding the Posts request, there are three tests pending because it is difficult to test the response for 500 errors.

Note that this solves the issue only for the AWS provider. However, I wrote the test suite as a black box one, so we can move it up and run it against all the providers' implementations. At this moment the best would be open tickets for reproducing the solution for each provider or wait until someone needs it.

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-integration-tests/integration/providers/aws/cors.integration.ts
+++ b/packages/framework-integration-tests/integration/providers/aws/cors.integration.ts
@@ -3,28 +3,30 @@ import { internet } from 'faker'
 import fetch from 'cross-fetch'
 import { expect } from '@boostercloud/framework-provider-aws/test/expect'
 
-describe('The Authentication API', () => {
+describe('Given the Authentication API', () => {
   let clientId: string
-  const methodsToCheck = ['GET', 'POST', 'PUT', 'DELETE', 'OPTIONS', 'PATCH']
-  const preflightOptions = generatePreflightOptionsList(methodsToCheck)
-
+  const username = internet.email()
+  const password = createPassword()
+  const role = 'SuperUserNoConfirmation'
   before(async () => {
     clientId = await authClientID()
   })
 
-  context('/auth/sign-up', () => {
+  context('When /auth/sign-up', () => {
     let signUpUrl: string
-    let validSignUpBody: string
-    const invalidSignUpBody = JSON.stringify({})
+    let validAuthBody: string
+    const invalidAuthBody = JSON.stringify({})
+    const methodsToCheck = ['GET', 'POST', 'PUT', 'DELETE', 'OPTIONS', 'PATCH']
+    const preflightOptions = generatePreflightOptionsList(methodsToCheck)
 
     before(async () => {
       signUpUrl = await signUpURL()
-      validSignUpBody = JSON.stringify({
-        clientId: clientId,
-        username: internet.email(),
-        password: createPassword(),
+      validAuthBody = JSON.stringify({
+        clientId,
+        username,
+        password,
         userAttributes: {
-          role: 'UserWithEmail',
+          role,
         },
       })
     })
@@ -44,11 +46,10 @@ describe('The Authentication API', () => {
           headers: {
             'Content-Type': 'Application/json',
           },
-          body: validSignUpBody,
+          body: validAuthBody,
         })
 
-        expect(response.status).to.be.eq(200, `Response body was: ${JSON.stringify(await response.json())}`)
-        expect(response.headers.get('Access-Control-Allow-Origin')).to.be.eq('*')
+        await verifyResponseAndAllowedOriginHeader(response, 200, '*', await response.json())
       })
 
       it('should return the Access-Control-Allow-Origin header for 400 responses', async () => {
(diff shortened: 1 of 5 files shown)


packages/framework-integration-tests/integration/providers/aws/auth.integration.ts
```diff
@@ -33,7 +33,7 @@ describe('With the auth API', () => {
     mockCartId = random.uuid()
   })
 
-  context('an internet rando', () => {
+  context('an internet random', () => {
```
NickSeagull, line 36: Actually the correct term is rando (Source: https://www.merriam-webster.com/dictionary/rando) 😅

NickSeagull (approved):
I'm not sure if this will work on browsers, but yet thanks a lot LGTM!

packages/framework-integration-tests/integration/providers/aws/auth.integration.ts
javiertoledo: 👍 
```suggestion
  context('an internet rando', () => {
```

packages/framework-integration-tests/integration/providers/aws/cors.integration.ts
```diff
@@ -0,0 +1,184 @@
...
+        .to.include('OPTIONS')
+        .and.to.include('POST')
+    }
+
+    async function verifyResponseAndAllowedOriginHeader(response: Response, expectedHttpStatus: number, expectedAllowedOrigin: string, jsonBody: any) {
+      expect(response.status).to.be.eq(expectedHttpStatus, `Response body was: ${JSON.stringify(jsonBody)}`)
+      expect(response.headers.get('Access-Control-Allow-Origin')).to.be.eq(expectedAllowedOrigin)
+    }
+  })
+})
```
javiertoledo, line 184: Nice tests!

javiertoledo (changes requested):
Great work here. I think we should add some integration tests to check the actual request before merging it though.

itrion:
@javiertoledo, @NickSeagull 

Would you please, be more clear about the integration tests you expect?

This PR adds tests that are making real requests to the auth endpoints and verifying the response headers (acting as a browser)

javiertoledo:
LOL for some reason I read that integration tests file as unit tests, so you're right, the integration tests are done 😅 

Then... would it make sense to add some unit tests to have the changes in `packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/auth-stack.ts` under control?

javiertoledo:
One thing to note is that I separated integration tests from the deploy tests, and they're now in a different directory, you might want to rebase this branch and move the integration test file to [the `integration/providers/aws/functionality` directory](https://github.com/boostercloud/booster/tree/master/packages/framework-integration-tests/integration/providers/aws/functionality). I had to do this in the master branch because I was unable to run GH actions from a branch...

itrion:
I thought about it and decided to avoid those tests because they would be too coupled to the AWS CDK API. 

Those unit tests would be injecting mocks and verifying the interactions with the mock, but mocking the CDK has disadvantages:

- It will require us to update the test suite when they change the API, and we wouldn't be sure it actually works.
- There would be changes in the internals of the CDK that won't be caught by our test suite because it is mocked.

So, IMO the value here is ensuring the what without looking too deep at the how

javiertoledo:
The value of having these tests in place is that if we suddenly change that code, the tests will tell us what changed. If we only have black-box tests, we'll know that there's an error, but when it fails, we'll need to troubleshoot the problem

javiertoledo (approved):

javiertoledo:
Please make sure you rebase this branch to `master` and run the integration tests adding a comment that reads `integration: all`

alvaroloes (approved):

javiertoledo:
integration: all

javiertoledo:
I think there's nothing remaining in this PR, so I'll started the integration tests.

javiertoledo:
It looks like the integration tests failed, you might need to rebase this branch again T_T

alvaroloes:
@javiertoledo Where do you see that integration test failed? I don't see the check here 😳

javiertoledo:
It failed during compilation, but the checks were never updated here, I have no idea why: https://github.com/boostercloud/booster/runs/1123683842?check_suite_focus=true

javiertoledo:
integration: all

NickSeagull:
Everything passed right? We can merge this I guess...

javiertoledo:
Integration tests never passed in Github actions. I'll run them locally and hit merge if they pass

javiertoledo:
I found some issues, the project doesn't seem to be compiling anymore. I'll need to recreate this PR...

PR #398, opened 2020-09-02
[BOOST-651] - Destroy/Delete entity

charlietfe:
## Description

This PR adds a new feature to allow the deletion of entities. To delete an entity we've provided a static method called `Booster.destroyEntity(entityName, entityID)` which deletes all entity events and snapshots.

Here is a basic deletion example in a DeleteProduct Command:

```
@Command({
  authorize: [Admin],
})
export class DeleteProduct {
  public constructor(readonly productId: UUID) {}

  public async handle(register: Register): Promise<void> {
    await Booster.destroyEntity(Product, this.productId)
    register.events(new ProductDeleted(this.productId))
  }
}
```

In the example we explicitly register a `ProductDeleted` event simply to update the Product Read Model accordingly.  

## Changes

We've created all the code needed in the provider to allow deletions. To perform the delete operation we are performing 2 queries to get all the events and snapshots associated to an entity. Once we get the all the events to be deleted we'll perform a `BatchWriteItem` operation.  

## Checks
- [X] Project Builds

## Additional information

This operation is only available for the AWS provider.


packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -9,12 +9,10 @@ import { Converter } from 'aws-sdk/clients/dynamodb'
 const originOfTime = new Date(0).toISOString()
 
 export function rawEventsToEnvelopes(rawEvents: DynamoDBStreamEvent): Array<EventEnvelope> {
-  return rawEvents.Records.map(
+  return rawEvents.Records.filter((item: DynamoDBRecord) => item.dynamodb?.NewImage).map(
```
charlietfe, line 12: There is a reason for this change and that's because now we are removing items from the dynamo db events table, we'll receive that as a change to be published in the dynamodb stream. The deleted items doesn't contain the `NewImage` field so instead of throwing an exception and avoid processing the stream, here simply we'll get only the items with `NewImage`.

Also we don't need to process deleted events in the stream since we've a way to register explicitly a new deleted event on the command side and it will be processed as always.

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -110,3 +110,55 @@ export async function storeEvents(
...
+                eventEnvelope.entityID,
+                eventEnvelope.kind
+              ),
+              [eventsStoreAttributes.sortKey]: eventEnvelope.createdAt,
+            },
+          },
+        })),
+      },
+    }
+    await dynamoDB.batchWrite(params).promise()
```
charlietfe, line 159: For the future we should create chunks of 25 items to be written. The [batchWriteItem](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html) operation has that limit.

packages/framework-core/src/booster.ts
```diff
@@ -102,6 +103,13 @@ export class Booster {
   ): Promise<TEntity | undefined> {
     return fetchEntitySnapshot(this.config, this.logger, entityClass, entityID)
   }
+
+  public static destroyEntity<TEntity extends EntityInterface>(
```
NickSeagull, line 107: I'd add a doc comment documenting this function for IDE's autocomplete sake

packages/framework-core/src/services/event-store.ts
```diff
@@ -50,6 +50,11 @@ export class EventStore {
     }
   }
 
+  public async destroyEntity(entityName: string, entityID: UUID): Promise<void> {
+    this.logger.debug(`[EventStore#storeSnapshot] Destroying for entity ${entityName} with ID ${entityID}`)
```
NickSeagull, line 54: Small typo

```suggestion
    this.logger.debug(`[EventStore#storeSnapshot] Destroying entity ${entityName} with ID ${entityID}`)
```

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -9,12 +9,10 @@ import { Converter } from 'aws-sdk/clients/dynamodb'
...
 
 export function rawEventsToEnvelopes(rawEvents: DynamoDBStreamEvent): Array<EventEnvelope> {
-  return rawEvents.Records.map(
+  return rawEvents.Records.filter((item: DynamoDBRecord) => item.dynamodb?.NewImage).map(
     (record: DynamoDBRecord): EventEnvelope => {
-      if (!record.dynamodb?.NewImage) {
-        throw new Error('Received a DynamoDB stream event without "NewImage" field. It is required')
-      }
-      return Converter.unmarshall(record.dynamodb?.NewImage) as EventEnvelope
+      // Here we are really sure that we've only new records to emit
```
NickSeagull, line 14:
```suggestion
      // Here we are really sure that we've got only new records to emit
```

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -25,16 +23,18 @@ export async function readEntityEventsSince(
   logger: Logger,
   entityTypeName: string,
   entityID: UUID,
-  since?: string
+  since?: string,
+  kind?: EventEnvelope['kind']
 ): Promise<Array<EventEnvelope>> {
   const fromTime = since ? since : originOfTime
+  const eventKind = kind ? kind : 'event'
```
NickSeagull, line 30: These two lines can be:
```suggestion
  const fromTime = since ?? originOfTime
  const eventKind = kind ?? 'event'
```

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -110,3 +110,55 @@ export async function storeEvents(
...
+    'snapshot'
+  )
+
+  if (events) {
+    let eventsToDelete = [...events]
+    if (snapshots) {
+      eventsToDelete = [...eventsToDelete, ...snapshots]
+    } else {
+      logger.debug(`[EventsAdapter#destroyEntity] No snapshots found for entity ${entityTypeName} with id: ${entityID}`)
+    }
```
NickSeagull, line 141: This is a matter of personal preference, but perhaps you like the same version as me 😃, with less indentation

```suggestion
    if (!snapshots) {
      logger.debug(`[EventsAdapter#destroyEntity] No snapshots found for entity ${entityTypeName} with id: ${entityID}`)
    }
    const eventsToDelete = [...events, ...snapshots ?? []]
```

packages/framework-provider-aws/src/library/events-adapter.ts
NickSeagull: Can you add a `TODO` comment about this in this place, please?

packages/framework-provider-aws/test/library/events-adapter.test.ts
```diff
@@ -131,8 +132,75 @@ describe('the events-adapter', () => {
...
+      value: {
+        id: entityId,
+      },
+      typeName: random.word(),
+      entityTypeName: entityName,
+      requestID: random.uuid(),
+      createdAt: date.past().toISOString(),
+    }
+  }
+  return events
```
NickSeagull, line 201: Perhaps this, in order to avoid future off-by-one errors?

```suggestion
  return new Array<EventEnvelope>(numberOfEvents)
    .fill(undefined)
    .map(() => ({
      version: 1,
      entityID: entityId,
      kind: kind,
      value: {
        id: entityId,
      },
      typeName: random.word(),
      entityTypeName: entityName,
      requestID: random.uuid(),
      createdAt: date.past().toISOString(),
    }))
```

NickSeagull (commented):
Lovely! Left some suggestions

packages/framework-core/src/entity-destroyer.ts
```diff
@@ -0,0 +1,12 @@
+import { UUID, BoosterConfig, EntityInterface, Logger, Class } from '@boostercloud/framework-types'
+import { EventStore } from './services/event-store'
+
+export async function destroyEntity<TEntity extends EntityInterface>(
+  config: BoosterConfig,
+  logger: Logger,
+  entityClass: Class<TEntity>,
+  entityID: UUID
+): Promise<void> {
+  const eventStore = new EventStore(config, logger)
```
javiertoledo, line 10: It's weird that you have to create an event store instance when there's a single event store in the whole system. We should eventually review this implementation detail...

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -110,3 +110,56 @@ export async function storeEvents(
...
+  const events = await readEntityEventsSince(dynamoDB, config, logger, entityTypeName, entityID)
+  const snapshots = await readEntityEventsSince(
+    dynamoDB,
+    config,
+    logger,
+    entityTypeName,
+    entityID,
+    originOfTime,
+    'snapshot'
+  )
```
javiertoledo, line 133: I think that you could get both in a single query and avoid modifying the `readEntityEventsSince` function by looking for all the objects in the events table that share the same entityTypeName and entityID:

```typescript
const eventsAndSnapshots = await dynamoDB
     .query({
       TableName: config.resourceNames.eventsStore,
       ConsistentRead: true,
       KeyConditionExpression: `${eventsStoreAttributes.partitionKey} = :partitionKey`,
       ExpressionAttributeValues: {
         ':partitionKey': partitionKeyForEvent(entityTypeName, entityID),
       },
       ScanIndexForward: true, // Ascending order (older timestamps first)
    })
    .promise()
```

Also, I don't think you need to set the `ScanindexForward` parameter, it doesn't really matter in which order you delete them.

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -110,3 +110,56 @@ export async function storeEvents(
...
+                eventEnvelope.entityTypeName,
+                eventEnvelope.entityID,
+                eventEnvelope.kind
+              ),
+              [eventsStoreAttributes.sortKey]: eventEnvelope.createdAt,
+            },
+          },
+        })),
+      },
+    }
```
javiertoledo, line 157: Wow, I had to doublecheck the DynamoDB documentation because I didn't believe that you can't just send a delete operation for an entire partition key. This sucks and could eventually become a problem for entities with enough events that they don't all fit in the lambda memory...

javiertoledo (changes requested):
Looking good! I left some minor comments. I'm missing some integration tests to make that everything keeps working in AWS as we add more changes in the future 😜

packages/framework-core/src/entity-destroyer.ts
charlietfe: @javiertoledo Agreed, I think we should create the event store at a Booster main class level and call it directly. In that way we could remove the `entity-detroyer.ts` and also de `entity-snapshot-fetcher.ts`. Makes sense?

packages/framework-provider-aws/src/library/events-adapter.ts
charlietfe: Regarding `ScanIndexForward` since I'm using an existing method, I need to check the implications of not setting it. Let me check.

packages/framework-provider-aws/src/library/events-adapter.ts
charlietfe: @javiertoledo The `batchWrite` works in chunks of 25 items at a time, so we might be start thinking on paginating the internal query and loop through results. Also every delete request deletes exactly one item a time because we are using the `partitionKey` and `sortKey`, what's the potential problem here?

NickSeagull (approved):

packages/framework-provider-aws/src/library/events-adapter.ts
javiertoledo: Gotcha... If we're anticipating that we can only perform 25 items at once, I think that it would make sense to manage that in this PR to avoid leaving loose ends that we know could become a problem in the future...

The problem I see is: Imagine you're trying to delete an entity with a very long track record. For instance, let's say we implement an async cloud-based version of the PONG game where each user adds a move command every time they hit the up or down arrow keys, and the game state is calculated after reducing all the up and down events from both players. If we want to delete a game that's been played for let's say, a month, it could happen that the events related to a single game don't fit in the available RAM memory of the lambda that's trying to delete the game.

This problem is partially solved if you apply pagination, but then the problem is: what if the time needed to go over all these events and delete them in chunks of 25 goes beyond the time execution limits of the lambda? Then we have a similar situation.

Maybe it's too early to worry about these matters at this moment and we should just merge this as it is and think about these problems where the things actually fail... what do you think?

packages/framework-provider-aws/src/library/events-adapter.ts
charlietfe: @javiertoledo The problem with our partition key is that contains also de `kind` field. So we can get the snapshots and events at the same time. Another option could be a `scan` instead of a `query`. But we should take in into account the cost of this operation, anyway who is deleting things all the time? Thoughts?

charlietfe: What if we don't rely on querying first, just delete the whole thing?

javiertoledo: Gotcha, in that case, I think it's better to keep your current implementation. This use case should be avoided in most situations, you should have a very good reason to delete past events, like complying with GDPR regulations or something like that, so it's okay-ish if it's slow. What we have to make sure of is that it doesn't fail. I think we should at least create an issue to review this in the future to implement pagination or use step functions or something like that to make sure that the process doesn't fail for long event streams.

juanjoman (approved):
LGTM! 🚀

packages/framework-provider-aws/src/library/events-adapter.ts
charlietfe: @javiertoledo Finally I've included the pagination on deletions and the `destroyEntity` method inside `events-adapter.ts` is self contained. There is no need to execute two queries to get snapshots and then events, now we are performing a paginated scan, filtering by `entityName` and `entityID` getting 25 items per page.  Also I've restored the `readEntityEventsSince` to its original version since I'm not using it.

In addition,in the integration tests I've included at least one hundred product updates to verify that all the events and snapshots are being deleted.

PR #404, opened 2020-09-04
Delete read models from projections

juanjoman:
## Description
With this change, a developer will be able to delete their read models by returning the `deleteReadModel` type. Here's an example:
```
@ReadModel
export class UserReadModel {
  public constructor(readonly username: string, /* ...(other interesting fields from users)... */) {}
  
  @Projects(User, 'id')
  public static projectUser(entity: User, current?: UserReadModel): ProjectionResult<UserReadModel> {
    if (isReadModelDeletable()) {
        return deleteReadModel // This is a type from ProjectionResult
    }
    return new UserReadModel(...)
  }

  private isReadModelDeletable(): boolean {
    ...
  }
```

## Changes
1. Insert a `deleteReadModel` in the `read-model-store` flow, which ends up calling `read-model-adapter` and deleting a read model with a specific id
2. Changed the return value of projections from `TReadModel` to `ProjectionResult<TReadModel>`. This way we can set more values in the future if needed and we also don't return null, which is not that meaningful
3. Modified the `read-model` template to adapt it to the new `ProjectionResult` type
4. Added integration and unit testing
5. Updated the docs

## Checks
- [X] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -573,7 +573,7 @@ export class PostReadModel {
   public constructor(public id: UUID, readonly title: string, readonly author: string) {}
 
   @Projects(Post, 'postId')
-  public static projectPost(entity: Post, currentPostReadModel?: PostReadModel): PostReadModel {
+  public static projectPost(entity: Post, currentPostReadModel?: PostReadModel): ProjectionResult<PostReadModel> {
     return new PostReadModel(entity.id, entity.title, entity.author)
   }
 }
@@ -1349,12 +1349,12 @@ export class ReadModelName {
   ) {}
 
   @Projects(SomeEntity, 'entityField')
-  public static projectionName(entity: SomeEntity, currentEntityReadModel?: ReadModelName): ReadModelName {
+  public static projectionName(entity: SomeEntity, currentEntityReadModel?: ReadModelName): ProjectionResult<ReadModelName> {
     return new ReadModelName(/* initialize here your constructor properties */)
   }
   
   @Projects(SomeEntity, 'othetEntityField')
-  public static projectionName(entity: SomeEntity, currentEntityReadModel?: ReadModelName): ReadModelName {
+  public static projectionName(entity: SomeEntity, currentEntityReadModel?: ReadModelName): ProjectionResult<ReadModelName> {
     return new ReadModelName(/* initialize here your constructor properties */)
   }
   /* as many projections as needed */
@@ -1363,7 +1363,7 @@ export class ReadModelName {
 
 #### Read models naming convention
 
-As it has been previously commented, semantics plays an important role in designing a coherent system and your application should reflect your domain concepts, we recommend to chooose a representative domain name and use the `ReadModel` suffix in your read models name.
+As it has been previously commented, semantics plays an important role in designing a coherent system and your application should reflect your domain concepts, we recommend choosing a representative domain name and use the `ReadModel` suffix in your read models name.
 
 Despite you can place your read models in any directory, we strongly recommend you to put them in `<project-root>/src/read-models`. Having all the read models in one place will help you to understand your application's capabilities at a glance.
 
@@ -1402,13 +1402,32 @@ export class UserReadModel {
   public constructor(readonly username: string, /* ...(other interesting fields from users)... */) {}
   
   @Projects(User, 'id')
-  public static projectUser(entity: User, current?: UserReadModel) { // Here we update the user fields}
+  public static projectUser(entity: User, current?: UserReadModel): ProjectionResult<UserReadModel> { // Here we update the user fields}
 
   @Projects(Post, 'ownerId')
-  public static projectUserPost(entity: Post, current?: UserReadModel) { //Here we can adapt the read model to show specific user information related with the Post entity}
+  public static projectUserPost(entity: Post, current?: UserReadModel): ProjectionResult<UserReadModel> { //Here we can adapt the read model to show specific user information related with the Post entity}
 }
 ```
 In the previous example we are projecting the `User` entity using the user `id` and also we are projecting the `User` entity based on the `ownerId` of the `Post` entity. Notice that both join keys are references to the `User` identifier, but it's not required that the join key is an identifier.
+
+You can also delete read models by returning the `deleteReadModel` type, as shown in the following example:
+```
+@ReadModel
+export class UserReadModel {
+  public constructor(readonly username: string, /* ...(other interesting fields from users)... */) {}
+  
+  @Projects(User, 'id')
+  public static projectUser(entity: User, current?: UserReadModel): ProjectionResult<UserReadModel> {
+    if (isReadModelDeletable()) {
+        return deleteReadModel 
+    }
+    return new UserReadModel(...)
+  }
+
+  private isReadModelDeletable(): boolean {
+    ...
+  }
+```
 
 #### Authorizing read models
 
@@ -1429,7 +1448,7 @@ export class CartReadModel {
     ) {}
 
   @Projects(Cart, "id")
-  public static projectCart(entity:Cart, currentReadModel: CartReadModel): CartReadModel {
+  public static projectCart(entity:Cart, currentReadModel: CartReadModel): ProjectionResult<CartReadModel> {
     return new CartReadModel(entity.id, entity.items)
   }
 }
--- a/docs/README.md
+++ b/docs/README.md
@@ -573,7 +573,7 @@ export class PostReadModel {
   public constructor(public id: UUID, readonly title: string, readonly author: string) {}
 
   @Projects(Post, 'postId')
-  public static projectPost(entity: Post, currentPostReadModel?: PostReadModel): PostReadModel | null {
+  public static projectPost(entity: Post, currentPostReadModel?: PostReadModel): PostReadModel | ReadModelAction {
     return new PostReadModel(entity.id, entity.title, entity.author)
   }
 }
@@ -1349,12 +1349,12 @@ export class ReadModelName {
   ) {}
 
   @Projects(SomeEntity, 'entityField')
-  public static projectionName(entity: SomeEntity, currentEntityReadModel?: ReadModelName): ReadModelName | null {
+  public static projectionName(entity: SomeEntity, currentEntityReadModel?: ReadModelName): ReadModelName | ReadModelAction {
     return new ReadModelName(/* initialize here your constructor properties */)
   }
   
   @Projects(SomeEntity, 'othetEntityField')
-  public static projectionName(entity: SomeEntity, currentEntityReadModel?: ReadModelName): ReadModelName | null {
+  public static projectionName(entity: SomeEntity, currentEntityReadModel?: ReadModelName): ReadModelName | ReadModelAction {
     return new ReadModelName(/* initialize here your constructor properties */)
   }
   /* as many projections as needed */
@@ -1402,24 +1402,24 @@ export class UserReadModel {
   public constructor(readonly username: string, /* ...(other interesting fields from users)... */) {}
   
   @Projects(User, 'id')
-  public static projectUser(entity: User, current?: UserReadModel): UserReadModel | null { // Here we update the user fields}
+  public static projectUser(entity: User, current?: UserReadModel): UserReadModel | ReadModelAction { // Here we update the user fields}
 
   @Projects(Post, 'ownerId')
-  public static projectUserPost(entity: Post, current?: UserReadModel): UserReadModel | null { //Here we can adapt the read model to show specific user information related with the Post entity}
+  public static projectUserPost(entity: Post, current?: UserReadModel): UserReadModel | ReadModelAction { //Here we can adapt the read model to show specific user information related with the Post entity}
 }
 ```
 In the previous example we are projecting the `User` entity using the user `id` and also we are projecting the `User` entity based on the `ownerId` of the `Post` entity. Notice that both join keys are references to the `User` identifier, but it's not required that the join key is an identifier.
 
-You can also delete read models by returning null, as shown in the following example:
+Projections can also return `ReadModelAction`. For now, you can delete read models by returning the `ReadModelAction.DELETE` value, as shown in the following example:
 ```
 @ReadModel
 export class UserReadModel {
   public constructor(readonly username: string, /* ...(other interesting fields from users)... */) {}
   
   @Projects(User, 'id')
-  public static projectUser(entity: User, current?: UserReadModel): UserReadModel | null {
+  public static projectUser(entity: User, current?: UserReadModel): UserReadModel | ReadModelAction {
     if (current?.deleted) {
-        return null
+        return ReadModelAction.DELETE
     }
     return new UserReadModel(...)
   }
@@ -1444,7 +1444,7 @@ export class CartReadModel {
     ) {}
 
   @Projects(Cart, "id")
-  public static projectCart(entity:Cart, currentReadModel: CartReadModel): CartReadModel | null {
+  public static projectCart(entity:Cart, currentReadModel: CartReadModel): CartReadModel | ReadModelAction {
     return new CartReadModel(entity.id, entity.items)
   }
 }
--- a/docs/README.md
+++ b/docs/README.md
@@ -573,7 +573,7 @@ export class PostReadModel {
   public constructor(public id: UUID, readonly title: string, readonly author: string) {}
 
   @Projects(Post, 'postId')
-  public static projectPost(entity: Post, currentPostReadModel?: PostReadModel): ProjectionResult<PostReadModel> {
+  public static projectPost(entity: Post, currentPostReadModel?: PostReadModel): PostReadModel | null {
     return new PostReadModel(entity.id, entity.title, entity.author)
   }
 }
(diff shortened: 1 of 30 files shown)


packages/cli/test/commands/new.test.ts
```diff
@@ -1,5 +1,56 @@
...
-  // TODO: Write proper tests
+  describe('Read model', () => {
+    const readModel = 'ExampleReadModel'
+    const readModelsRoot = './src/read-models/'
+    const readModelPath = `${readModelsRoot}${readModel}.ts`
+    afterEach(() => {
+      if (fs.existsSync(readModelPath)) {
+        fs.rmdir(readModelsRoot, { recursive: true }, (err: ErrnoException | null) => console.log(err))
+      }
+      restore()
```
juanjoman, line 20: To be honest I'm not very convinced about this test (creating a file -> check input -> delete file). Suggestions are accepted 🙏

docs/README.md
```diff
@@ -1397,14 +1397,33 @@ export class UserReadModel {
...
+  public static projectUser(entity: User, current?: UserReadModel): ProjectionResult<UserReadModel> {
+    if (isReadModelDeletable()) {
+        return deleteReadModel 
+    }
+    return new UserReadModel(...)
+  }
+
+  private isReadModelDeletable(): boolean {
+    ...
+  }
```
NickSeagull, line 1424: I'd try to make a less generic example, like adding a field called `deleted` or something, and then checking `if (current?.deleted)` or something

packages/framework-core/src/services/read-model-store.ts
```diff
@@ -43,6 +43,10 @@ export class ReadModelStore {
           ' to build new state of read model ${readModelName} with ID ${readModelID}'
         )
         const newReadModel = this.reducerForProjection(projectionMetadata)(entitySnapshot, readModel)
+
+        if (newReadModel === deleteReadModel) {
+          return this.provider.readModels.deleteReadModel(this.config, this.logger, readModelName, readModel)
+        }
```
NickSeagull, line 49: I'd add a `logger.debug` here

packages/framework-core/src/services/read-model-store.ts
```diff
@@ -69,7 +73,7 @@ export class ReadModelStore {
     return joinKey
   }
 
-  private reducerForProjection(projectionMetadata: ProjectionMetadata): Function {
+  public reducerForProjection(projectionMetadata: ProjectionMetadata): Function {
```
NickSeagull, line 76: Why is this public now?

packages/framework-core/test/services/read-model-store.test.ts
```diff
@@ -89,6 +97,23 @@ describe('ReadModelStore', () => {
...
+        replace(
+          ReadModelStore.prototype,
+          'reducerForProjection',
+          fake.returns(() => deleteReadModel)
+        )
+        const readModelStore = new ReadModelStore(config, logger)
+
+        await readModelStore.project(anEntitySnapshot)
+        expect(config.provider.readModels.store).not.to.have.been.called
+        expect(config.provider.readModels.deleteReadModel).to.have.been.calledTwice
```
NickSeagull, line 113: Why twice?

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -307,22 +307,12 @@ describe('Cart end-to-end tests', () => {
               `,
             })
           },
-          (result) => result?.data?.ProductReadModel?.deleted && result?.data?.ProductReadModel?.id === productId
+          () => true
```
NickSeagull, line 310: What's this `true`? xD

packages/framework-core/src/decorators/projection.ts
```diff
@@ -35,4 +35,6 @@ function registerProjection(originName: string, projectionMetadata: ProjectionMe
   })
 }
 
-type ProjectionMethod<TEntity, TReadModel> = TypedPropertyDescriptor<(_: TEntity, readModel?: TReadModel) => TReadModel>
+type ProjectionMethod<TEntity, TReadModel> = TypedPropertyDescriptor<
+  (_: TEntity, readModel?: TReadModel) => ProjectionResult<TReadModel>
```
javiertoledo, line 39: Why not returning a `TReadModel | null` or a `Promise<TReadModel>` instead? The event dispatcher should have enough context to figure out which read model to delete, and we avoid introducing an extra interface to the public-facing SDK.

packages/framework-core/src/services/read-model-store.ts
javiertoledo, line 49: I think that here you could do something like this to avoid the extra type:

```suggestion
        if (newReadModel === null) {
          return this.provider.readModels.deleteReadModel(this.config, this.logger, readModelName, readModel)
        }
```

javiertoledo (changes requested):
At a first glance, I'd reconsider the design of this solution to trigger the deletion of the read model when the projections return a null value (or a rejected promise). Is there a reason why you discarded that idea?

packages/framework-core/src/decorators/projection.ts
juanjoman: Hey @javiertoledo there is no real advantage, it was just to make it more readable for the final user:
```
@ReadModel
export class UserReadModel {
  public constructor(readonly username: string, /* ...(other interesting fields from users)... */) {}
  
  @Projects(User, 'id')
  public static projectUser(entity: User, current?: UserReadModel): UserReadModel {
    if (....)) {
        return null // This could sound like "do nothing"
    }
    return new UserReadModel(...)
  }
```

But yeah it's true that the additional type is not needed at all, let me change it 👀

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
juanjoman: @NickSeagull There is a mandatory check for the data returned. I had to return true to get all data from the query and then check that it's null. If I filter by the `id` and the query doesn't find the result, the test automatically fails and returns nothing.

packages/framework-core/src/services/read-model-store.ts
juanjoman: Just for testing purposes @NickSeagull: 
```
replace(
          ReadModelStore.prototype,
          'reducerForProjection',
          fake.returns(() => deleteReadModel)
        )
```

packages/framework-core/test/services/read-model-store.test.ts
juanjoman: hey @NickSeagull `anEntitySnapshot` has two projections associated, I'm deleting both. Maybe is it better to just mock one? 🤔

packages/framework-core/src/decorators/projection.ts
juanjoman: Changed!

NickSeagull: I strongly disagree with returning `null`. The user can mess-up, return null by accident and delete production data. I don't think that we should allow this. Please reconsider this @javiertoledo @juanjoman

packages/framework-core/src/services/read-model-store.ts
javiertoledo: You can use the `rewire` library to load private functions without exposing them. There's an example here: https://github.com/boostercloud/booster/blob/master/packages/cli/test/commands/deploy.test.ts#L11

juanjoman: Hey @javiertoledo I've been trying to do that but I can't find the way of "stubbing" exported classes methods. All examples I found are from functions outside classes. I can get the ReadModelStore and even print it but it's printed as a function, and I think that's a rewire limitation: https://www.npmjs.com/package/rewire#limitations.

```
const readModelStoreRewire = rewire('../../src/services/read-model-store')
        readModelStoreRewire.__set__({
          ReadModelStore: {
            config,
            logger,
            reducerForProjection: () => null,
          },
        })
        const readModelStore = new readModelStoreRewire.ReadModelStore(config, logger)

        await readModelStore.project(anEntitySnapshot)
```

⬆️ The `project` method gets called but ignores my setters

packages/framework-core/test/services/read-model-store.test.ts
NickSeagull: I was just asking :) its okay! :)

alvaroloes:
After reviewing, I agree completely with @NickSeagull: returning `null` is something dangerous as it could happen without notice. It is also non-meaningful and we should document it clearly.

However, I think the original approach was a bit complex too. Here is a proposal that is expressive and not that complex:
1. Define the following enum (in `framework-types`):
```typescript
enum ReadModelAction {
  DELETE,
}
```
2. Make the projections return the read model **or** that enum. This way, it is clear that you want to delete your read model and it is impossible to do it unnoticedly:
```typescript
  @Projects(Product, 'id')
  public static updateWithProduct(product: Product): ProductReadModel | ReadModelAction {
    if (product.deleted) {
      return ReadModelAction.DELETE
    } else {
      return new ProductReadModel(...)
    }
  }
```

No more changes are needed. What do you all think? @juanjoman @javiertoledo @NickSeagull

juanjoman:
> After reviewing, I agree completely with @NickSeagull: returning `null` is something dangerous as it could happen without notice. It is also non-meaningful and we should document it clearly.
> 
> However, I think the original approach was a bit complex too. Here is a proposal that is expressive and not that complex:
> 
> 1. Define the following enum (in `framework-types`):
> 
> ```ts
> enum ReadModelAction {
>   DELETE,
> }
> ```
> 
> 1. Make the projections return the read model **or** that enum. This way, it is clear that you want to delete your read model and it is impossible to do it unnoticedly:
> 
> ```ts
>   @Projects(Product, 'id')
>   public static updateWithProduct(product: Product): ProductReadModel | ReadModelAction {
>     if (product.deleted) {
>       return ReadModelAction.DELETE
>     } else {
>       return new ProductReadModel(...)
>     }
>   }
> ```
> 
> What do you all think? @juanjoman @javiertoledo @NickSeagull

That looks nice! It preserves also the possibility to add more actions in the future

packages/framework-core/src/decorators/projection.ts
```diff
@@ -36,5 +36,5 @@ function registerProjection(originName: string, projectionMetadata: ProjectionMe
 }
 
 type ProjectionMethod<TEntity, TReadModel> = TypedPropertyDescriptor<
-  (_: TEntity, readModel?: TReadModel) => TReadModel | null
+  (_: TEntity, readModel?: TReadModel) => TReadModel | ReadModelAction
```
alvaroloes: One last suggestion: if you write here something like (maybe you need some parentesis): 
```typescript
type ProjectionMethod<TEntity, TReadModel> = TypedPropertyDescriptor<
     (_: TEntity, readModel?: TReadModel) => TReadModel | ReadModelAction  
  |  (_: TEntity, readModel?: TReadModel) => TReadModel
>
```

You would let users avoid writing the `| ReadModelAction` part in the projection signature in those projections that never delete a ReadModel. This way, the users' code would be cleaner 😃 

What do you think?

alvaroloes (commented):
Thanks a lot! I like it 👏 

I left one more sugestion, let me know what you think

javiertoledo:
Guys, I think that to avoid returning `null` you're overcomplicating the design a little bit. Some things to consider:
* I agree that one problem of `null` is that it can be generated accidentally and end up deleting a read model, but one thing to consider is that in an ideal world Read Models are caches, so if that happens, that's a recoverable issue, you should be able to fix the bug and regenerate it. No production data will be actually deleted in the process.
* I'm afraid you're trying to create some kind of "open design" that allows possible future actions we could come up with, and this seems to be a case of designing "just in case". I'd avoid doing that in any case, even if you want to avoid returning null, it's better to design a solution that solves the current problem, we can always change it in the future.
* Promises are the most transparent way of dealing with this because you either produce a valid read model or the promise is rejected. The question here is, should you actually do anything different than deleting the read model if you can't generate a read model with the current data? If you ignore error scenarios, how is going Booster to recover from that, given that there will always be a point in the event stream where at some past set of events, you can't generate the read model? Would it make sense to delete any failed read models so the next iteration can start from scratch? I don't know, but ask yourselves: what should the framework do if the projection fails?
* If we still want to solve this with a special type, would it make sense to adopt a `Maybe`-like pattern here so we don't need to show the user that complicated type signature?:

```typescript
type ProjectionMethod<TEntity, TReadModel> = TypedPropertyDescriptor<
     (_: TEntity, readModel?: TReadModel) => Maybe<TReadModel>
>
```

Instead of that `Maybe`, we can use a custom type that offers different options, but as all the alternatives are going to be standard in Booster anyway, I'd vote for hiding them.

packages/framework-core/src/decorators/projection.ts
juanjoman: that sounds great!! I was thinking about that yesterday but couldn't find a super quick solution for the templates. I think I will generate the templates (a.k.a `new read-model (...)`) without the `ReadModelAction` and let the user add it. Thank you!

juanjoman:
> @alvaroloes @juanjoman @NickSeagull, I think that to avoid returning `null` you're overcomplicating the design a little bit. Some things to consider:
> 
> * I agree that one problem of `null` is that it can be generated accidentally and end up deleting a read model, but one thing to consider is that in an ideal world Read Models are caches, so if that happens, that's a recoverable issue, you should be able to fix the bug and regenerate it. No production data will be actually deleted in the process.
> * I'm afraid you're trying to create some kind of "open design" that allows possible future actions we could come up with, and this seems to be a case of designing "just in case". I'd avoid doing that in any case, even if you want to avoid returning null, it's better to design a solution that solves the current problem, we can always change it in the future.
> * Promises are the most transparent way of dealing with this because you either produce a valid read model or the promise is rejected. The question here is, should you actually do anything different than deleting the read model if you can't generate a read model with the current data? If you ignore error scenarios, how is going Booster to recover from that, given that there will always be a point in the event stream where at some past set of events, you can't generate the read model? Would it make sense to delete any failed read models so the next iteration can start from scratch? I don't know, but ask yourselves: what should the framework do if the projection fails?
> * If we still want to solve this with a special type, would it make sense to adopt a `Maybe`-like pattern here so we don't need to show the user that complicated type signature?:
> 
> ```ts
> type ProjectionMethod<TEntity, TReadModel> = TypedPropertyDescriptor<
>      (_: TEntity, readModel?: TReadModel) => Maybe<TReadModel>
> >
> ```
> 
> Instead of that `Maybe`, we can use a custom type that offers different options, but as all the alternatives are going to be standard in Booster anyway, I'd vote for hiding them.

Hey @javiertoledo do you mean returning to something like this? (but without the specific delete action but some neutral one like `Nothing`) https://github.com/boostercloud/booster/pull/404/commits/c950389926f33c22061edf4386c7822a7bca3458.
 
cc @NickSeagull @alvaroloes

alvaroloes:
@javiertoledo I agree that with the enum it seems we are leaving the door open to more actions, but that wasn't the intention. It was the cleanest way I found.

I kind of like using Maybe, but there are two problems:
- The intention is not clear: "If I return `Maybe.nothing` the read model would be deleted or it won't be updated?"
- The return is more complicated, as we always need to wrap the responses in a Mayby.

Having said this, we can improve the signature by creating a type for the current approach:
```typescript
  // --> Current approach
  @Projects(Product, 'id')
  public static updateWithProduct(product: Product): ProductReadModel | ReadModelAction {
    if (product.deleted) {
      return ReadModelAction.DELETE
    } else {
      return new ProductReadModel(...)
    }
  }

  // --> If we create the following type:
  type ProjectionResult<TReadModel> = TReadModel | ReadModelAction

  // --> Then the signature would be 
  @Projects(Product, 'id')
  public static updateWithProduct(product: Product): ProjectionResult<ProductReadModel> {
    if (product.deleted) {
      return ReadModelAction.DELETE
    } else {
      return new ProductReadModel(...)
    }
  }
```

What do you think? @juanjoman @javiertoledo

javiertoledo:
I could live with that. It's a good middle ground solution. Now that you mention it, it could make sense to have three options:
* Return a read model to update
* Return a "delete" to.. you know.. delete.
* Return a third option to leave the object untouched. That way the user can respond to exceptions or other circumstances without changing the stored data, triggering data pushes to the clients, or invalidating any caches...

alvaroloes:
Good idea, we can add that action to the `ReadModelAction` enum. Something like:
```typescript
enum ReadModelAction {
  Delete,
  Nothing
}
```

juanjoman (commented):
There is a local-provider test failing 😮 . I'll check later

alvaroloes:
Thanks a lot @juanjoman !!!!

One thing, just in case: Remember to update the documentation (both text and examples) with this new functionality

Once done, feel free to merge and, please, release a new version ("feat") with some release notes saying that you can now delete ReadModels or do nothing from projections ;-)

docs/README.md
```diff
@@ -1402,14 +1402,35 @@ export class UserReadModel {
...
-  public static projectUserPost(entity: Post, current?: UserReadModel) { //Here we can adapt the read model to show specific user information related with the Post entity}
+  public static projectUserPost(entity: Post, current?: UserReadModel): ProjectionResult<UserReadModel> { //Here we can adapt the read model to show specific user information related with the Post entity}
 }
 ```
 In the previous example we are projecting the `User` entity using the user `id` and also we are projecting the `User` entity based on the `ownerId` of the `Post` entity. Notice that both join keys are references to the `User` identifier, but it's not required that the join key is an identifier.
 
+As you may have notice from the `ProjectionResult` type, projections can also return `ReadModelAction`, which includes:
+
+1. Deletion of read models by returning the `ReadModelAction.Delete` value
+2. You can also use `ReadModelAction.Nothing` to keep the read model as it is too!
```
alvaroloes, line 1416: Trying to avoid "also" + "too" in the same sentence and refrasing a bit:
```suggestion
2. You can also return `ReadModelAction.Nothing` to keep the read model untouched
```

alvaroloes (approved):
Approved! I left one suggestion and this one:
- Could you please add an integration test to verify that when you send a, let's say, `DeleteProduct` command, the read model is, effectively, deleted?

PR #407, opened 2020-09-08
[Boost-651] - Entity Deletion

charlietfe:
## Description

This PR adds a new feature to allow the deletion of entities. To delete an entity we've provided a static method called `Booster.destroyEntity(entityName, entityID)` which deletes all entity events and snapshots.

Here is a basic deletion example in a DeleteProduct Command:

```
@Command({
  authorize: [Admin],
})
export class DeleteProduct {
  public constructor(readonly productId: UUID) {}

  public async handle(register: Register): Promise<void> {
    await Booster.destroyEntity(Product, this.productId)
    register.events(new ProductDeleted(this.productId))
  }
}
```

In the example we explicitly register a `ProductDeleted` event simply to update the Product Read Model accordingly.  

## Changes

We've created all the code needed in the provider to allow deletions. To perform the delete operation we are performing a scan operation to get all the events and snapshots associated to an entity. Once we get the all the events to be deleted we'll perform a `BatchWriteItem` operation.

The scan query is being paginated by 25 items at time due to a limitation of `BatchWriteItem` which can only operate under that limit. To paginate items we are using the `LastEvaluatedKey` and `ExclusiveStartKey` as are being described [here](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html)

In addition and to check that all events and snapshots are deleted, even if we have a long history of them for an entity, we've created a test for that.

## Checks
- [X] Project Builds

## Additional information

This operation is only available for the AWS provider.


packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -110,3 +108,57 @@ export async function storeEvents(
   await dynamoDB.batchWrite(params).promise()
   logger.debug('[EventsAdapter#storeEvents] EventEnvelope stored')
 }
+
+export async function destroyEntity(
```
charlietfe, line 112: I've included the pagination on deletions and the destroyEntity method inside events-adapter.ts is self contained. There is no need to execute two queries to get snapshots and then events, now we are performing a paginated scan, filtering by entityName and entityID getting 25 items per page. Also I've restored the readEntityEventsSince to its original version since I'm not using it.

In addition,in the integration tests I've included at least one hundred product updates to verify that all the events and snapshots are being deleted.

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -3,18 +3,16 @@ import { BoosterConfig, EventEnvelope, Logger, UUID } from '@boostercloud/framew
...
 import { partitionKeyForEvent } from './partition-keys'
-import { Converter } from 'aws-sdk/clients/dynamodb'
+import { Converter, DocumentClient } from 'aws-sdk/clients/dynamodb'
 
 // eslint-disable-next-line @typescript-eslint/no-magic-numbers
 const originOfTime = new Date(0).toISOString()
 
 export function rawEventsToEnvelopes(rawEvents: DynamoDBStreamEvent): Array<EventEnvelope> {
-  return rawEvents.Records.map(
+  return rawEvents.Records.filter((item: DynamoDBRecord) => item.dynamodb?.NewImage).map(
```
charlietfe, line 12: There is a reason for this change and that's because now we are removing items from the dynamo db events table, we'll receive that as a change to be published in the dynamodb stream. The deleted items doesn't contain the `NewImage` field so instead of throwing an exception and avoid processing the stream, here simply we'll get only the items with `NewImage`.

Also we don't need to process deleted events in the stream since we've a way to register explicitly a new deleted event on the command side and it will be processed as always.

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -325,6 +325,175 @@ describe('Cart end-to-end tests', () => {
         expect(productData).to.be.deep.equal(expectedResult)
       })
     })
+
+    context('Deletions', () => {
+      let mockSku: string
+      let mockDisplayName: string
+      let mockDescription: string
+      let mockPriceInCents: number
+      let mockCurrency: string
```
NickSeagull, line 334: Why not:
```suggestion
      let mockProduct : {
        sku: string,
        displayName: string,
        description: string,
        priceInCents: number,
        currency: string
      }
``` 

So then you can do... (see next comments)

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -325,6 +325,175 @@ describe('Cart end-to-end tests', () => {
...
+
+      let productId: string
+
+      beforeEach(async () => {
+        mockSku = random.uuid()
+        mockDisplayName = commerce.productName()
+        mockDescription = lorem.paragraph()
+        mockPriceInCents = random.number({ min: 1 })
+        mockCurrency = finance.currencyCode()
```
NickSeagull, line 343:
```suggestion
        mock = {
          sku: random.uuid()
          displayName: commerce.productName()
          description: lorem.paragraph()
          priceInCents: random.number({ min: 1 })
          currency: finance.currencyCode()
        }
```

...and then...

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -325,6 +325,175 @@ describe('Cart end-to-end tests', () => {
...
+
+        await client.mutate({
+          variables: {
+            sku: mockSku,
+            displayName: mockDisplayName,
+            description: mockDescription,
+            priceInCents: mockPriceInCents,
+            currency: mockCurrency,
+          },
```
NickSeagull, line 352:
```suggestion
          variables: { ...mock },
```

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -325,6 +325,175 @@ describe('Cart end-to-end tests', () => {
...
+          (result) => result?.data?.ProductReadModels?.some((product: any) => product.sku === mockSku)
+        )
+
+        const product = products.data.ProductReadModels.find((product: any) => product.sku === mockSku)
+        productId = product.id
+      })
+
+      it('should delete the entity as expected', async () => {
+        const MAX_UPDATES = 100
+        const adminEmail: string = internet.email()
```
NickSeagull, line 402: I believe that the type annotation is not needed:
```suggestion
        const adminEmail = internet.email()
```

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -325,6 +325,175 @@ describe('Cart end-to-end tests', () => {
...
+            })
+          },
+          (result) =>
+            result?.data?.ProductReadModels?.some((product: any) => product.sku === mockSku + `-update-${MAX_UPDATES}`)
+        )
+
+        // Delete the product given an id, it will delete all the events and snapshots
+        await adminClient.mutate({
+          variables: {
+            productId: productId,
```
NickSeagull, line 448: You can write this as
```suggestion
            productId,
```

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -325,6 +325,175 @@ describe('Cart end-to-end tests', () => {
...
+            }
+          `,
+        })
+
+        // Retrieve deleted entity
+        const queryResult = await waitForIt(
+          () => {
+            return client.query({
+              variables: {
+                productId: productId,
```
NickSeagull, line 462:
```suggestion
                productId,
```

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -110,3 +108,57 @@ export async function storeEvents(
...
+  config: BoosterConfig,
+  logger: Logger,
+  entityTypeName: string,
+  entityID: UUID
+): Promise<void> {
+  logger.debug('[EventsAdapter#destroyEntity] Destroying entity')
+  // BatchWrite limitation, in the future we should move this value to BoosterConfig. We are assuming that this process will be
+  // fit into lambda memory and time execution limits.
+  const MAX_EVENTS_TO_DELETE = 25
+  let nextKey: DocumentClient.Key | undefined = undefined
```
NickSeagull, line 123: You can do this:
```suggestion
  let nextKey: DocumentClient.Key
```

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -110,3 +108,57 @@ export async function storeEvents(
...
+  let result = null
```
NickSeagull, line 124: No need for `null` I believe:
```suggestion
  let result
```

NickSeagull (changes requested):
Cool! Left some comments

packages/framework-provider-aws/src/library/events-adapter.ts
charlietfe: Not really since is the type of the LastEvaluatedKey in the scan result.

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
charlietfe: I agree, let's change that.

charlietfe: Fixed [here](https://github.com/boostercloud/booster/pull/407/commits/a165f7a37c4b48185a08ff610c08f48e7941459e)

packages/framework-provider-aws/src/library/events-adapter.ts
charlietfe: Fixed [here](https://github.com/boostercloud/booster/pull/407/commits/a165f7a37c4b48185a08ff610c08f48e7941459e)

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
charlietfe: Fixed!

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
charlietfe: Fixed, same as above.

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
charlietfe: Fixed, same as above.

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
charlietfe: Fixed, same as above.

packages/framework-provider-aws/src/library/events-adapter.ts
NickSeagull: Oh I see, but in that case, you don't have to initialize to `undefined` I believe

charlietfe: It was removed @NickSeagull

NickSeagull (approved):

packages/framework-core/src/booster.ts
```diff
@@ -101,6 +104,19 @@ export class Booster {
   ): Promise<TEntity | undefined> {
     return fetchEntitySnapshot(this.config, this.logger, entityClass, entityID)
   }
+
+  /**
+   * Destroy all the entity events and snapshots. That action will cause the entity deletion, since
+   * we'll be not able to reduce its state based on its events and snapshots.
+   * @param entityName Name of the entity class
```
juanjoman, line 111: The argument is called `entityClass`, I don't know if you wanted to call it `entityName` or if this is a typo ⬇️ 
```suggestion
   * @param entityClass Name of the entity class
```

packages/framework-core/src/services/event-store.ts
```diff
@@ -50,6 +50,11 @@ export class EventStore {
     }
   }
 
+  public async destroyEntity(entityName: string, entityID: UUID): Promise<void> {
```
juanjoman, line 53: Is this `async` needed? I don't see an await here `return this.provider.events.destroy(this.config, this.logger, entityName, entityID)`
```suggestion
  public destroyEntity(entityName: string, entityID: UUID): Promise<void> {
```

juanjoman (changes requested):
Overall LGTM, but I have some requests 🙏

packages/framework-core/src/services/event-store.ts
charlietfe: Yeah is not needed!

juanjoman (approved):
LGTM!

packages/framework-core/src/booster.ts
```diff
@@ -101,6 +104,19 @@ export class Booster {
   ): Promise<TEntity | undefined> {
     return fetchEntitySnapshot(this.config, this.logger, entityClass, entityID)
   }
+
+  /**
+   * Destroy all the entity events and snapshots. That action will cause the entity deletion, since
+   * we'll be not able to reduce its state based on its events and snapshots.
+   * @param entityClass Entity class
```
rdiaz82, line 111: The params are not needed for the doc comments typedoc takes care of it

rdiaz82 (approved):
LGTM just a silly comment

javiertoledo:
Should we close this PR? I think we decided to stop this in favor of a more generic event rewrite feature

javiertoledo:
Closing, as we're re-evaluating the approach

PR #412, opened 2020-09-10
I115 - Allow generating a new project with default parameters

### LINKED ISSUE #115: [CLI] Allow a parameter to run the new:project command with default values
It would be great to have a parameter like `-y` in the `project:new` command so when we run 

```shell
boost project:new my-boosted-project -y
```
The CLI doesn't ask the user for inputs and everything is created with default values

**Current behavior**

```shell
> boost new:project my-boosted-project
? What's your project description? (default: "") 
> Intro
? What's the first version? (default: 0.1.0) 
> Intro
? Who's the author? (default: "") 
> Intro
? What's the website? (default: "") 
> Intro
? What license will you be publishing this under? (default: MIT) 
> Intro
? What's the URL of the repository? (default: "") 
> Intro
? What's the package name of your provider infrastructure library?
> Intro
(AWS)
ℹ boost new 🚧
✔ Creating project root
✔ Generating config files
✔ Installing dependencies
ℹ Project generated!
```

**expected behavior**
```shell
> boost new:project my-boosted-project -y
ℹ boost new 🚧
✔ Creating project root
✔ Generating config files
✔ Installing dependencies
ℹ Project generated!
```

I propose `-y` because that's kind of standard in CLIs for `yes to all`

juanjoman:
## Description
Now, when running `boost new:project <my-project> --default`, booster will create the project with all default parameters, without asking the user.

## Changes
- Added a `--default` (or `-d`) flag
- Stopped the "CLI prompt flow" and returned the default parameters from `parseConfig` function when that flag is present
- Wrote tests and refactor common cases from both ones

## Checks
- [X] Project Builds

Fixes #115

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -334,8 +334,13 @@ configured [here](#set-up-an-aws-account).
 ℹ Project generated!
 ```
 
-**Note:** If you prefer to create the project with default parameters (i.e. --license='MIT'), you can run the command as `boost new:project booster-blog --default`. Keep
-in mind that the default cloud provider is AWS!
+**Note:** If you prefer to create the project with default parameters, you can run the command as `boost new:project booster-blog --default`. The default
+parameters are as follows:
+- Project name: The one provided when running the command, in this case "booster-blog"
+- Provider: AWS
+- Description, author, homepage and repository: ""
+- License: MIT
+- Version: 0.1.0
 
 > Booster CLI commands follow this structure: `boost <subcommand> [<flags>] [<parameters>]`. 
 > Let's break down the command we have just executed:
--- a/docs/README.md
+++ b/docs/README.md
@@ -334,6 +334,9 @@ configured [here](#set-up-an-aws-account).
 ℹ Project generated!
 ```
 
+**Note:** If you prefer to create the project with default parameters (i.e. --license='MIT'), you can run the command as `boost new:project booster-blog --default`. Keep
+in mind that the default cloud provider is AWS!
+
 > Booster CLI commands follow this structure: `boost <subcommand> [<flags>] [<parameters>]`. 
 > Let's break down the command we have just executed:
 >
--- a/packages/cli/src/commands/new/project.ts
+++ b/packages/cli/src/commands/new/project.ts
@@ -47,7 +47,7 @@ export default class Project extends Command {
     default: flags.boolean({
       char: 'd',
       description: 'generates the project with default parameters (i.e. --license=MIT)',
-      default: true,
+      default: false,
     }),
   }
 
--- a/packages/cli/src/commands/new/project.ts
+++ b/packages/cli/src/commands/new/project.ts
@@ -44,6 +44,11 @@ export default class Project extends Command {
       description:
         'package name implementing the cloud provider integration where the application will be deployed (i.e: "@boostercloud/framework-provider-aws"',
     }),
+    default: flags.boolean({
+      char: 'd',
+      description: 'generates the project with default parameters (i.e. --license=MIT)',
+      default: true,
+    }),
   }
 
   public static args = [{ name: 'projectName' }]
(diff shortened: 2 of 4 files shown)


packages/cli/src/commands/new/project.ts
```diff
@@ -44,6 +44,11 @@ export default class Project extends Command {
       description:
         'package name implementing the cloud provider integration where the application will be deployed (i.e: "@boostercloud/framework-provider-aws"',
     }),
+    default: flags.boolean({
+      char: 'd',
+      description: 'generates the project with default parameters (i.e. --license=MIT)',
+      default: true,
```
NickSeagull, line 50: Shouldn't this be `default: false`? If I recall correctly, `default` makes the flag enabled by default if it is not specified, which is not what we need

NickSeagull (changes requested):

packages/cli/src/commands/new/project.ts
juanjoman: Whoops you're right!!

NickSeagull (approved):
Thanks!

docs/README.md
```diff
@@ -334,6 +334,9 @@ configured [here](#set-up-an-aws-account).
 ℹ Project generated!
 ```
 
+**Note:** If you prefer to create the project with default parameters (i.e. --license='MIT'), you can run the command as `boost new:project booster-blog --default`. Keep
+in mind that the default cloud provider is AWS!
```
alvaroloes, line 338: I would extend this explanation and tell the user which are the default values of all parameters, Wouldn't you?

packages/cli/src/commands/new/project.ts
```diff
@@ -107,6 +112,21 @@ export const parseConfig = async (
   flags: Partial<ProjectInitializerConfig>,
   boosterVersion: string
 ): Promise<ProjectInitializerConfig> => {
+  if (flags.default) {
+    return Promise.resolve({
+      projectName: flags.projectName as string,
```
alvaroloes, line 117: Why do we need a cast here?

alvaroloes (approved):
Approved. Although I left two comments. The most important is the one regarding docs

packages/cli/src/commands/new/project.ts
juanjoman: I don't think it's needed, I'll check and remove it if it's not. Thanks!!

juanjoman: Oh I saw what happened here. `projectName` is `string | undefined` so it doesn't fit the `ProjectInitializerConfig` if we don't cast it to be a string. There is another way of avoiding this? @alvaroloes

docs/README.md
juanjoman: Changed!

packages/cli/src/commands/new/project.ts
alvaroloes: 🤔  aha! That's because flags type is `Partial<ProjectInitializerConfig>`. Maybe we could create a better type with the name of the project not "partial" but don't worry now about this ;-)

BTW, the exact way to tell Typescript "Believe me, this is not null" is with the bang operator `!`. I think it is cleaner, but I'm not sure if it is discouraged by our linter (although we get the same result)

alvaroloes (approved):
Re-approved! Thanks!

packages/cli/src/commands/new/project.ts
javiertoledo: The thing is that this cast would make the CLI try to create a new project named "undefined" if for some reason the project name is not present, so it shouldn't be optional at this point. You can fix it checking that `projectName` is defined before, TypeScript knows that you checked and let you assign the variable without casting. We're checking for its nullity in line 59 of this same file, in the `run` method. You could put the `projectName` in a string typed variable from that point.

javiertoledo (approved):
Nice code, take a look to the comment if you want to nitpick type errors

packages/cli/src/commands/new/project.ts
juanjoman: Hmm I've tried to:
1. Create a constant with the project name and pass it through param (removing it from `ProjectInitializerConfig`). The Script.init(..).step(...).... doesn't accept that.
2. Setting the variable with a cast after the null check. It still says that it should be `string | undefined`

cc @javiertoledo

javiertoledo:
integration: all

javiertoledo:
integration: local

PR #413, opened 2020-09-12
Updated yarn.lock and fixed errors to address security patches

javiertoledo:
Rebuilt yarn.lock to get the latest compatible versions of all dependencies


javiertoledo:
integration: all

packages/framework-provider-local/src/services/event-registry.ts
```diff
@@ -10,7 +10,10 @@ export class EventRegistry {
 
   public async query(query: object): Promise<Array<EventEnvelope>> {
     const queryPromise = new Promise((resolve, reject) =>
-      this.events.find(query, (err, docs) => err ? reject(err) : resolve(docs))
+      this.events.find(query).exec((err, docs) => {
+        if (err) reject(err)
+        else resolve(docs)
```
javiertoledo, line 15: For some reason, this was failing to infer the right type for `docs`, so I changed it to use the `exec` API instead of the projection function. I'm not sure that using the projection function is the right place to "get the results", it looks more like a "map" function to transform the data before modifying the query in the next steps. According to the doc comments, using `exec` seems to be the "right way" to run a nedb query...

javiertoledo: Oh, the linter was also trolling me with a weird loop, suggesting me to add parentheses to the ternary expression and when saying that the parentheses were not needed, so I ended up rewriting it with an `if` statement 🤷

juanjoman (approved):

PR #414, opened 2020-09-13
Infrastructure rockets: the way to extend Booster infrastructure in AWS. (Also transformed the static site deployment feature into the first rocket)

javiertoledo:
In this PR we introduce the concept of "rockets" as the way to extend Booster. It's likely that we start to use this pattern more and more to make the framework more configurable and easier to extend.

Some notes and ideas to take into account before reviewing this PR:
* Infrastructure rockets must be loaded dynamically to avoid importing them in the user's project, as that would require them to include the whole CDK in the deployed lambdas.
* I expect this change to slightly accelerate the deployments in the framework integration tests.
* One key of this change is removing the static page deployment feature from the framework completely and move it to a separate project that does not interfere with the development of the framework. 
* The next logical change would be removing the plugin from this repository and move it to its own one, but we'll keep the first plugins in the Booster monorepo until the feature is stabilized.
* If we want to provide a way to extend the runtime too, rocket developers will need to provide two packages, one that's loaded in the context of the user's project, the other, loaded dynamically on deploy time (as the one in this PR).
* Rockets are dependent on a specific provider because different providers have different ways to extend the infrastructure. In this PR we only changed the AWS provider to have a constructor function that accepts the plugins.
* One great thing about this implementation is that it makes it relatively easy for advanced users to manipulate the project stack adding any resources they want with the AWS CDK.

### TODOs

- [x] Deployment of resources defined by the plugin
- [x] Nuking resources deployed by the plugin
- [x] Implement tests for the plugins load process in AWS and AWS infra packages
- [x] Implement tests for the passing of the plugins through the different layers of AWS infrastructure package
- [x] Implement tests for deploying the resources defined by the plugin
- [x] Implement tests for destroying the resources defined by the plugin
- [x] Implement tests for ApplicationStackBuilder management of plugins
- [x] Add documentation about infrastructure plugins
~~- [ ] Implement an integration test with a simpler "reference rocket" that demonstrates that the stack extensions are deployed.~~


packages/framework-provider-aws-infrastructure/src/index.ts
```diff
@@ -1,6 +1,12 @@
+import { ProviderInfrastructure, PluginDescriptor } from '@boostercloud/framework-types'
+import { loadPlugin } from './infraestructure-plugin'
 import { deploy, nuke } from './infrastructure'
+export { InfrastructurePlugin } from './infraestructure-plugin'
```
NickSeagull, line 4: Typo here, its `infrastructure`

```suggestion
export { InfrastructurePlugin } from './infrastructure-plugin'
```

packages/framework-provider-aws-infrastructure/src/infraestructure-plugin.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export interface InfrastructurePlugin {
+  mountStack: (config: BoosterConfig, stack: Stack) => void
+}
+
+export function loadPlugin(pluginDescriptor: PluginDescriptor): InfrastructurePlugin {
+  const pluginBuilder = require(pluginDescriptor.packageName)
+  if (!pluginBuilder)
+    throw new Error(
+      `Could not load the plugin package ${pluginDescriptor.packageName}. Make sure you installed it in your project devDependencies.`
```
NickSeagull, line 12:
```suggestion
      `Could not load the plugin package ${pluginDescriptor.packageName}. Make sure you've installed it in your project devDependencies.`
```

packages/framework-provider-aws-infrastructure/src/infrastructure/index.ts
```diff
@@ -11,6 +11,7 @@ import { Observable, Subscriber } from 'rxjs'
 import { RequireApproval } from 'aws-cdk/lib/diff'
 import * as colors from 'colors'
 import { emptyS3Bucket } from './s3utils'
+import { InfrastructurePlugin } from '../infraestructure-plugin'
```
NickSeagull, line 14:
```suggestion
import { InfrastructurePlugin } from '../infrastructure-plugin'
```

packages/framework-provider-aws-infrastructure/test/infrastructure/deploy.test.ts
```diff
@@ -118,7 +118,7 @@ describe('the deployment process', () => {
 
       // Just checks that the assemble method does not fail, meaning that the stack is build correctly according to the
       // AWS validations
-      expect(() => Infrastructure.assemble(config)).not.to.throw()
+      expect(() => Infrastructure.assemble(config, [])).not.to.throw()
```
NickSeagull, line 121: Perhaps its better to keep consistency in parameter order, and make `assemble` to accept the array as a first parameter, like `deploy`?

packages/framework-provider-aws-infrastructure/test/infrastructure/stacks/application-stack.test.ts
```diff
@@ -116,6 +110,7 @@ describe('the application stack builder', () => {
     // Lambdas
     expect(numberOfLambdas).to.equal(4)
     expect(appStack.tryFindChild(preSignUpValidator)).not.to.be.undefined
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
NickSeagull, line 113: At this point, I'd remove the `any` and would type `lambda` as `lambda: { environment: Record<string, string> }`

NickSeagull (changes requested):

alvaroloes:
It is a great start Javi! I like the approach.

I see that plugins in this PR can only change/add/modify infrastructure. I envision that plugins would also need to add functionality to the provider runtime and the framework core.

I would try to put together a way of allowing plugins in the three Booster areas (core, provider runtime, provider infrastructure) based on the approach you have followed here (although not in such detail to begin with)

javiertoledo:
I think it would be nice to avoid implementing core or runtime plugins until we actually have a scenario where that's needed. Maybe a good one would be moving the current auth implementation to a plugin, but I think that we would need to agree on how we want to manage the auth tokens in Booster first.

packages/framework-provider-aws-infrastructure/test/infrastructure/deploy.test.ts
javiertoledo: I'll do the opposite because it makes sense that the array of plugins is optional.

packages/framework-provider-aws-infrastructure/test/infrastructure/stacks/application-stack.test.ts
javiertoledo: I tried, but it's not that easy in practice 😬

NickSeagull:
Hello, how's this going? I'm super excited about this feature 😄

javiertoledo:
Going fine, just completing the tests and docs :-D

juanjoman (approved):
LGTM!! Just left some nit-picky comments 👏

packages/framework-provider-aws/src/index.ts
```diff
@@ -28,50 +28,65 @@ import {
...
-  connections: {
-    storeData: storeConnectionData.bind(null, dynamoDB),
-    fetchData: fetchConnectionData.bind(null, dynamoDB),
-    deleteData: deleteConnectionData.bind(null, dynamoDB),
-    sendMessage: sendMessageToConnection,
-  },
-  // ProviderInfrastructureGetter
-  infrastructure: () =>
-    require(require('../package.json').name + '-infrastructure').Infrastructure as ProviderInfrastructure,
+/* We load the infrastructure package dinamically here to avoid including it in the
```
NickSeagull, line 31: Typo 
```suggestion
/* We load the infrastructure package dynamically here to avoid including it in the
```

NickSeagull (approved):
Looks awesome, I'm hyped to start using this!

Wanted to create a Discord bot using Booster, but then realized that I would need a webhook rocket 👀

packages/framework-provider-aws-infrastructure/test/infrastructure/nuke.test.ts
```diff
@@ -0,0 +1,321 @@
...
+        info: fake(),
+        error: console.error,
+      } as unknown) as Logger
+
+      await nukeModule.nuke(config, logger)
+
+      expect(StackServiceConfiguration.getStackServiceConfiguration).to.have.been.calledWithMatch(config)
+
+      revertNukeToolkit()
+      revertNukeApp()
```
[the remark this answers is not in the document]
javiertoledo: Not really, because all these functions are local instances initialized within the test case 😜

javiertoledo:
As there are some known issues in the way integration tests run with breaking changes, and given the size of this PR, I think that it makes sense to release this functionality now and implement an integration test in a future PR, so I'm setting this one ready for review!

NickSeagull (approved):

packages/framework-provider-aws/src/index.ts
```diff
@@ -34,55 +34,70 @@ import { rawScheduledInputToEnvelope } from './library/scheduled-adapter'
...
+      fetchData: fetchConnectionData.bind(null, dynamoDB),
+      deleteData: deleteConnectionData.bind(null, dynamoDB),
+      sendMessage: sendMessageToConnection,
+    },
+    // ScheduledCommandsLibrary
+    scheduled: {
+      rawToEnvelope: rawScheduledInputToEnvelope,
+    },
+    // ProviderInfrastructureGetter
+    infrastructure: () => loadInfrastructurePackage().Infrastructure(rockets),
```
alvaroloes, line 99: What a 💣

packages/framework-provider-aws/test/index.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      it('returns a `ProviderLibrary` object', () => {
+        expect(providerLibrary).to.be.an('object')
+        expect(providerLibrary.api).to.be.an('object')
+        expect(providerLibrary.auth).to.be.an('object')
+        expect(providerLibrary.connections).to.be.an('object')
+        expect(providerLibrary.events).to.be.an('object')
+        expect(providerLibrary.graphQL).to.be.an('object')
+        expect(providerLibrary.infrastructure).to.be.a('function')
+        expect(providerLibrary.readModels).to.be.an('object')
+      })
```
alvaroloes, line 26: This test is not needed. As everything is under the Typescript compiler realm, it is impossible to return different types on each of these fields or to return something different than `ProvidersLibrary` from `AWSProvider()`

packages/framework-provider-aws/test/index.test.ts
```diff
@@ -0,0 +1,69 @@
...
```
alvaroloes, line 58: Same comment as before. The compiler got us covered here!

packages/rocket-static-sites-aws-infrastructure/.mocharc.yml
```diff
@@ -0,0 +1,10 @@
+diff: true
+require: 'ts-node/register'
+extension:
+  - ts
+package: './package.json'
+recursive: true
+reporter: 'spec'
+timeout: 5000
+full-trace: true
+bail: true
```
alvaroloes, line 10: Curious: Shouldn't this be inside the test folder and not in the root?

packages/rocket-static-sites-aws-infrastructure/src/static-website-stack.ts
```diff
@@ -0,0 +1,76 @@
+import { CfnOutput, RemovalPolicy, Stack } from '@aws-cdk/core'
```
alvaroloes, line 1: Reviewed until here. Long but AMAZING PR. I'll keep reviewing tomorrow.

alvaroloes (commented):
So far so great!

packages/framework-provider-aws-infrastructure/test/index.test.ts
```diff
@@ -0,0 +1,153 @@
...
+          if (providerInfrastructure.deploy)
+            await expect(providerInfrastructure.deploy(fakeConfig, logger)).not.to.eventually.be.rejected
+
+          // It receives the thrown Error object, not just the message
+          expect(logger.error).to.have.been.calledWithMatch(errorMessage)
+        })
+      })
+
+      describe('nuke', () => {
+        xit('initializes nuke with no rockets')
```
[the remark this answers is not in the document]
javiertoledo: It means "pending test" and mocha shows it as pending in the test list with a different color when they're run 😎

docs/README.md
```diff
@@ -2569,6 +2569,94 @@ Booster environments are extremely flexible. As shown in the first example, your
...
+  unmountStack: SomePrivateObject.unmountStack.bind(null, params),
+})
+
+export default YourRocketInitializator
+```
+
+In `mountStack` you will receive an initialized AWS CDK stack that you can use to append new resources. Check out [the Stack API in the official CDK documentation](https://docs.aws.amazon.com/cdk/latest/guide/stacks.html#stack_api). This is the same stack instance that Booster uses to deploy its resources, so your resources will be deployed automatically along with the Booster's ones in the same stack.
+
+The application stack, including the resources added by your rocket are automatically nuked along with the application stack, but there are some situations on which it's convenient to delete or move the contents of the resources created by you. In the appl `unmountStack` you'll have the opportunity to run any code before deleting the stack. This function receives an `utils` object with the same tools that Booster uses to perform common actions like emptying the contents of an S3 bucket (Non-empty buckets are kept by default when a stack is deleted).
+
```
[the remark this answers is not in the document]
javiertoledo: That makes a lot of sense!

packages/framework-provider-aws/test/index.test.ts
javiertoledo: The compiler is great, and you're probably right, but who trusts what happens in the JS runtime realm!!?? :trollface:

packages/rocket-static-sites-aws-infrastructure/src/static-website-stack.ts
alvaroloes: LOL, when I said tomorrow, I meant, next week 😂 
Resuming the review!

packages/rocket-static-sites-aws-infrastructure/src/static-website-stack.ts
```diff
@@ -0,0 +1,76 @@
...
+  CloudFrontWebDistribution,
+  OriginAccessIdentity,
+  ViewerCertificate,
+  ViewerProtocolPolicy,
+} from '@aws-cdk/aws-cloudfront'
+import { existsSync } from 'fs'
+import { RocketUtils } from '@boostercloud/framework-provider-aws-infrastructure'
+
+export type AWSStaticSiteParams = {
+  localDistPath?: string
```
alvaroloes, line 14: Where does the `localDist` part come from? 

From the code, this seems to be the root folder of the static file. What do you think about calling it `rootPath`?

packages/rocket-static-sites-aws-infrastructure/src/static-website-stack.ts
```diff
@@ -0,0 +1,76 @@
...
+    const indexFile = params.indexFile ?? 'index.html'
+    const errorFile = params.errorFile ?? '404.html'
+    if (existsSync(localDistPath)) {
+      const staticWebsiteOIA = new OriginAccessIdentity(stack, 'staticWebsiteOIA', {
+        comment: 'Allows static site to be reached only via CloudFront',
+      })
+      const staticSiteBucket = new Bucket(stack, 'staticWebsiteBucket', {
+        websiteIndexDocument: indexFile,
+        websiteErrorDocument: errorFile,
+        bucketName: params.bucketName,
```
alvaroloes, line 32: I see the `bucketName` does not have a default name, but I think we could provide one and the user would not need to specify it.

What do you think about using `config.applicationName` + `static-site-bucket`?

alvaroloes (commented):
Finishe review!! I just left two more questions

alvaroloes (approved):

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/application-stack.ts
```diff
@@ -8,13 +8,18 @@ import { RestApi } from '@aws-cdk/aws-apigateway'
 import { CfnApi, CfnStage } from '@aws-cdk/aws-apigatewayv2'
 import { baseURLForAPI } from '../params'
 import { setupPermissions } from './permissions'
-import StaticWebsiteStack from './static-website-stack'
+import { InfrastructureRocket } from '../../rockets/infrastructure-rocket'
+
+// Extracted this here to be able to mock it, it's surprisingly hard to mock a constructor!
+function buildStack(app: App, applicationStack: string, props?: StackProps): Stack {
+  return new Stack(app, applicationStack, props)
+}
```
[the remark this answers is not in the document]
javiertoledo: 🤔 this is a good idea, but that'd require a couple of changes, I'll leave it for another PR. I created the issue #481 to rework this file.

packages/rocket-static-sites-aws-infrastructure/.mocharc.yml
javiertoledo: It looks like we have this file in the package root in each of the packages 🤔 

I guess that mocha will look for it recursively, If you think we should move it, let's create an issue to do this in a separate PR in all packages!

packages/rocket-static-sites-aws-infrastructure/src/static-website-stack.ts
javiertoledo: This would require many changes because in the current implementation we don't have access to `config`. I've created issue #482 to tackle this.

javiertoledo:
Integration tests failed, I think it's because this branch doesn't have the changes in #476, so I'd wait for that PR to be merged before trying this again.

packages/framework-integration-tests/integration/cli/cli.project.integration.ts
```diff
@@ -297,8 +297,23 @@ describe('Project', () => {
...
+
+          for (const packageName of ['cli', 'framework-provider-aws-infrastructure']) {
+            await exec(`npm install -D ../../../${packageName}`, { cwd: projectPath(projectName) })
+          }
+
+          await exec('npm install', { cwd: projectPath(projectName) })
+          // For some reason, the project doesn't compile if we don't re-compile the modules.
+          // Maybe rewriting the dependencies messes up lerna's tricks ¯\_(ツ)_/¯
+          await exec('lerna clean --yes && lerna bootstrap && lerna run clean && lerna run compile')
+
```
javiertoledo, line 314: This is crazy slow. For some reason, I was forced to rebuild all modules to make the project compile, but with this change, we're using the code under development instead of the libraries installed from npm. I don't know if there are other places in the code that would need to use this trick. It's very likely that we could optimize this in the future, but for this PR, this is more than enough...

packages/cli/src/templates/project/config-ts.ts
```diff
@@ -4,6 +4,6 @@ import { Provider } from '{{{providerPackageName}}}'
 
 Booster.configure('production', (config: BoosterConfig): void => {
   config.appName = '{{{ projectName }}}'
-  config.provider = Provider
+  config.provider = Provider()
```
javiertoledo, line 7: I was forced to make all providers export a function to avoid breaking the project generator. Having a heterogeneous interface like the one I proposed makes code generation way harder.

PR #426, opened 2020-09-23
Delete read models when projecting null

juanjoman:
**Note:** This is a copy from https://github.com/boostercloud/booster/pull/404, but fixing CLI integration tests
## Description
With this change, a developer will be able to delete their read models by returning the `deleteReadModel` type. Here's an example:
```
@ReadModel
export class UserReadModel {
  public constructor(readonly username: string, /* ...(other interesting fields from users)... */) {}
  
  @Projects(User, 'id')
  public static projectUser(entity: User, current?: UserReadModel): ProjectionResult<UserReadModel> {
    if (isReadModelDeletable()) {
        return deleteReadModel // This is a type from ProjectionResult
    }
    return new UserReadModel(...)
  }

  private isReadModelDeletable(): boolean {
    ...
  }
```

## Changes
1. Insert a `deleteReadModel` in the `read-model-store` flow, which ends up calling `read-model-adapter` and deleting a read model with a specific id
2. Changed the return value of projections from `TReadModel` to `ProjectionResult<TReadModel>`. This way we can set more values in the future if needed and we also don't return null, which is not that meaningful
3. Modified the `read-model` template to adapt it to the new `ProjectionResult` type
4. Added integration and unit testing
5. Updated the docs

## Checks
- [X] Project Builds

**Note2:** There is something going on on my local environment with integration tests, I'm going to run them here just to know where the issue is

### DIFF
--- a/packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
+++ b/packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
@@ -307,22 +307,12 @@ describe('Cart end-to-end tests', () => {
               `,
             })
           },
-          (result) => result?.data?.ProductReadModel?.deleted && result?.data?.ProductReadModel?.id === productId
+          () => true
         )
 
         const productData = queryResult.data.ProductReadModel
-        const expectedResult = {
-          __typename: 'ProductReadModel',
-          id: productId,
-          sku: '<DELETED>',
-          displayName: '',
-          description: '',
-          price: null,
-          availability: 0,
-          deleted: true,
-        }
 
-        expect(productData).to.be.deep.equal(expectedResult)
+        expect(productData).to.be.null
       })
     })
   })
--- a/packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
+++ b/packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
@@ -11,6 +11,7 @@ import { random, commerce, finance, lorem, internet } from 'faker'
 import { expect } from 'chai'
 import gql from 'graphql-tag'
 import { CartItem } from '../../src/common/cart-item'
+import { sleep } from '../../integration/providers/helpers'
 
 describe('Cart end-to-end tests', () => {
   let client: ApolloClient<NormalizedCacheObject>
@@ -283,6 +284,9 @@ describe('Cart end-to-end tests', () => {
             }
           `,
         })
+
+        console.log('Waiting 1 second for deletion to complete...')
+        await sleep(1000)
 
         client = await graphQLClient(userAuthInformation.accessToken)
         // Retrieve updated entity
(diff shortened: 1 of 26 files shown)


juanjoman (commented):
bot: integration

juanjoman (commented):
bot: integration

alvaroloes:
@juanjoman Are you having problems with integration tests? It seems the checks are not appearing...

juanjoman (commented):
integration: all

javiertoledo:
For some reason, the integration tests were skipped: https://github.com/boostercloud/booster/actions/runs/268659144

javiertoledo (approved):
Looks great!

javiertoledo:
integration: all

juanjoman (commented):
About those two integration tests failing:

1. AWS functionality: This requires the project to be deployed. Maybe we are running it separately? In that case, it just says that the project doesn't exist and fails. Otherwise, it passes (checked on local)
2. AWS end-to-end: Working on local

By the way, I can't see the errors in the detail section of the checks, can you? @alvaroloes @javiertoledo

NickSeagull (approved):
Happy that we finally aren't relying on `null` and instead on ReadModel.Delete

juanjoman (commented):
integration: all

juanjoman:
integration: all

juanjoman:
integration: all

juanjoman:
integration: all

juanjoman:
I really don't understand anything. End to end fails and AWS functionality integration is about to fail: https://github.com/boostercloud/booster/actions/runs/284405266

I ran these (unit and integration) on local today and were passing 🤷‍♂️

juanjoman:
integration: all

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -284,6 +285,9 @@ describe('Cart end-to-end tests', () => {
           `,
         })
 
+        console.log('Waiting 1 second for deletion to complete...')
+        await sleep(1000)
```
juanjoman, line 289: I had to add this second because it seems like querying was "happening before deletion" or there was a non 100% sync flow. :wat: cc @javiertoledo @alvaroloes @NickSeagull

NickSeagull: Wouldn't this work with a `waitForIt`?

juanjoman: hmm let me try, I didn't see that the delete doesn't have a waitForIt, that may be it

juanjoman: nope it doesn't seem to work @NickSeagull

NickSeagull: Got it, I wouldn't spend much more time on this if it's working

alvaroloes:
@juanjoman I guess you can already merge this pull request and publish the release notes, correct?

PR #427, opened 2020-09-24
Refactor: Removed the Observable interface for deploy and nuke scripts

javiertoledo:
I've replaced the annoying `Observable` interface in deploy and nuke scripts with a simpler logger class. There are several reasons why this change makes sense, and it's a refactor that I've been wanting to do for a while:

* This interface has been causing errors in tests
* It forced some weird designs in the provider implementations and affects all providers.
* It required to import the whole `rxjs` library in the project (it's still used by some dependencies though)

I also fixed/improved a couple of tests.


javiertoledo:
integration: all

NickSeagull (approved):
Lovely, thanks!

javiertoledo:
integration: all

PR #438, opened 2020-10-07
Enable CORS in the auth endpoints

### LINKED ISSUE #387: Enable CORS in the auth endpoints

javiertoledo:
> Notice that this PR was recreated after rebasing PR #388

## Description
This PR resolves #387 

## Changes
The Auth Stack generates the auth API with preflight support (accept OPTIONS requests) for the `auth/sign-up`, `auth/sign-in` ,and `auth/sign-out` endpoints. Also, the corresponding `POST` methods return the `Access-Control-Allow-Origin: *' header.

The integration test verifies the OPTIONS requests for almost all the used HTTP verbs. Regarding the Posts request, there are three tests pending because it is difficult to test the response for 500 errors.

Note that this solves the issue only for the AWS provider. However, I wrote the test suite as a black box one, so we can move it up and run it against all the providers' implementations. At this moment the best would be open tickets for reproducing the solution for each provider or wait until someone needs it.

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-integration-tests/integration/providers/aws/cors.integration.ts
+++ b/packages/framework-integration-tests/integration/providers/aws/cors.integration.ts
@@ -3,28 +3,30 @@ import { internet } from 'faker'
 import fetch from 'cross-fetch'
 import { expect } from '@boostercloud/framework-provider-aws/test/expect'
 
-describe('The Authentication API', () => {
+describe('Given the Authentication API', () => {
   let clientId: string
-  const methodsToCheck = ['GET', 'POST', 'PUT', 'DELETE', 'OPTIONS', 'PATCH']
-  const preflightOptions = generatePreflightOptionsList(methodsToCheck)
-
+  const username = internet.email()
+  const password = createPassword()
+  const role = 'SuperUserNoConfirmation'
   before(async () => {
     clientId = await authClientID()
   })
 
-  context('/auth/sign-up', () => {
+  context('When /auth/sign-up', () => {
     let signUpUrl: string
-    let validSignUpBody: string
-    const invalidSignUpBody = JSON.stringify({})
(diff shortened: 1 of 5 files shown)


javiertoledo:
All tests passed locally, so I'm merging this feature

Issue #156, opened 2020-06-05
`--verbose` flag when creating new projects?


NickSeagull:
It'd be great if we could see what's going on when Booster is installing dependencies.

dotalo:
Adding a `--skip-install ` or `--no-install` flag for skipping the installation of dependencies may also be useful.

PR #444, opened 2020-10-11
feat(cli): implement --skipInstall flag

Fecony:
## Description
CLI flag to skip dependency installation. According to https://github.com/boostercloud/booster/issues/156#issuecomment-667982702 

## Checks
- [x] Project Builds
 
## Additional information
There is no documentation about CLI commands

### DIFF
--- a/packages/cli/src/commands/new/project.ts
+++ b/packages/cli/src/commands/new/project.ts
@@ -49,7 +49,7 @@ export default class Project extends Command {
       description: 'generates the project with default parameters (i.e. --license=MIT)',
       default: false,
     }),
-    'skip-install': flags.boolean({
+    skipInstall: flags.boolean({
       description: 'skip dependencies installation',
       default: false,
     }),
@@ -78,7 +78,7 @@ const run = async (flags: Partial<ProjectInitializerConfig>, boosterVersion: str
     .step('Creating project root', generateRootDirectory)
     .step('Generating config files', generateConfigFiles)
 
-  if (flags['skip-install']) {
+  if (flags.skipInstall) {
     return boostScript.info('Project generated!').done()
   }
 
@@ -135,7 +135,7 @@ export const parseConfig = async (
       repository: '',
       boosterVersion,
       default: flags.default,
-      'skip-install': false,
+      skipInstall: false,
     })
   }
 
@@ -169,6 +169,6 @@ export const parseConfig = async (
     repository,
     boosterVersion,
     default: false,
-    'skip-install': false,
+    skipInstall: false,
   })
 }
--- a/packages/cli/src/commands/new/project.ts
+++ b/packages/cli/src/commands/new/project.ts
@@ -49,6 +49,10 @@ export default class Project extends Command {
       description: 'generates the project with default parameters (i.e. --license=MIT)',
       default: false,
     }),
+    'skip-install': flags.boolean({
+      description: 'skip dependencies installation',
+      default: false,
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/new/project.ts
```diff
@@ -49,6 +49,10 @@ export default class Project extends Command {
       description: 'generates the project with default parameters (i.e. --license=MIT)',
       default: false,
     }),
+    'skip-install': flags.boolean({
```
NickSeagull, line 52: I'd make this `skipInstall` to keep the same casing as the other options

packages/cli/src/commands/new/project.ts
```diff
@@ -69,13 +73,20 @@ export default class Project extends Command {
...
+
+  if (flags['skip-install']) {
+    return boostScript.info('Project generated!').done()
+  }
+
+  return boostScript
     .step('Installing dependencies', installDependencies)
     .info('Project generated!')
     .done()
+}
```
NickSeagull, line 89: Lovely! Perhaps we can make this a bit shorter like:
```suggestion
  Script.init(`boost ${Brand.energize('new')} 🚧`, parseConfig(new Prompter(), flags, boosterVersion))
    .step('Creating project root', generateRootDirectory)
    .step('Generating config files', generateConfigFiles)
    .step(
      'Installing dependencies', 
      (config: BoosterConfig) => flags.skipInstall ? undefined : installDependencies(config)
    )
    .info('Project generated!')
    .done()
}
```

NickSeagull (changes requested):
Thanks a lot for this! I just left a couple of comments, also:

* Could you please add an explanation of feature to the `docs/README.md` file?
* Also, could you please add a test to `cli/test/commands/new.test.ts` ensuring that the `installDependencies` function is **not** called? You can copy and paste one of the other tests as a starting point 😄

Fecony:
@NickSeagull 

- [x] Added test 
- [x] Changed flag name to match other flags naming

>Could you please add an explanation of feature to the docs/README.md file?
There is no CLI section in documentation 😥

packages/cli/src/commands/new/project.ts
Fecony: Tried to do this, but `step` method will print message that it is installing dependencies. 
Looks like we need to change `step` method to skip this step if action returns undefined

NickSeagull (approved):
Awesome! LGTM

alvaroloes:
Hi @Fecony, thanks a lot for contributing. Just one question: 
When using this flag would be useful?

Fecony:
Hey, @alvaroloes, actually it depends. 

Some people don't install dependencies because they don't want to wait forever or even few seconds. 
Maybe they want to just take a look at cod, play with cli commands and framework itself (without running it). 

Or people may just be sitting without internet connection 😄
In this case they may have other project with node_modules that they can link

alvaroloes (approved):

alvaroloes:
Ok, thanks for answering! 
Yeah, my point was that 99% of the time you would run `npm install` after generating the project, as it would not compile otherwise. Booster CLI is leaned towards trying to do as much work for you as it can, so that's the reason why even dependencies are installed.

However, I can see that, in some situations, it could be useful as you say (probably for advanced users)
All right!

javiertoledo (approved):
Nice, thanks!

javiertoledo:
integration: all

Issue #449, opened 2020-10-12
Integration test for CLI don't work


VeroGarp:
## Current Behavior

Integration test don't pass on master

## Expected behavior

Integration test pass

## Possible Solution

No idea

## Additional information

See the attached file
[integration_test_error.txt](https://github.com/boostercloud/booster/files/5366815/integration_test_error.txt)

#### Environment

- Booster version: master
- Node/npm version: v14.13.1/6.14.8
- OS: Windows 10

javiertoledo:
Will be fixed in #476

PR #450, opened 2020-10-12
fix(cli): Generate file names with proper casing

### LINKED ISSUE #306: Make generators generate file names with proper casing
Right now, the generators are generating filenames with the same casing as the class.

This is not the standard for JS/TS projects, as they should be kebab-case instead of PascalCase.

This task is a matter of 

- Going into each of the generators
- In the call to `generate`, replace `info.name` with the appropriate call to the `inflections` library

VeroGarp:
## Description

Fixes #306

## Changes

Used inflection for modify the names of files with underscore function, and then replace underscore by dashes.

## Checks
- [x] Project Builds
 
## Additional information

Integration tests for CLI dont work, see #449


packages/cli/src/commands/new/entity.ts
```diff
@@ -60,10 +61,13 @@ const run = async (name: string, rawFields: Array<string>, rawEvents: Array<stri
     .done()
 
 function generateImports(info: EntityInfo): Array<ImportDeclaration> {
-  const eventsImports: Array<ImportDeclaration> = info.events.map((eventData) => ({
-    packagePath: `../events/${eventData.eventName}`,
-    commaSeparatedComponents: eventData.eventName,
-  }))
+  const eventsImports: Array<ImportDeclaration> = info.events.map((eventData) => {
+    const fileName = inflection.underscore(eventData.eventName).replace(/_/g, '-')
```
javiertoledo, line 65: Oh! When I saw this line I thought that `inflection.dasherize` should do the job better, but I tried and... surprise! It doesn't! It looks like other people had the same issue and there are even PRs to solve it, but the repository seems abandoned https://github.com/dreamerslab/node.inflection/pull/72

You could avoid manually replacing underscores with dashes doing something like this:

```suggestion
    const fileName = inflection.transform( eventData.eventName, [ 'underscore', 'dasherize' ])
```

But maybe a better choice would be to try the library `inflected` (https://www.npmjs.com/package/inflected), which looks like a more up to date drop-in replacement. If you want to check if it works better, it could make sense to replace it in the whole project... (you'll likely need to do some other changes in other files, I bet the compiler will guide you, but if you need help, don't hesitate to ask in Discord 😉)

packages/cli/src/commands/new/entity.ts
```diff
@@ -15,6 +15,7 @@ import * as path from 'path'
 import { generate } from '../../services/generator'
 import { templates } from '../../templates'
 import { checkItIsABoosterProject } from '../../services/project-checker'
+import inflection = require('inflection')
```
javiertoledo, line 18: I think that you can avoid the call to `require` doing this:

```suggestion
import * as inflection from 'inflection'
```

javiertoledo (approved):
Hey @VeroGarp, thanks a lot for this, this change is 💯, super useful to keep consistency with the language standards. You also helped us to notice that we're using an inflection library that doesn't seem maintained anymore, so It could make a lot of sense to replace it with [inflected](https://www.npmjs.com/package/inflected) if it works better (I haven't tried it myself). If you want to change it in this PR I'd be happy to review it again, but if you don't have the time, could you please create a new issue to come back to this later? Thanks a lot!

packages/cli/src/commands/new/entity.ts
VeroGarp: Hi @javiertoledo ! I can't find the "transform" function so I use "dasherize" and "underscore". You can see it here: https://github.com/boostercloud/booster/blob/5069a978f11be9c41c0602e261b634a7c0e25a2f/packages/cli/src/commands/new/entity.ts#L65

packages/cli/src/commands/new/entity.ts
VeroGarp: Thanks! Fixed here: https://github.com/boostercloud/booster/blob/5069a978f11be9c41c0602e261b634a7c0e25a2f/packages/cli/src/commands/new/entity.ts#L18

packages/cli/src/commands/new/read-model.ts
```diff
@@ -61,10 +62,13 @@ const run = async (name: string, rawFields: Array<string>, rawProjections: Array
     .done()
 
 function generateImports(info: ReadModelInfo): Array<ImportDeclaration> {
-  const eventsImports: Array<ImportDeclaration> = info.projections.map((projection) => ({
-    packagePath: `../entities/${projection.entityName}`,
-    commaSeparatedComponents: projection.entityName,
-  }))
+  const eventsImports: Array<ImportDeclaration> = info.projections.map((projection) => {
+    const fileName = inflected.dasherize(inflected.underscore(projection.entityName))
```
NickSeagull, line 66: We are using the same code in 4 different places, can you create a new file under `cli/src/common` called `filenames.ts` that with a function called `classNameToFileName` that does this, but only in one place?

NickSeagull (approved):
Thanks for the PR 😄 

Can you please refactor all the equal code into a unique file, so we only have one place to change it? This will help maintaining the project

You can do this in another PR and just get this one merged

PR #451, opened 2020-10-13
fix cli integration tests

adriangmweb:
## Description
CLI integration tests are failing and the new version with the `ScheduledCommands` functionality can't be built. This was happening because the CLI tests were trying to export the new `boosterTriggerScheduledCommand` function from the core, but it didn't exist because it was installing the previous version of Booster `v8.1`. 

## Changes
This fix removes the export of that function, allowing the CLI tests to use the previous version, once this is merged I'll submit some changes to the `ScheduledCommands`, including the usage of these functions in the tests.

## Checks
- [x] Project Builds

### DIFF
--- a/packages/cli/src/templates/project/index-ts.ts
+++ b/packages/cli/src/templates/project/index-ts.ts
@@ -5,7 +5,6 @@ export {
   boosterPreSignUpChecker,
   boosterServeGraphQL,
   boosterNotifySubscribers,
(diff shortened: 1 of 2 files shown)


NickSeagull:
Thanks, this is very useful ❤️

NickSeagull:
integration: cli

NickSeagull (approved):

juanjoman (approved):
🥇

PR #452, opened 2020-10-13
Scheduled commands improvements

adriangmweb:
## Description
Included some improvements requested by @alvaroloes. 

## Changes
* Removed unnecessary flag for fields
* Now Scheduled Lamdba just gets created when there are scheduled commands defined.

## Checks
- [x] Project Builds


adriangmweb:
> One small suggestion, related to the weird API for scheduling, perhaps we can think in a way to do it more explicitly, with additional fields?
> 
> Something like:
> 
> ```ts
> params: {
>   eachHourAt: 5,
> }
> ```
> 
> or
> 
> ```ts
> params: {
>   every: {minutes: 5},
> }
> ```
> 
> ?

Honestly, I really like your second approach. I followed the way this is done by convention, but it would be really helpful for the final user to have an easier option!

NickSeagull:
Yeah, the "convention" is established by AWS, which in theory, the user shouldn't know anything of! 😂

Let's see what other @boostercloud/booster-core members think

javiertoledo:
I'd opt for [CRON expressions syntax](https://en.m.wikipedia.org/wiki/Cron). It's okay if you want to add some syntactic sugar on top of that, but for v1 I'd bet for the most standard way possible

juanjoman (approved):
LGTM! About the cronjob expressions, I think CRON syntax is well-known so for now I don't have a problem with keeping it.

NickSeagull:
Don't like CRON and friends at all, but if everyone agrees, so be it ¯\_(ツ)_/¯

alvaroloes:
Regarding the discussion about CRON or not: I would leave the options as they are right now, just as they are in CRON. They are not that friendly, but they are powerful and almost everybody that has dealt with recurrent tasks has probably needed to use them.

BUT! Remember that this is code. We or even the users can always create easier ways to specify the options. For example, this command:
```typescript
@ScheduledCommand({
  minute: '0/5',
})
export class CheckCartCount {
  // ...
}
```
Can be written this way:
```typescript
@ScheduledCommand(everyMinutes(5))
export class CheckCartCount {
  // ...
}
```
If we have the following function:
```typescript
function everyMinutes(minutes: number): ScheduleInterface {
  return { minutes: '0/' + minutes }
}
```

We can even export a "cron-helper.ts" with several functions like this to ease this task. This way, we have the powerfulness of raw corn expressions (and people that know it can use them) and _also_ simple functions for people that want simple time expressions.

What do you think?

alvaroloes (approved):

adriangmweb:
I think that might be useful, but I don't think this feature is going to be so used that we should spend more time on it right now. I would rather leave it like this, and if in the future we have to change it like if we implement it as a rocket, we will spend the time then.

adriangmweb:
integration: all

adriangmweb:
integration: all

NickSeagull:
Agree, let's get this merged :D

Issue #367, opened 2020-08-13
AWS::Lambda::Function pre-sign-up-validator Unzipped size must be smaller than 262144000 bytes


davidverdu:
While I'm deploying a Booster application this error is displayed and, consequently, a rollback occurs, preventing boost from creating all elements in AWS:

```text
26/57 | 11:12:17 | CREATE_FAILED        | AWS::Lambda::Function                  | pre-sign-up-validator (presignupvalidator8052074C) Unzipped size must be smaller than 262144000 bytes (Service: AWSLambdaInternal; Status Code: 400; Error Code: InvalidParameterValueException; Request ID: 2798c5eb-e996-4112-a8a8-439fa4f6636d)
	new Function (/booster_project/node_modules/@aws-cdk/aws-lambda/lib/function.ts:460:35)
	\_ AuthStack.buildUserPool (/booster_project/node_modules/@boostercloud/framework-provider-aws-infrastructure/dist/infrastructure/stacks/auth-stack.js:26:33)
	\_ AuthStack.build (/booster_project/node_modules/@boostercloud/framework-provider-aws-infrastructure/dist/infrastructure/stacks/auth-stack.js:19:35)
	\_ ApplicationStackBuilder.buildOn (/booster_project/node_modules/@boostercloud/framework-provider-aws-infrastructure/dist/infrastructure/stacks/application-stack.js:27:62)
	\_ assemble (/booster_project/node_modules/@boostercloud/framework-provider-aws-infrastructure/dist/infrastructure/index.js:46:61)
	\_ Object.synthesizer (/booster_project/node_modules/@boostercloud/framework-provider-aws-infrastructure/dist/infrastructure/index.js:72:44)
	\_ AppStacks.synthesizeStacks (/booster_project/node_modules/aws-cdk/lib/api/cxapp/stacks.ts:205:41)
	\_ AppStacks.listStacks (/booster_project/node_modules/aws-cdk/lib/api/cxapp/stacks.ts:179:33)
	\_ deployApp (/booster_project/node_modules/@boostercloud/framework-provider-aws-infrastructure/dist/infrastructure/index.js:59:38)
	\_ processTicksAndRejections (internal/process/task_queues.js:93:5)
```

I'm using MacOS and Booster 0.5.1. Here I attach my package.json:

```json
{
  "name": "booster_project",
  "description": "",
  "version": "0.1.0",
  "author": "",
  "dependencies": {
    "@boostercloud/framework-core": "^0.5.1",
[...]

PR #455, opened 2020-10-13
Pruning devDependencies before deploying

### LINKED ISSUE #441: Install dependencies in production mode before deploying
Right now we are just deploying the app as it is, but `npm` supports installing the deps in production mode, which halves the size of the zip at least by half.

sanskar-p:
## Description
Added a step before deploying to prune dependencies and added unit tests for the same.

Resolves #441 

## Changes
<Changes made> 
1. Added new step to remove dev dependencies before deployment
2. Added unit tests to check the above

## Checks
- [ ] Project Builds


javiertoledo (changes requested):
Thanks, nice change! I just left a couple of comments to improve the test styling to make sure we use a consistent style.

packages/cli/test/commands/deploy.test.ts
```diff
@@ -41,6 +41,33 @@ describe('deploy', () => {
...
+        })
+
+        const pruneDependenciesSpy = spy(deploy, deploy.pruneDependencies)
+        await runTasks('test-env', fakeLoader, fakeDeployer)
+        
+        expect(pruneDependenciesSpy).to.have.been.calledOnce
+
+        expect(ctx.stdout).to.include('Deployment complete')
+        expect(fakeDeployer).to.have.been.calledOnce
+      })
```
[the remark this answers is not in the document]
sanskar-p: That's right! It wasn't supposed to be there :sweat_smile:. I'll make the changes right away!

packages/cli/test/commands/deploy.test.ts
```diff
@@ -41,6 +41,33 @@ describe('deploy', () => {
       })
     })
 
+    context('when it prunes dev dependencies', () => {
```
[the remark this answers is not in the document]
sanskar-p: Thanks @javiertoledo, what do you suggest I change the description to? Or do you think I should just remove the `context` block entirely?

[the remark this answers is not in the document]
javiertoledo: Yes, I think that the `context` block could be redundant here, what do you think?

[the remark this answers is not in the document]
sanskar-p: I had written it to make the syntax consistent with the other tests in the file. But yeah, I also think it's redundant.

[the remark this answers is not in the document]
sanskar-p: @javiertoledo, check out the [`new commit`](https://github.com/boostercloud/booster/pull/455/commits/d19e206be8aae9284594bfed37bad4ccbe4b7a05)

NickSeagull:
Hey @sanskar-p the tests are failing, please run them using `lerna run test` and fix them

packages/cli/test/commands/deploy.test.ts
```diff
@@ -41,6 +41,29 @@ describe('deploy', () => {
...
+        appName: 'fake app',
+        region: 'tunte',
+        entities: {},
+      })
+
+      const fakeDeployer = fake((_config: unknown, logger: Logger) => {
+        logger.info('this is a progress update')
+      })
+
+      const pruneDependenciesSpy = spy(deploy, deploy.pruneDependencies)
```
NickSeagull, line 58: I believe this should be
```suggestion
      const pruneDependenciesSpy = spy(deploy, 'pruneDependencies')
```

packages/cli/test/commands/deploy.test.ts
```diff
@@ -41,6 +41,29 @@ describe('deploy', () => {
...
+      await runTasks('test-env', fakeLoader, fakeDeployer)
+        
+      expect(pruneDependenciesSpy).to.have.been.calledOnce
```
NickSeagull, line 61: Maybe this should be
```suggestion
      expect(deploy.pruneDependencies).to.have.been.calledOnce
```

alvaroloes:
@sanskar-p Hi! Thanks a lot for this. I didn't know about that command in npm 👏 

While I like the approach you followed, I have several concerns: 
1.- After deploying the application, the dev dependencies are gone, so the user won't be able to keep developing as usual until they run again `npm install`, which is a bit odd. 
2.- We are assuming that dependencies are installed so we "prune" them. But this is only true if the user is deploying from their development environment. If the deploy is done in a CI/CD system, there are no dependencies installed so we should run `npm install --production` instead of prune.

A possible solution for 1 we could add an extra step to restore dev dependencies, but it doesn't make sense if we are in a CI/CD environment.

Anyway, I _think_ (not sure though) that if you run `npm install --production` it behaves like this:
- If there are no dependencies installed at all, then the production ones are installed
- If there are dev dependencies, they are removed.

If that is true, then by just executing `npm install --production` (instead of prune) concern number 2 is gone. Now to solve concern number 1, we could offer some kind of flag, like `--restoreDevDepencencies`, that will execute one extra step after the deployment is finished to run `npm install` and, thus, leave the user environment as it was before the deployment.

What do you think?

NickSeagull:
Regarding the last part @alvaroloes , I think that we can restore the dependencies by default, and add perhaps a `--skipRestoreDependencies`

NickSeagull:
Hello @sanskar-p , any progress on this?

sanskar-p:
Hey @NickSeagull, sorry I had been busy for the past few days. I will make the suggested changes/improvements right away.

sanskar-p:
For some reason, the test says that function is not being called. Not sure if it's actually not being called or if I've written the test wrongly.
Here's the relevant code:
`deploy.ts:`
```
async function pruneDependencies(config: BoosterConfig): Promise<void> {
  try {
    await exec('npm install --production', { cwd: projectDir(config) })
  } catch (e) {
    throw wrapExecError(e, 'Could not prune dev dependencies')
  }
}
```
```
Script.init(`boost ${Brand.dangerize('deploy')} [${environment}] 🚀`, loader)
    //TODO: We should install dependencies in production mode before deploying
    .step('Removing dev dependencies', pruneDependencies)
    .step('Deploying', (config) => deployer(config, logger))
    .step('Reinstalling dependencies', reinstallDependencies)
    .info('Deployment complete!')
    .done()
```

The test:
```
const pruneDependencies = deploy.__get__('pruneDependencies')            //because rewire library is being used

fancy.stdout().it('calls pruneDependencies function', async (ctx) => {
      const fakeProvider = {} as ProviderLibrary

      const fakeLoader = fake.resolves({
        provider: fakeProvider,
        appName: 'fake app',
        region: 'tunte',
        entities: {},
      })

      const fakeDeployer = fake((_config: unknown, logger: Logger) => {
        logger.info('this is a progress update')
      })

      const pruneDependenciesSpy = spy(pruneDependencies)
      await runTasks('test-env', fakeLoader, fakeDeployer)
        
      expect(pruneDependenciesSpy).to.have.been.calledOnce

      expect(ctx.stdout).to.include('Deployment complete')
      expect(fakeDeployer).to.have.been.calledOnce
    })

```

NickSeagull:
The test log says that something is not working properly with your setup spies. [Check it out here](https://github.com/boostercloud/booster/pull/455/checks?check_run_id=1251902808#step:8:1024)

I believe this is because you are getting the `pruneDependencies` function, but you're not setting it?

Try something like

```ts
const pruneDependencies = deploy.__get__('pruneDependencies')            //because rewire library is being used
const pruneDependenciesSpy = spy(pruneDependencies)
deploy.__set__('pruneDependencies', pruneDependenciesSpy)
```

packages/cli/src/commands/deploy.ts
```diff
@@ -5,6 +5,30 @@ import { BoosterConfig, Logger } from '@boostercloud/framework-types'
 import { Script } from '../common/script'
 import Brand from '../common/brand'
 import { logger } from '../services/logger'
+import { exec } from 'child-process-promise'
+import { wrapExecError } from '../common/errors'
+import * as path from 'path'
+
+
+let skipRestoreDependencies = false;
```
NickSeagull, line 13: I believe you can delete this global variable and pass it as a parameter (keep reading)

packages/cli/src/commands/deploy.ts
```diff
@@ -5,6 +5,30 @@ import { BoosterConfig, Logger } from '@boostercloud/framework-types'
...
+
+async function pruneDependencies(config: BoosterConfig): Promise<void> {
+  try {
+    await exec('npm install --production', { cwd: projectDir(config) })
+  } catch (e) {
+    throw wrapExecError(e, 'Could not prune dev dependencies')
+  }
+}
+
+async function reinstallDependencies(config: BoosterConfig): Promise<void> {
```
NickSeagull, line 23: You can pass the parameter here like
```suggestion
async function reinstallDependencies(skipRestoreDependencies: boolean, config: BoosterConfig): Promise<void> {
```

And then, in the command part... (keep reading)

packages/cli/src/commands/deploy.ts
```diff
@@ -16,7 +40,9 @@ const runTasks = async (
 ): Promise<void> =>
   Script.init(`boost ${Brand.dangerize('deploy')} [${environment}] 🚀`, loader)
     //TODO: We should install dependencies in production mode before deploying
+    .step('Removing dev dependencies', pruneDependencies)
     .step('Deploying', (config) => deployer(config, logger))
+    .step('Reinstalling dependencies', reinstallDependencies)
```
NickSeagull, line 45: Here, you can pass `skipRestoreDependencies` as a boolean parameter to the `runTasks` function, and pass it to `reinstallDependencies` like so:
```suggestion
    .step('Reinstalling dependencies', reinstallDependencies.bind(null, skipRestoreDependencies))
``` 

The `bind` method receives a possible `this` as a first parameter (we don't use it so we pass `null`), and then the rest of the parameters in order, for the missing parameters it returns a new function. [Here's a much better explanation by @javiertoledo](https://dev.to/theagilemonkeys/currying-in-typescript-just-for-fun-5h9).

And finally, in the `run` method... (keep reading)

packages/cli/src/commands/deploy.ts
```diff
@@ -37,7 +67,17 @@ export default class Deploy extends Command {
       console.log('Error: no environment name provided. Usage: `boost deploy -e <environment>`.')
       return
     }
+
+    if(flags.skipRestoreDependencies){
+      skipRestoreDependencies = true;
+    }
+
```
NickSeagull, line 74: You can delete this now, because we don't use the global variable
```suggestion
```
And ...

packages/cli/src/commands/deploy.ts
```diff
@@ -37,7 +67,17 @@ export default class Deploy extends Command {
...
     process.env.BOOSTER_ENV = flags.environment
     await runTasks(flags.environment, compileProjectAndLoadConfig(), deployToCloudProvider)
```
NickSeagull, line 76: Here you add the parameter you've added before, passing `flags.skipRestoreDependencies` directly
```suggestion
    await runTasks(flags.skipRestoreDependencies, flags.environment, compileProjectAndLoadConfig(), deployToCloudProvider)
```

NickSeagull (changes requested):
Thanks a lot! This is looking great! I just added a suggestion about the global variable you added, and how to remove it :)

sanskar-p:
@NickSeagull, thanks for your suggestions! When I call the `runTasks` function in the test, I'm setting the value of the `skipRestoreDependencies` as false. Is that fine or would you suggest an alternative for that?

javiertoledo (approved):
Looking good, thanks for all the changes!

NickSeagull (approved):
Everything looks fine, thanks!

NickSeagull:
@sanskar-p can you resolve the conflicts, so we can merge this, please?

sanskar-p:
Done, @NickSeagull

sanskar-p:
Hey guys,
I was trying to fix the unit tests and the below test is failing in `deploy.test.ts`. I can't seem to fix it and I'm not sure why I'm even getting it as I didn't change any of the related files. 
Any advice?

image

sanskar-p:
Hey guys,
I was trying to fix the unit tests and the below test is failing in deploy.test.ts. I can't seem to fix it and I'm not sure why I'm even getting it as I didn't change any of the related files.
Any advice?

image

javiertoledo:
Hey @sanskar-p, thanks for the changes! I'll investigate what's going on with that test, it's weird that it fails because of that!

packages/cli/src/commands/deploy.ts
```diff
@@ -5,18 +5,42 @@ import { BoosterConfig, Logger } from '@boostercloud/framework-types'
...
 const runTasks = async (
+  skipRestoreDependencies: boolean,
   loader: Promise<BoosterConfig>,
   deployer: (config: BoosterConfig, logger: Logger) => Promise<void>
 ): Promise<void> =>
   Script.init(`boost ${Brand.dangerize('deploy')} [${currentEnvironment()}] 🚀`, loader)
     //TODO: We should install dependencies in production mode before deploying
+    .step('Removing dev dependencies', pruneDependencies)
     .step('Deploying', (config) => deployer(config, logger))
+    .step('Reinstalling dependencies', reinstallDependencies.bind(null, skipRestoreDependencies))
```
javiertoledo, line 43: Maybe we could skip this step? We could show success as soon as the deployer ends and let the developer decide if they want to reinstall the dependencies or not, here we're forcing them to wait until the whole process is done...

packages/cli/src/commands/deploy.ts
```diff
@@ -29,12 +53,24 @@ export default class Deploy extends Command {
...
+      char: 's',
+      description: 'skips restoring dependencies after deployment'
+    })
   }
 
   public async run(): Promise<void> {
     const { flags } = this.parse(Deploy)
     if (initializeEnvironment(logger, flags.environment)) {
-      await runTasks(compileProjectAndLoadConfig(), deployToCloudProvider)
+      await runTasks(false, compileProjectAndLoadConfig(), deployToCloudProvider)
```
javiertoledo, line 65: Instead of passing the flag, you could pass here the `pruneDependencies` and `reinstallDependencies` functions already initialized with the flag option, or pass them conditionally only when the flag is true.

packages/cli/src/commands/deploy.ts
```diff
@@ -5,18 +5,42 @@ import { BoosterConfig, Logger } from '@boostercloud/framework-types'
...
+
+async function reinstallDependencies(skipRestoreDependencies: boolean, config: BoosterConfig): Promise<void> {
+  try {
+    if(skipRestoreDependencies === false){
+      await exec('npm install', { cwd: projectDir(config) })
+    }
+  } catch (e) {
+    throw wrapExecError(e, 'Could not reinstall dependencies')
+  }
+}
```
javiertoledo, line 29: I think that it would make sense to move functions like these to a separate `../common/dependencies.ts` file

packages/cli/test/commands/deploy.test.ts
```diff
@@ -27,7 +29,7 @@ describe('deploy', () => {
         const fakeDeployer = fake()
         replace(environment, 'currentEnvironment', fake.returns('test-env'))
```
javiertoledo, line 30: I think that you should mock `pruneDependencies` and `reinstallDependencies` here to make sure they're not actually running the `exec`s

javiertoledo (changes requested):
@sanskar-p I think that some of the conflicts were not properly resolved, I left a couple of comments. I also want to know the opinion of @charlietfe, to make sure we're doing this at the right moment during the deployment process.

packages/cli/src/commands/deploy.ts
```diff
@@ -5,18 +5,42 @@ import { BoosterConfig, Logger } from '@boostercloud/framework-types'
...
+
```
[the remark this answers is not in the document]
charlietfe: Some notes here, if you have previous execution you will have installed dev dependencies in `node_modules` so the `npm install --production` won't delete those dependencies. That's why we need to execute the `prune` instead. But if we do that the deployment won't work because as you said @javiertoledo, we still have a provider dev dependency. Should we add the provider dependencies alongside the `cli` to avoid that?

Another option could be passing a function that will be called just in time after resolving the provider and before deployment, some kind of hook, but I'm still not sure where to execute that.

And the last option and the dirtiest one could use a different directory for dev and prod dependencies, but that involves some changes on the fly I think.

javiertoledo:
@sanskar-p this issue has become a priority because of #367, so I'm closing this branch and move your commits to a branch I can edit to apply the suggestions I made and create a sandboxed integration tests. Thanks a lot for your work on this fix!

Issue #318, opened 2020-08-05
Local Provider: Queries API


LaiaPerez88:
The current local provider implementation is based on ExpressJS and the local document database nedb.

Implement the controllers and the corresponding required methods from the ProviderInterface to support GraphQL queries over Read Models.

**Acceptance criteria:**

- The system can process a GraphQL query and returns the corresponding data from the Read Models.
- Unit tests for the new/changed code.
- Integration and/or end-to-end tests that demonstrate that the query interface returns the data stored in the Read Models.

NickSeagull:
Fixed by #473

PR #473, opened 2020-10-22
Local Provider - Read Models Simple Queries

### LINKED ISSUE #317: Local Provider: Entities projection into read models
**Description**
The current local provider implementation is based on ExpressJS and the local document database nedb.

Implement the corresponding required methods from the ProviderInterface to make sure that new read models are projected and stored after every snapshot is processed in the corresponding tables in the nedb database. Tables should be generated automatically on the server startup.

**Acceptance criteria:**

* All the expected tables for the configured read models must be created when the server is started if they don’t exist.
* Every time that an event is processed, a new snapshot is generated, triggering the projection of the corresponding read models and storing their values in the right tables.
* Unit tests for the new/changed code.
* Integration and/or end-to-end tests that demonstrate that snapshots are projected into read models and stored in the corresponding tables.

NickSeagull:
## Description

This PR allows to use the local provider to do simple queries (that is getting ALL specific read models from the app),
It's not much but it allows at least start trying Booster apps locally

## Changes

* Added read model registry
* Added read model adapter
* Added appropriate tests

## Checks
- [x] Project Builds
~- [ ] Updated documentation accordingly~ Will add it when I finish the feature

Fixes #317


packages/framework-provider-local/src/library/read-model-adapter.ts
```diff
@@ -0,0 +1,63 @@
...
+  UUID,
+} from '@boostercloud/framework-types'
+import { ReadModelRegistry } from '../services/read-model-registry'
+
+export async function rawReadModelEventsToEnvelopes(
+  config: BoosterConfig,
+  logger: Logger,
+  rawEvents: Array<unknown>
+): Promise<Array<ReadModelEnvelope>> {
+  return rawEvents.map((event) => event as ReadModelEnvelope)
```
alvaroloes, line 16: If you are not doing any structure conversion, you can simply do:
```suggestion
  return rawEvents as Array<ReadModelEnvelope>
```

packages/framework-provider-local/src/library/read-model-adapter.ts
```diff
@@ -0,0 +1,63 @@
...
+  db: ReadModelRegistry,
+  _config: BoosterConfig,
+  logger: Logger,
+  readModelName: string,
+  readModelID: UUID
+): Promise<ReadModelInterface> {
+  const response = await db.query({ typeName: readModelName, value: { id: readModelID } })
+  const item = response[0]
+  if (!item) {
+    console.log(`[ReadModelAdapter#fetchReadModel] Failed to fetch read model ${readModelName} with ID ${readModelID}`)
```
alvaroloes, line 29: I would not say `Failed`, as it is not a failure not to find a read model. I would say something like "not found".

packages/framework-provider-local/src/library/read-model-adapter.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export async function searchReadModel(
+  db: ReadModelRegistry,
+  _config: BoosterConfig,
+  logger: Logger,
+  readModelName: string,
+  filters: Record<string, Filter<unknown>>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): Promise<Array<any>> {
```
alvaroloes, line 57:
```suggestion
): Promise<Array<ReadModelInterface>> {
```

packages/framework-provider-local/src/library/read-model-adapter.ts
```diff
@@ -0,0 +1,63 @@
...
+  logger.debug('[ReadModelAdapter#searchReadModel] Arrived these filters: ', filters)
```
alvaroloes, line 58: What do you think about failing if there are filters defined saying "not yet supported"?

Also, could you please create a ticket to implement the filtering?

packages/framework-provider-local/src/services/read-model-registry.ts
```diff
@@ -0,0 +1,40 @@
...
+  public async deleteAll(): Promise<number> {
+    const deletePromise = new Promise((resolve, reject) =>
+      this.readModels.remove({}, { multi: true }, (err, numRemoved: number) => {
+        if (err) reject(err)
+        else resolve(numRemoved)
+      })
+    )
+
+    return (await deletePromise) as number
+  }
```
alvaroloes, line 39: Is `deleteAll` used somewhere? If not, I would not include it here.

Also, it would be great to have a `deleteById` function, as it is now possible delete ReadModels ;-)

alvaroloes (changes requested):
Left some comments. Also: _Could you please add integration tests?_

NickSeagull:
Thanks for reminding me about integration tests in the local provider, completely forgot!

packages/framework-provider-local/src/library/read-model-adapter.ts
NickSeagull: Done here https://github.com/boostercloud/booster/pull/473/commits/20070ab9cbf9f4f22d6a72af2357c85bdc86fcd8

packages/framework-provider-local/src/library/read-model-adapter.ts
NickSeagull: Done here https://github.com/boostercloud/booster/pull/473/commits/ca68b39d2f681f3d2d67629f2a1fc95075b32f8c

packages/framework-provider-local/src/library/read-model-adapter.ts
NickSeagull: This will fail with the following error message:
```text
@boostercloud/framework-provider-local: src/index.ts(38,5): error TS2322: Type '(_config: BoosterConfig, logger: Logger, readModelName: string, filters: Record<string, Filter<unknown>>) => Promise<ReadModelInterface[]>' is not assignable to type '<TReadModel extends ReadModelInterface>(config: BoosterConfig, logger: Logger, entityTypeName: string, filters: Record<string, Filter<unknown>>) => Promise<...>'.
@boostercloud/framework-provider-local:   Type 'Promise<ReadModelInterface[]>' is not assignable to type 'Promise<TReadModel[]>'.
@boostercloud/framework-provider-local:     Type 'ReadModelInterface[]' is not assignable to type 'TReadModel[]'.
@boostercloud/framework-provider-local:       Type 'ReadModelInterface' is not assignable to type 'TReadModel'.
@boostercloud/framework-provider-local:         'ReadModelInterface' is assignable to the constraint of type 'TReadModel', but 'TReadModel' could be instantiated with a different subtype of constraint 'ReadModelInterface'.
```

I spent like 1 hour trying to figure it out, but ended up leaving an `any`

adriangmweb (approved):
Looking good! And this is really necessary to make people comfortable when starting using Booster 💯!

adriangmweb:
About the integration tests, if they're still failing maybe wouldn't be worth including them in another PR once it's fixed instead of delaying this one?

NickSeagull:
I can write some integration tests for this one, and test them individually @adriangmweb , they shouldn't be dependant on the others

packages/framework-provider-local/src/library/read-model-adapter.ts
alvaroloes: WTF! Ok then

packages/framework-provider-local/src/library/read-model-adapter.ts
NickSeagull: Actually there's already a ticket for this https://github.com/boostercloud/booster/issues/318

NickSeagull: Fixed in https://github.com/boostercloud/booster/pull/473/commits/f9fb9521fa7b31273c089704b9fea1e7e10347b7

packages/framework-provider-local/src/services/read-model-registry.ts
NickSeagull: Fixed in https://github.com/boostercloud/booster/pull/473/commits/ee1feeb58df2a2cee791965079c7c0e786b2efc5

packages/framework-integration-tests/integration/providers/local/setup.ts
```diff
@@ -1,14 +1,9 @@
 import { start } from '../../../src/deploy'
-import { ChildProcess, execSync } from 'child_process'
+import { ChildProcess } from 'child_process'
```
javiertoledo, line 2: I discovered that we have the package [`child-process-promise`](https://www.npmjs.com/package/child-process-promise) among our dependencies, which are wrappers around `child_process` functions without explicitly using promisify in our code.

packages/framework-provider-local/src/services/read-model-registry.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  public async query(query: object): Promise<Array<ReadModelEnvelope>> {
+    const queryPromise = new Promise((resolve, reject) =>
+      this.readModels.find(query).exec((err, docs) => {
+        if (err) reject(err)
+        else resolve(docs)
+      })
+    )
+
+    return (await queryPromise) as Array<ReadModelEnvelope>
```
javiertoledo, line 19: I think that the await and the async modifiers are optional in this case. I wonder if this could work:

```suggestion
    return queryPromise as Promise<Array<ReadModelEnvelope>>
```

packages/framework-provider-local/src/services/read-model-registry.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  public async deleteById(id: UUID): Promise<number> {
+    const deletePromise = new Promise((resolve, reject) =>
+      this.readModels.remove({ value: { id } }, { multi: true }, (err, numRemoved: number) => {
+        if (err) reject(err)
+        else resolve(numRemoved)
+      })
+    )
+
+    return (await deletePromise) as number
```
javiertoledo, line 38: Same here, I think you don't need the async/await

packages/framework-provider-local/test/services/read-model-registry.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+describe('the read model registry', () => {
+  let initialReadModelsCount: number
+  let mockReadModel: ReadModelEnvelope
+
+  let readModelRegistry: ReadModelRegistry
+
+  beforeEach(async () => {
+    initialReadModelsCount = random.number({ min: 2, max: 10 })
+    readModelRegistry = new ReadModelRegistry()
```
javiertoledo, line 18: If you're using the actual database, this could be considered an "integration test" and the real I/O will likely make it slower. Would it make sense to mock the DB or use the in-memory version of nedb?

javiertoledo (changes requested):
Looking good, would it make sense to transform the tests where you're using the database into integration tests or mock the database to make them work faster as unit tests?

packages/framework-provider-local/src/library/read-model-adapter.ts
```diff
@@ -0,0 +1,65 @@
...
+} from '@boostercloud/framework-types'
+import { ReadModelRegistry } from '../services/read-model-registry'
+
+export async function rawReadModelEventsToEnvelopes(
+  config: BoosterConfig,
+  logger: Logger,
+  rawEvents: Array<unknown>
+): Promise<Array<ReadModelEnvelope>> {
+  return rawEvents as Array<ReadModelEnvelope>
+}
```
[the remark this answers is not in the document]
NickSeagull: This is useful for all the providers (AWS/Azure/K8s...) as they do not provide you the structure that Booster expects, this is a special case for the local provider, but the rest of the providers require the transformation function.

packages/framework-provider-local/src/library/read-model-adapter.ts
```diff
@@ -0,0 +1,65 @@
...
+  readModelName: string,
+  filters: Record<string, Filter<unknown>>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): Promise<Array<any>> {
+  if (Object.keys(filters).length != 0) {
+    throw Error('Filters not supported yet by local provider, but got:' + JSON.stringify(filters))
+  }
+  const query = { typeName: readModelName }
+  const result = await db.query(query)
+  logger.debug('[ReadModelAdapter#searchReadModel] Search result: ', result)
```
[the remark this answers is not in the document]
NickSeagull: Yes, but at the moment we're not using it

packages/framework-integration-tests/integration/providers/local/setup.ts
NickSeagull: Not really sure what you mean here, do you want me to avoid using the `execSync` function?

packages/framework-provider-local/test/services/read-model-registry.test.ts
NickSeagull: I'll approach this on a new PR

packages/framework-provider-local/src/services/read-model-registry.ts
NickSeagull: Done in https://github.com/boostercloud/booster/pull/473/commits/3c8b3a4c2f22227b10e16c3689c8c1f4c03e051d

packages/framework-provider-local/src/services/read-model-registry.ts
NickSeagull: Done in https://github.com/boostercloud/booster/pull/473/commits/3c8b3a4c2f22227b10e16c3689c8c1f4c03e051d

NickSeagull:
Most of the changes are made, can I get an approval so we can work in another branch with the next steps? 🥺

adriangmweb (approved):
LGTM! 🚀

javiertoledo (approved):

packages/framework-integration-tests/integration/providers/local/setup.ts
javiertoledo: No, just mentioning that we have this other way to import `exec` that is "promisified" by default:

```typescript
import { exec } from 'child-process-promise'

...
await exec('ls -la') // You have the option to wait for the response using await
```

alvaroloes (approved):

PR #477, opened 2020-10-27
Feature: Use JWT authorization tokens to grant access

charlietfe:
## Description
In this PR we have added minimal changes to allow the usage of external JWT tokens, in this way, the user could use its external auth provider or it could use their own custom solution. 

## TL;DR

In order to use the generic token decoder/verifier, the tokens must be signed using the RSA256 algorithm (all providers allow this method) and use an OpenID Connect server to verify those tokens.

The tokens contain a set of standard claims (fields) based on this [spec](https://openid.net/specs/openid-connect-core-1_0.html#Claims), so we might use some of them to fill or recreate our `UserEnvelopes`.

An OpenID Connect server can be any publicly accessible endpoint, like an S3 bucket if you want, but most providers offer one, so don't worry. In that server, we should have a public file following [this spec](https://openid.net/specs/openid-connect-discovery-1_0.html). This file contains the **PUBLIC_KEY** to verify the tokens, so from Booster perspective we only need two fields: `issuer` and `jwks_uri` as a part of the configuration.

Regarding custom claims, we still expect the user roles in a `custom:roles` field so the token must include it. All providers include a way of customizing those fields.

Here are some nice articles about this topic (all of them are based on API Gateway validators but we are doing the same under the hood):

https://www.jeremydaly.com/verifying-self-signed-jwt-tokens-with-aws-http-apis/
https://dev.to/martzcodes/token-authorizers-with-apigatewayv2-tricks-apigwv1-doesn-t-want-you-to-know-41jn

## Changes

We've moved the verification functionality to the core in `BoosterAuth` and we will use that directly from the `booster-graphql-dispatcher.ts`, in that way this implementation, is totally independent of the providers. From providers, we'll only need the Authorization Headers and that's it. 

For that reason we've changed the `GraphQLRequestEnvelope` to include the token instead of a user got it from Cognito, now for us, a token is enough.

If the envelope contains the token from the request we'll proceed to verify it against the provider, for that we've added a tokenVerifier config property with the issuer and jwks_uri fields.

We still support the auth api in Booster, **but we are verifying the Cognito jwt tokens in a standard way**, to achieve that, when we create the infrastructure we are setting two environment variables for the Cognito issuer and jwks_uri to be available for the lambdas, that means that always we'll have jwtVerifier object set to use the standard verification. 

## Concerns

We should cache in the future the verification file anyway, the JwksClient has some cache capabilities but in a serverless environment, the memory cache will last while the lambda is live. But this is for the next round.

## Checks
- [X] Project Builds
- [x] Unit tests
- [x] Integration tests


packages/framework-types/src/config.ts
```diff
@@ -45,6 +45,8 @@ export class BoosterConfig {
   /** Environment variables set at deployment time on the target lambda functions */
   public readonly env: Record<string, string> = {}
 
+  private _tokenVerifier?: { issuer: string; jwksUri: string }
```
charlietfe, line 48: By now this is not mandatory

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/api-stack-velocity-templates.ts
```diff
@@ -45,6 +45,7 @@ export const CognitoTemplates = {
     response: `#set($root = $input.path('$'))
                {
                    "accessToken": "$root.AuthenticationResult.AccessToken",
+                   "idToken": "$root.AuthenticationResult.IdToken",
```
charlietfe, line 48: The idToken from Cognito allows us to add custom claims

packages/framework-provider-aws/src/library/graphql-adapter.ts
```diff
@@ -18,11 +19,20 @@ export async function rawGraphQLRequestToEnvelope(
       graphQLValue = JSON.parse(request.body)
     }
 
+    let currentUser = undefined
```
charlietfe, line 22: Ensuring backward compatibility here. Some guys are still using our internal Cognito implementation.

packages/framework-core/src/booster-auth.ts
```diff
@@ -62,6 +65,66 @@ export class BoosterAuth {
...
+    return new Promise((resolve, reject) => {
+      const getKey = (header: jwt.JwtHeader, callback: jwt.SigningKeyCallback): void => {
+        if (!header.kid) {
+          callback(new Error('JWT kid not found.'))
+          return
+        }
+
+        client.getSigningKey(header.kid, function(err: any, key: jwksRSA.SigningKey) {
+          if (err) {
+            callback(err, '')
```
charlietfe, line 97: This callback doesn't accept null so an empty string will be enough here.

packages/framework-core/src/booster-auth.ts
```diff
@@ -62,6 +65,66 @@ export class BoosterAuth {
...
+          }
+          const signingKey = key.getPublicKey()
+          callback(null, signingKey)
+        })
+      }
+
+      jwt.verify(token, getKey, verifyOptions, (err: any, decoded: any) => {
+        if (err) {
+          reject(err)
+          return
```
adriangmweb, line 108: Isn't this return unnecessary? The promise is already being rejected

charlietfe: Agreed!

adriangmweb (commented):
This is awesome! Enabling users to use their preferred JWT Auth provider to support new features is just great 🚀

packages/framework-core/src/services/graphql/websocket-protocol/graphql-websocket-protocol.ts
```diff
@@ -78,7 +77,8 @@ export class GraphQLWebsocketHandler {
   private async handleInit(connectionID: string, clientMessage: GraphQLInit): Promise<void> {
     let userEnvelope: UserEnvelope | undefined
     if (clientMessage.payload?.Authorization) {
-      userEnvelope = await this.authManager.fromAuthToken(clientMessage.payload.Authorization)
+      const token = clientMessage.payload.Authorization.replace('Bearer ', '')
```
charlietfe, line 80: Maybe we should move the sanitizer inside BoosterAuth.verifyToken.

juanjoman (approved):
Amazing work! 🤯  Just left some minor suggestions but LGTM!!

packages/framework-provider-aws/src/library/graphql-adapter.ts
charlietfe: Fixed!

packages/framework-core/src/booster-auth.ts
charlietfe: Fixed.

packages/framework-core/src/services/graphql/websocket-protocol/graphql-websocket-protocol.ts
charlietfe: Fixed

packages/framework-core/test/booster-auth.test.ts
```diff
@@ -250,3 +253,63 @@ describe('the "isUserAuthorized" method', () => {
...
+    const token = jwks.token({
+      sub: userId,
+      iss: issuer,
+      'custom:role': 'User',
+      email: email,
+      phone_number: phoneNumber,
+      exp: 0,
+    })
+
+    await expect(BoosterAuth.verifyToken(config, token)).to.eventually.to.be.rejected
```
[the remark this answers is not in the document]
charlietfe: Fixed.

packages/framework-core/test/services/graphql/websocket-protocol/graphql-websocket-protocol.test.ts
```diff
@@ -8,19 +8,19 @@ import {
...
 import { GraphQLWebsocketHandler } from '../../../../src/services/graphql/websocket-protocol/graphql-websocket-protocol'
 import { ExecutionResult } from 'graphql'
 import { expect } from '../../../expect'
+import { BoosterAuth } from '../../../../src/booster-auth'
 
 describe('the `GraphQLWebsocketHandler`', () => {
   let config: BoosterConfig
   let websocketHandler: GraphQLWebsocketHandler
-  let authManager: ProviderAuthLibrary
+  // let authManager: ProviderAuthLibrary
```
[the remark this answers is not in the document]
charlietfe: Fixed

packages/framework-core/src/booster-auth.ts
NickSeagull: Can you add this as a comment?
```suggestion
            // This callback doesn't accept null so an empty string is enough here
            callback(err, '')
```

packages/framework-core/src/booster-auth.ts
```diff
@@ -62,6 +65,65 @@ export class BoosterAuth {
...
+          resolve(BoosterAuth.tokenToUserEnvelope(decoded))
+        } catch (err) {
+          reject(err)
+        }
+      })
+    })
+  }
+
+  private static tokenToUserEnvelope(decodedToken: any): UserEnvelope {
+    const username = decodedToken.email ? decodedToken.email : decodedToken.phone_number
```
NickSeagull, line 116: Simplification
```suggestion
    const username = decodedToken.email ?? decodedToken.phone_number
```

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -59,12 +54,33 @@ export class BoosterGraphQLDispatcher {
     }
   }
 
+  private async verifyTokenFromEnvelop(
+    envelope: GraphQLRequestEnvelope
+  ): Promise<GraphQLRequestEnvelope | GraphQLRequestEnvelopeError> {
+    if ('token' in envelope && envelope.token) {
```
NickSeagull, line 60: Why not
```suggestion
    if (envelope.token) {
```

packages/framework-core/test/booster-auth.test.ts
```diff
@@ -250,3 +253,63 @@ describe('the "isUserAuthorized" method', () => {
...
+
+    expect(await BoosterAuth.verifyToken(config, token)).to.deep.equals(expectedUser)
+  })
+
+  it('fails if a different issuer emitted the token', async () => {
+    const token = jwks.token({
+      iss: 'firebase',
+    })
+
+    await expect(BoosterAuth.verifyToken(config, token)).to.eventually.to.be.rejected
```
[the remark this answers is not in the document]
charlietfe: Fixed

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/auth-stack.ts
```diff
@@ -20,12 +15,14 @@ export class AuthStack {
...
 
-  public build(): void {
+  public build(): UserPool | undefined {
     if (this.config.thereAreRoles) {
       const userPool = this.buildUserPool()
       this.buildUserPoolClient(userPool)
       this.buildAuthAPI(userPool)
+      return userPool
     }
+    return undefined
```
NickSeagull, line 25: We can avoid returning `undefined` explicitly, as it's what it returns implicitly if I recall correctly
```suggestion
```

packages/framework-provider-aws/src/library/graphql-adapter.ts
```diff
@@ -22,7 +19,7 @@ export async function rawGraphQLRequestToEnvelope(
       requestID,
       eventType,
       connectionID,
-      currentUser: await userEnvelopeFromAuthToken(userPool, request.headers?.Authorization),
+      token: request.headers?.Authorization,
```
NickSeagull, line 22: What if `Authorization` is undefined?

NickSeagull (changes requested):
Thanks a lot! This is great :D Left some comments

packages/framework-provider-aws/src/library/graphql-adapter.ts
charlietfe: @NickSeagull we might have open read models or something like that, so in that case the Authorization is null, then in the graphql message handler we won't check the token.

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/auth-stack.ts
charlietfe: Since we have an if statement that returns is required, all routes should return a value.

packages/framework-core/src/booster-graphql-dispatcher.ts
charlietfe: This was to distinguish between types, but I think isn't needed

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/auth-stack.ts
NickSeagull: Ah I see :)

packages/framework-core/src/booster-auth.ts
charlietfe: Fixed!

packages/framework-core/src/booster-auth.ts
charlietfe: Done

packages/framework-core/src/booster-graphql-dispatcher.ts
charlietfe: Done.

NickSeagull (approved):

javiertoledo (approved):
Beautiful and useful change! I miss many things, but I guess we can work on them in upcoming PRs (I'm all in to pick some of the work!), we must make sure that we create issues and work on all of the following:

- [ ] Integration tests: Send an actual request with a JWT token and check that it works
- [ ] Documentation: How to use this feature, maybe write a couple of sections/articles about how to use it with Cognito and Auth0. 
- [ ] Design how we're going to manage roles in upcoming releases (I saw you already started that discussion with Alvaro)
- [ ] Remove (or move to a plugin) the current auth endpoints (Users that use Auth0 won't need them)
- [ ] It would be nice to support this feature at least in the local provider too

packages/framework-core/src/booster-auth.ts
```diff
@@ -62,6 +65,65 @@ export class BoosterAuth {
...
+
+  private static tokenToUserEnvelope(decodedToken: any): UserEnvelope {
+    const { phone_number, email, 'custom:role': role } = decodedToken
+    const username = email ? email : phone_number
+
+    return {
+      username,
+      role: role?.trim() ?? '',
+    }
+  }
```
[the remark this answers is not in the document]
charlietfe: @javiertoledo I like that idea and I think it makes more sense than having a UserEnvelope with only a username and a role. Let's add it to the list of next tasks.

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -61,6 +62,12 @@ export class BoosterGraphQLDispatcher {
 
   private async handleMessage(envelope: GraphQLRequestEnvelope | GraphQLRequestEnvelopeError): Promise<DispatchResult> {
     this.logger.debug('Starting GraphQL operation')
+
+    if ('token' in envelope && envelope.token) {
+      this.logger.debug('Decoding current user from auth token')
+      envelope.currentUser = await BoosterAuth.verifyToken(this.config, envelope.token)
```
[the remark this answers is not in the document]
charlietfe: If the verifier fails, we are creating a `GraphQLRequestEnvelopeError` with an `InvalidParameterError` error inside. This error will be resolved by the specific handlers.

```
private async handleMessage(envelope: GraphQLRequestEnvelope | GraphQLRequestEnvelopeError): Promise<DispatchResult> {
    this.logger.debug(`Starting GraphQL operation: ${JSON.stringify(envelope)}`)

    const envelopeOrError = await this.verifyTokenFromEnvelop(envelope) // <---- If it fails a GraphQLRequestEnvelopeError will be returned.

    if (cameThroughSocket(envelopeOrError)) {
      return this.websocketHandler.handle(envelopeOrError) // <---- Here we will have an envelop with the error
    }
    return this.runGraphQLOperation(envelopeOrError) // <---- Same here
  }
```
Makes sense, right? @javiertoledo

packages/framework-core/src/booster-auth.ts
```diff
@@ -62,6 +65,65 @@ export class BoosterAuth {
     // Check the roles then
     return userHasSomeRole(user, authorizedRoles)
   }
+
+  public static async verifyToken(config: BoosterConfig, token?: string): Promise<UserEnvelope> {
+    if (!config.tokenVerifier) {
+      return Promise.reject('Token verifier not configured')
```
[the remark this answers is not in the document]
charlietfe: Fixed in 705fb1ecea6fb91491b5d32bc2e4284b3da71e81

packages/framework-core/src/booster-auth.ts
```diff
@@ -62,6 +65,65 @@ export class BoosterAuth {
...
+          }
+          const signingKey = key.getPublicKey()
+          callback(null, signingKey)
+        })
+      }
+
+      jwt.verify(token, getKey, verifyOptions, (err: any, decoded: any) => {
+        if (err) {
+          reject(err)
+        }
```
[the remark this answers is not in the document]
charlietfe: Fixed in 705fb1ecea6fb91491b5d32bc2e4284b3da71e81

packages/framework-core/src/booster-auth.ts
```diff
@@ -62,6 +65,65 @@ export class BoosterAuth {
...
+        try {
+          resolve(BoosterAuth.tokenToUserEnvelope(decoded))
+        } catch (err) {
+          reject(err)
+        }
```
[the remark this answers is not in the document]
charlietfe: Fixed in 705fb1ecea6fb91491b5d32bc2e4284b3da71e81

packages/framework-core/src/booster-graphql-dispatcher.ts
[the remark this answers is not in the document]
javiertoledo: So if I understood it, there are three possible scenarios:

* A request with no token is assumed to be an anonymous request and can call public commands and read models, but will fail with restricted commands or read models.
* A request with an invalid token will fail no matter if the command they're calling is public or not
* A request with a valid token will succeed for public and authorized commands and read models

Is that right? For me, this is fine for now.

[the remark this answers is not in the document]
charlietfe: Exactly, those are the scenarios we are covering so this is fine.

external-content duckduckgo com

packages/framework-provider-aws-infrastructure/src/infrastructure/params.ts
```diff
@@ -26,6 +28,8 @@ export function lambda(
       ...config.env, // Adds custom environment variables set by the user in the config file
       [environmentVarNames.restAPIURL]: baseURLForAPI(config, stack, apis.restAPI.restApiId),
       [environmentVarNames.websocketAPIURL]: baseURLForAPI(config, stack, apis.websocketAPI.ref),
+      [environmentVarNames.jwtIssuer]: userPool ? issuer(stack, userPool?.userPoolId) : '',
+      [environmentVarNames.jwksUri]: userPool ? jwksUri(stack, userPool?.userPoolId) : '',
```
alvaroloes, line 32: You don't need (as fas as I know) the question mark in these expressions as you already checked it is not null, so you can just do `userPool.userPoolId`

packages/framework-provider-aws/src/constants.ts
```diff
@@ -22,4 +22,6 @@ export const connectionsStoreAttributes = {
 export const environmentVarNames = {
   restAPIURL: 'BOOSTER_REST_API_URL',
   websocketAPIURL: 'BOOSTER_WEBSOCKET_API_URL',
+  jwtIssuer: JWT_ENV_VARS.BOOSTER_JWT_ISSUER,
+  jwksUri: JWT_ENV_VARS.BOOSTER_JWKS_URI,
```
alvaroloes, line 26: I'd say you don't need `jwtIssuer` nor `jwkUri` constants, as you already have those constant in the generic types package (`JWT_ENV_VARS.*`)

You can import `'@boostercloud/framework-types'` from "aws" and "aws-infrastructure" perfectly 😃

alvaroloes (approved):
Just two minor comments

packages/framework-core/test/booster-auth.test.ts
```diff
@@ -250,3 +253,63 @@ describe('the "isUserAuthorized" method', () => {
...
+
+  afterEach(async () => {
+    await jwks.stop()
+  })
+
+  it('decode and verify an auth token with the custom roles', async () => {
+    const token = jwks.token({
+      sub: userId,
+      iss: issuer,
+      'custom:role': 'User',
```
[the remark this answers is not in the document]
alvaroloes: Exactly. @charlietfe and I give this exact same concept a twist and thought about using a function in the `auth` field, so that the user can do whatever they want, not just comparing claims as strings.
Among other reasons, it is perfectly possible (Hasura does this) that one of the claims is an array of possible roles, so to check the authorization we should see if the user role is inside that array of roles.

Because of these corner cases, using a function is much more flexible. For example:
```typescript
@Command({
  auth: (userEnvelope: UserEnvelope): boolean => { // The UserEnvelop has the claims extracted from the token
    return ['writeAll', 'writePayments'].contains(userEnvelope.claims['permission']) // For example
  }
})
export class ConfirmPayment {
  ...
}
```

This way, Booster now forgets completely about the authorization part too (including roles!) making this ultra-decoupled.

However, this could be clunky for people (a lambda there and repeated multiple times). So the idea is to provide the same behavior we have now with Booster provided functions. This is an example of your example but with a function that just does a straight comparison:

```typescript
@Command({
  auth: HasClaims({
    role: 'User', // The token includes a 'custom:role': 'User' claim
    permission: 'FileAccess', // The token includes a 'custom:permission': 'FileAccess' claim
    origin: 'CartService', // The token includes a 'custom:origin': 'CartService' claim
  })
})
export class ConfirmPayment {
  ...
}
```
Finally, as we are going to move the current authentication feature to a rocket, _it is that rocket the one that could use a role-based authentication_. For example, imagine we are using the canonical `BoosterAuth` rocket, which contains exactly the same we have right now in the core. The example would then look like:
```typescript
@Command({
  auth: BoosterAuth.hasRole('Admin')
})
export class ConfirmPayment {
  ...
}
```

💥  So we get a lot of flexibility maintaining practically the same user experience!!

@javiertoledo @charlietfe Let me know what you think

packages/framework-core/src/booster-auth.ts
```diff
@@ -62,6 +65,62 @@ export class BoosterAuth {
...
+    })
+  }
+
+  private static tokenToUserEnvelope(decodedToken: any): UserEnvelope {
+    const username = decodedToken.email ?? decodedToken.phone_number
+    const role = decodedToken['custom:role']
+    return {
+      username,
+      role: role?.trim() ?? '',
+    }
```
alvaroloes, line 118: @charlietfe I forgot! Could you please add here (and to the UserEnvelope type) a field called `id` that contains the id of the user? (it is the `sub` claim)

We have received several requests for having that

packages/framework-provider-aws/src/constants.ts
charlietfe: Umm, I'm doing that at the beginning of the file and the `JWT_ENV_VARS` are defined at `framework-types` level like that:

```
export const JWT_ENV_VARS = {
  BOOSTER_JWT_ISSUER: 'BOOSTER_JWT_ISSUER',
  BOOSTER_JWKS_URI: 'BOOSTER_JWKS_URI',
}
```

What I'm missing? @alvaroloes

packages/framework-core/src/booster-auth.ts
charlietfe: Sure thing!

packages/framework-provider-aws-infrastructure/src/infrastructure/params.ts
charlietfe: Yup! my fault!

packages/framework-provider-aws/src/constants.ts
charlietfe: Ohh I see, forget my last comment

packages/framework-core/src/booster-auth.ts
alvaroloes: Thanks!

packages/framework-types/src/envelope.ts
```diff
@@ -77,6 +78,7 @@ export interface ConnectionDataEnvelope {
 }
 
 export interface UserEnvelope {
+  id?: string
```
alvaroloes, line 81: Why optional? I think the `sub` claim is always in the token.

Asking this because it would make the user code a bit messier, as they would need to _always_ check if the id is not null, while it will never be (if the UserEnvelope is not null, then the id will always be there)

packages/framework-core/test/booster-auth.test.ts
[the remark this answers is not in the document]
javiertoledo: Flexibility is always great, but a flexible interface comes with extra decisions for the final user and less-readable DSLs: you have to understand that you need to put a function there, what it should do, and what's the catalog of available "shortcut" functions (Like `HasRoles`), while a plain type could be autocompleted by the IDE. While this looks like a minor detail, it can impact the way people learn to use the framework, so I'd give extra thought to this DSL. One of Booster's principles is to minimize the amount of "code" and config required for common tasks. I love the idea of having a "fallback function" that is run when no authorization parameters are passed, but putting that function in a decorator parameter sounds weird; it's not usually where people expect to put code... I'd propose supporting the following scenarios:

1. Lower-level claim-based authorization: Requires all the described claims to be present and match. In this case, claims semantics are defined in the Auth provider.
```typescript
@Command({
  requireClaims: {
    role: 'User',
    permission: 'FileAccess',
    origin: 'CartService',
  }
})
export class ConfirmPayment {
  ...
}
```

2. Implement an `authorize` method in the command class where the action can be canceled by throwing an exception, or the token can be extended before injecting it in the command:
```typescript
@Comand
export class ConfirmPayment {
  public static function authorize(jwtToken: JWTToken, command: ConfirmPayment): JWTToken {
    // Maybe we expect a special claim in this command that includes a message ciphered with a secret from the payment provider
    verifyAuthenticityOfPaymentProvider(jwtToken.claims['evil-corp:creepy-secret'])
    return jwtToken
  }
  
  public static function handle(...) {}
}
```

3. It's important to mention that a command where neither one of these DSLs or the `authorize` method is defined should be inaccessible. Booster should even avoid creating the entries in the GraphQL schema. So it'd be also convenient to provide a special shortcut to make them publicly accessible:
```typescript
@Command({
  public: true
})
export class CreateUser {
  public static function handle(...) {}
}
```

With these features we should cover most scenarios. In the future, we can provide extra shortcuts for common cases like authorization via roles or via permissions:

```typescript
@Command({
  requireRole: 'Admin',
})
export class CreateUser {
  public static function handle(...) {}
}
```

packages/framework-types/src/envelope.ts
charlietfe: Mainly for one reason @alvaroloes, we still have some code specially in the `auth-adapter.ts` which is still using the `CognitoUserPoolEvent` to re-create the UserEnvelope. In those cases I don't know if that object contains the userId. In next iterations I will fix that.

packages/framework-core/test/booster-auth.test.ts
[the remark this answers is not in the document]
charlietfe: @javiertoledo  @alvaroloes After giving a thought, I'm more convinced that we should avoid functions as a part of the DSL definition, why? Because it will let the developers to create crazy functions involving even async operations to check i.e a role with a 3rd party provider or something like that, and for me it doesn't feel natural. 

Is better to have a well defined DSL which internally will call the default `authorize` method checking the required roles or on the contrary overwrite `authorize` method and provide your own logic if you want.

We are still covering all the initial use cases we had in mind with a lot of flexibility, also if you want to still create helper functions inside the framework to call them inside `authorize` you can do it as well, but without poluting the DSL with code. Anyway the problem with that approach is how the user will find those helper functions, and even more we should add more documentation to specify how to use them.

PR #476, opened 2020-10-25
Harder, Better, Faster, Stronger CLI integration tests

javiertoledo:
## Description
We have been experiencing reliability issues with CLI integration tests. They were directly changing the files from the sample project in the package `framework-integration-tests`, introducing issues that then affected deployment tests and others, as they all rely on these files to be unaltered.

_You may want to [listen to the OST of this PR](https://www.youtube.com/watch?v=gW5vEUWT5Lk) while reviewing it 😎_

## Changes
In this PR, I've reworked the CLI integration tests with the hopes of solving the mentioned reliability issues and make the tests faster the process. To do that I've done the following:

* Removed file management from `setup.ts` (Tests should be self-contained)
* Each test file creates a sandbox folder where they can mess up the file system in any way they want without altering future tests. For commands that require an existing project, there's a new `createSandboxProject` function that creates a new copy of the sample project in the integration tests package.
* Replaced `exec` calls with the promise-wrapped version from package `child-process-promise` for consistency and code simplicity.
* Created a `loadFixture` helper and moved calls to low-level `fs` functions to `fileHelper.ts` to make tests code more readable
* Replaced exact string comparations to matchers in many expect blocks to make them a bit more resilient to changes.
* prompt interaction in project generator tests now waits for the actual prompt instead of writing the next thing every second.
* Lots of changes to make the coder easier to understand by humans (esp. regexes)
* Bonus: Added a VSCode launch descriptor to ease debugging CLI integration tests.
* Bonus: Fixed the CLI config to keep a consistent context for `skipGit` and `skipInstall` flags (functionally there's no difference)

## Checks
- [x] Project Builds
- [x] Project passes tests and checks
~~-Updated documentation accordingly~~ (not needed)

## Additional notes

This change doesn't guarantee that other integration tests are reliable, I'll tackle that in future PRs.

PR #478, opened 2020-10-27
Simpler checkers

javiertoledo:
## Description
Check Runs have been working unreliably for a while now, and they were becoming harder and harder to manage, so I've reworked the corresponding Github actions to use a simpler/standard workflow. The intention of this PR is to show how this implementation works and reach an agreement with the team about the proposed workflow with the "lock" option. I locked and unlocked the PR to trigger the integration actions. Some of the tests will likely fail because this branch doesn't include the fixes from #476 

## Changes
* Linter and Tests are run on every push, nothing changes here from a functional perspective.
* Integration tests will stop responding to special comments like `integration: all`. Instead of that, the integration tests are run when the PR is locked.

## Why I think that this change makes sense

Locking a PR disables new comments (still allow owner comments) and means that the conversation is over. When you lock a PR you can set a reason, and this reason can be "Resolved", so this makes sense from the workflow POV. We can interpret that locking a PR as "Resolved" means that is time for integration and merge. We can always "unlock" the PR to fix issues if the tests fail and lock it again when we're done to run them again. I think that this could be an elegant solution to automate the integration tests in an easier way, but I'm totally open to alternate ideas. It is very convenient that the `locked` event is a PR event because that allows us to remove the Github check runs management from the jobs, where I suspect we introduced a bug that was making these tests misbehave.

## Checks
- [x] Project Builds

## Additional info

You can find the "Lock conversation" link in the bottom of the right side panel:
Screen Shot 2020-10-27 at 20 11 17

That link opens a modal where you can set the reason:
Screen Shot 2020-10-27 at 20 11 44

Everything is looking good so far. If you remember how this was working in previous commits, clicking on an action details sent you to the wrong commit actions and other weird stuff. All of these issues seem to be resolved here:
Screen Shot 2020-10-27 at 20 56 42


.github/workflows/sync.yml
```diff
@@ -1,19 +0,0 @@
...
-      - name: repo-sync
-        uses: wei/git-sync@v1
-        env:
-          SOURCE_REPO: 'git@github.com:boostercloud/booster.git'
-          SOURCE_BRANCH: 'master'
-          DESTINATION_REPO: ${{ secrets.SYNC_REPO }}
-          DESTINATION_BRANCH: 'master-mirror'
-          SSH_PRIVATE_KEY: ${{ secrets.BOT_SSH_PRIVATE_KEY }}
-        with:
-          args: $SOURCE_REPO $SOURCE_BRANCH $DESTINATION_REPO $DESTINATION_BRANCH
```
javiertoledo, line 19: I deleted this script because we won't need it anymore

.github/workflows/integration-tests.yml
```diff
@@ -1,518 +1,94 @@
...
-          route: PATCH /repos/:repository/check-runs/:check_run_id
-          repository: ${{ github.repository }}
-          mediaType: '{"previews": ["antiope"]}'
-          check_run_id: ${{ fromJson(steps.create_local_integration_check_step.outputs.data).id }}
-          conclusion: ${{ steps.integration_tests_run.outcome }}
-          output: '{"title":"Local Environment Integration Tests", "summary":"Integration tests finished. Here you can take a look at the logs", "text":"```console\n${{ steps.integration_tests_run.outputs.data }}\n``"}'
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - run: npx lerna bootstrap
+      - run: npx lerna run integration/local --stream
```
javiertoledo, line 94: Just like that...

NickSeagull (approved):
Ok, this simplifies a lot, but also it makes me worry about the fact that it locks conversation without being able to add more comments. Although I see that this would be the case for when we are about to merge it.

We should unlock the PR once the tests finish, so people in the future can comment as well, even if it is closed/merged. Maybe this can be automated?

It might be a good idea to add an automated comment that says something like

 "This PR was locked to run the integration tests, and will be unlocked afterwards"

alvaroloes:
Ideally, we should force integration tests to run before you are able to merge. This is not true right now (we didn't find a way) but just wanted to say the ideal scenario.

For me, it seems weird to lock a PR to execute integration tests, it is not obvious. I agree it can be learned if it simplifies things.
Questions and other points:
- When I lock the PR, are the checks shown in the PR?
- Inf integration tests fail, Can I still merge the PR? Ideally, this would be "no"
- It might be a bit cumbersome to lock-unlock, lock-unlock when you are fixing local tests.

alvaroloes (approved):
What a massive simplification. Thanks for answering! Now I'm all in with this solution.

PR #479, opened 2020-10-29
Reworked the way environments are managed by the CLI

javiertoledo:
## Description
This is a minor refactor. I noticed some details that the environment management in configuration that made me think that the way we were managing them was a bit fragile:

* We were dealing with system environment variables from the cli scripts
* We had a list of available environments in the config that was only used by the CLI to display them when there were errors. The config is the config for a specific environment, so it doesn't make sense that it has getters and setters for things that are unrelated.
* The tests were rewriting `require`

## Changes
* Created a new `environment` module with higher-level functions to write and read the current environment
* Moved the `configuredEnvironments` list from `BoosterConfig` to `BoosterApp`
* Fixed (and improved a little) the tests

## Checks
- [x] Project Builds
~~Updated documentation accordingly~~ (not needed)

### DIFF
--- a/packages/framework-core/src/booster.ts
+++ b/packages/framework-core/src/booster.ts
@@ -24,6 +24,7 @@ import { BoosterScheduledCommandDispatcher } from './booster-scheduled-command-d
  * - `region`: 'eu-west-1'
  */
 export class Booster {
+  public static readonly configuredEnvironments: Set<string> = new Set<string>()
   private static logger: Logger
   private static readonly config = new BoosterConfig(checkAndGetCurrentEnv())
   /**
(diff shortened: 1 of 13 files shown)


NickSeagull (approved):
Lovely!

packages/framework-core/src/booster.ts
```diff
@@ -42,7 +43,7 @@ export class Booster {
    * @param configurator A function that receives the configuration object to set the values
    */
   public static configure(environment: string, configurator: (config: BoosterConfig) => void): void {
-    this.config.addConfiguredEnvironment(environment)
+    this.configuredEnvironments.add(environment)
```
alvaroloes, line 46: I like this change. It seems this belongs here instead of the config class

alvaroloes (approved):

PR #493, opened 2020-11-16
fix: added booster token verifier class

charlietfe:
## Description
This PR contains an improvement to use the jwks client cache while the graphQL handler lambda is still executing. This change will allow us to avoid creating a jwks client on each token verification and also avoid hitting the key set endpoint producing a rate limit.

I did this change after detect that some tests were locked randomly without any reason, so maybe we've fixed the root case. 

## Changes
- Created a new class with all the functionality we had in Booster Auth.
- Fixed tests
- Set the cache limit to the max limit of the lambda execution time (AWS only). For now is enough.

## Checks
- [X] Project Builds
 
## Additional information
We'll need to research a bit more about how different providers use the jwks uris. If the providers invalidates them we should detect that, invalidate our cache and configure a new client.

### DIFF
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -31,25 +31,25 @@
       }
     },
     "@boostercloud/framework-core": {
-      "version": "0.8.2",
-      "resolved": "https://registry.npmjs.org/@boostercloud/framework-core/-/framework-core-0.8.2.tgz",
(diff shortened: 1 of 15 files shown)


juanjoman (approved):
LGTM!!

NickSeagull (approved):
LGTM!

PR #495, opened 2020-11-17
Fixed local integration tests

javiertoledo:
## Description
A new file introduced in an "automatic merge" missed some changes in the way local tests were refactored, causing errors.

## Changes
* Fixed the breaking tests

## Checks
- [ ] Project Builds
~- [ ] Updated documentation accordingly~

### DIFF
--- a/packages/framework-integration-tests/integration/providers/local/read-models.integration.ts
+++ b/packages/framework-integration-tests/integration/providers/local/read-models.integration.ts
@@ -4,19 +4,18 @@ import { changeCartItem, graphQLClient } from './utils'
 import { random } from 'faker'
 import { expect } from 'chai'
 import * as DataStore from 'nedb'
-import { readModelsDatabase } from '@boostercloud/framework-provider-local'
-import { ReadModelEnvelope } from '@boostercloud/framework-types'
 import { waitForIt } from '../aws/utils'
-import * as fs from 'fs'
+import { sandboxPath } from './constants'
+import * as path from 'path'
 
 describe('read-models', () => {
-  const readModels: DataStore<ReadModelEnvelope> = new DataStore(readModelsDatabase)
+  let readModels: DataStore<unknown>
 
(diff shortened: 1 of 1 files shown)


charlietfe (approved):

NickSeagull (approved):

packages/framework-integration-tests/integration/providers/local/read-models.integration.ts
```diff
@@ -4,19 +4,18 @@ import { changeCartItem, graphQLClient } from './utils'
 import { random } from 'faker'
 import { expect } from 'chai'
 import * as DataStore from 'nedb'
-import { readModelsDatabase } from '@boostercloud/framework-provider-local'
-import { ReadModelEnvelope } from '@boostercloud/framework-types'
```
javiertoledo, line 8: This is weird. In order to make this test pass in Github Actions (It works like a charm in dev machines (and I tried it in two different machines)), I had to remove this import to the package `@boostercloud/framework-types`. I'm not sure what's happening. I have checked that the right reference is in the `package.json`, deleted and recreated the lock files, and everything else that crossed my mind. If this test passed before merging PR #473, it could have something with a couple of changes I made recently:
* I removed from `gitignore` the `package-lock.json` files because of [this issue in Lerna](https://github.com/lerna/lerna/issues/2391) that was stopping the release.
* I sandboxed and symlinked with the versions under development with the test project before running local integration tests. This process changes the globally installed packages.

One or both of these changes could be causing the issue. One thing I'd propose is to get rid of yarn in favor of npm to start simplifying things a little bit. It looks like making `yarn.lock` and `package-lock.json` files coexist tends to cause issues...

NickSeagull: Yep, agree with this, lets try merging this thing and then get #496 merged as well :)

PR #496, opened 2020-11-19
Spike: Drop yarn in favor of npm

javiertoledo:
## Description

After noticing some weird issues with lerna, npm, and yarn, I started this PR to explore what it would take to drop yarn and use lerna with npm exclusively. In the process, I discovered that we were using some different dependency versions from different packages and there were missing references that were working just because we've been "lucky". This setup seems to be capable of detecting these situations faster.

## Changes

* Switching to npm required several changes:
  * The packages are now symlinked with the root monorepo project.
  * All dev dependencies are now in the root package. The leaf packages only contain production dependencies.
* All scripts are updated to use npm.
* I checked npm 7 new monorepo management capabilities, but I think we still need lerna, especially for the publish workflow.

## Open questions and pending tasks

- [ ] Tests are failing. As I've upgraded several npm packages in the process, so it's likely that I need to review that I've not introduced breaking changes accidentally.
- [ ] It's unclear if integration checks are going to work as expected, I'll need to investigate.
- [ ] In theory, with this setup, `lerna bootstrap` is not needed anymore, but I had to run it once to properly generate the projects `package-lock.json` files. I'm not sure how this works, so I'll need to make a few more tests.
- [ ] I have no idea how this change affects the publish script, as versions are now removed and changed by path references.

## Checks
- [ ] Project Builds


javiertoledo:
I went too far in this branch and I don't know what's going on anymore, so I'm closing it. We might retry doing this in the future if we find issues with yarn.

PR #506, opened 2020-12-16
Fix test failure when stack already exists

adriangmweb:
## Description
When deploying a Booter app that already exists in AWS, it throws an error because a test is expecting a `CREATE_COMPLETED` message from `CloudFromation`. This error is commonly seen after a failed integration test.

## Changes
The tests expect now either a `CREATE_COMPLETED` or a `UPDATE_COMPLETED` message, which allows updating an existing stack without errors.

## Checks
- [x] Project Builds


NickSeagull (approved):

rdiaz82 (approved):
LGTM

PR #518, opened 2020-12-18
Fix/event handlers promise all break

adriangmweb:
## Description
When an event handler failed, the rest don't get to execute because of a `promise.all` wapper behavior.

## Changes
* Updated root TypeScript version to match the version of the rest of the packages.
* Remove an unnecessary dependency for k8s that had conflicts with TypeScript new version.
* Changed `booter-event-dispatcher.ts` to use `Promise.allSettled` to avoid stopping the execution of all of them when one fails. 

## Checks
- [X] Project Builds
~- [] Updated documentation accordingly~ Not necessary

### DIFF
--- a/packages/framework-core/src/booster-event-dispatcher.ts
+++ b/packages/framework-core/src/booster-event-dispatcher.ts
@@ -37,7 +37,7 @@ export class BoosterEventDispatcher {
         case 'event':
           logger.debug('[BoosterEventDispatcher#eventProcessor]: Started processing workflow for event:', eventEnvelope)
           // TODO: Separate into two independent processes the snapshotting/read-model generation process from the event handling process
-          await Promise.all([
+          await Promise.allSettled([
             BoosterEventDispatcher.snapshotAndUpdateReadModels(eventEnvelope, eventStore, readModelStore, logger),
             BoosterEventDispatcher.handleEvent(eventEnvelope, config, logger),
           ])
--- a/packages/framework-core/src/booster-event-dispatcher.ts
+++ b/packages/framework-core/src/booster-event-dispatcher.ts
@@ -87,6 +87,6 @@ export class BoosterEventDispatcher {
     )
     results
       .filter((result) => result.status === 'rejected')
-      .forEach((result) => logger.debug(`Event handlere execution failed: ${(result as PromiseRejectedResult).reason}`))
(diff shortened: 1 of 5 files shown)


NickSeagull (approved):
Awesome, thanks a lot!

packages/framework-core/src/booster-event-dispatcher.ts
```diff
@@ -77,13 +77,16 @@ export class BoosterEventDispatcher {
...
       eventHandlers.map(async (eventHandler: EventHandlerInterface) => {
         const register = new Register(eventEnvelope.requestID, eventEnvelope.currentUser)
         logger.debug('Calling "handle" method on event handler: ', eventHandler)
         await eventHandler.handle(eventEnvelope.value as EventInterface, register)
         return RegisterHandler.handle(config, logger, register)
       })
     )
+    results
+      .filter((result) => result.status === 'rejected')
+      .forEach((result) => logger.debug(`Event handler execution failed: ${(result as PromiseRejectedResult).reason}`))
```
javiertoledo, line 90: This solution is fine for now, but we should separate event handlers to run in their own lambda each. I created issue #522 to do it in the future

javiertoledo (changes requested):
As we're upgrading TypeScript version and ES targets in this branch, it would be nice to check other uses of  `Promise.all` and consider replacing them by `Promise.allSettled` too. I think that in most scenarios we're using it only to await for all threads to complete and we're not taking into account the behavior of `Promise.all` under error circumstances.

adriangmweb:
I changed other `promise.allSettled` to avoid errors in some files. There are other `promise.all` that have the potential to be changed, but I believe that in those cases it would be better to fail for now as it would require a huge rewrite in order to resolve type conflicts.

tsconfig.json
```diff
@@ -12,6 +12,10 @@
     "noUnusedParameters": false,
     "noFallthroughCasesInSwitch": true,
     "forceConsistentCasingInFileNames": true,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "lib": [
+      "ES2019",
+      "ES2020"
+    ]
```
alvaroloes, line 19: What is this used for?

alvaroloes (approved):
Just a question!

tsconfig.json
adriangmweb: To be able to use `Promise.allSettled` with TypeScript

javiertoledo (approved):
Looks good now!

Issue #517, opened 2020-12-18
Untangle the dynamic requires needed by the CLI to deploy a project


javiertoledo:
## Description

Different parts of Booster run in very different environments to make possible a seamless multi-cloud developer experience:
* The code is written and debugged in a local development machine.
* It can be deployed to a cloud provider from the development machine or a CI/CD environment.
* The user project is executed in a FaaS environment.

In the two first scenarios, the entry point is the CLI, in the last one, it’s the user’s project. In all cases, the user project is loaded to read the dynamic `BoosterConfig` object that includes references to internal objects and classes.

In the following diagram, we show the current package architecture in Booster and how each package is related to the others. I’ve omitted the framework-types package for simplicity. Here we can see the two entry points and how the different packages are loaded for what reasons.

![Booster Architecture](https://user-images.githubusercontent.com/175096/102623296-30f1e580-413a-11eb-9a5a-7c3453c5e54c.png)

Some things that might call our attention are:
* The reason for the cli module to load the user’s project is to read the BoosterConfig object, which is a dynamic object that is generated when the project is initialized.
* The reason why the cli needs to load the BoosterConfig dynamically is that it provides a function to load the Provider Infrastructure Package.
* The only reason why the Provider Implementation needs to be aware of the Provider Infrastructure Implementation is to load it dynamically, which is only required to provide this functionality to the cli.

The way Booster currently loads these dependencies and the high level of indirection is making it hard to debug the deployment process and producing issues with dependencies that our users have been reporting like:

* https://github.com/boostercloud/booster/issues/509
* https://github.com/boostercloud/booster/issues/510

## Possible Solution

Introduce a pre-processing step as part of the project build process to generate and dump a serializable version of the `BoosterConfig` object. This could be either dumped to a file or just written in the standard output to allow piping it to other commands.

This configuration file will contain enough information about the code to allow the infrastructure modules to generate table names, API endpoint names, and any other details required by the deployment, nuke, or start processes. This should allow us to refactor the Booster architecture as follows, removing the current dependencies between the CLI and the user project, and likely solving the missing dependency issues we're experiencing:

![Booster Architecture with static BoosterConfig](https://user-images.githubusercontent.com/175096/102623893-1bc98680-413b-11eb-8562-ed9cf8ebd002.png)

NickSeagull:
[...]

PR #524, opened 2020-12-22
Removed the dynamic require of the infrastructure package through the provider library to load it directly from the cli

javiertoledo:
## Description

In this PR I've removed the `infrastructure` function from the `ProviderLibrary` type and replaced it by a `packageDescriptor` function that returns an object with the name and version of the provider library.

With this change, the `framework-provider-*-infrastructure` packages are independent of the `framework-provider-*` packages, hopefully avoiding dependency issues when the infrastructure packages were not defined in the user project's `package.json` file.

This is the first step towards the changes proposed in #517 to fully decouple the runtime from the infrastructure code.

## Changes
- [x] Removed the `infrastructure` function from `ProviderLibrary`.
- [x] Added a new `rockets` option to `BoosterConfig` to set up the rockets instead of passing them as a parameter to the provider library initialization (This changes slightly the way you configure rockets, see the doc changes)
- [x] Unified all dynamic requires in the `cli` module in a single `dynamicLoad` service.
- [ ] Checks that packages loaded dynamically are installed and that the version matches before trying to execute them
- [x] Fixed unit tests
- [ ] Fixed integration tests 

## Checks
- [x] Project Builds

### DIFF
--- a/packages/cli/src/services/dynamic-loader.ts
+++ b/packages/cli/src/services/dynamic-loader.ts
@@ -1,0 +1,6 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export async function dynamicLoad(module: string): Promise<any> {
+  // We wrap the call to import here to ease testing dynamic module loading
+  // The `import` expression returns a promise that will be rejected when the module is not found
+  return import(module)
+}
--- a/packages/cli/src/services/dynamic-loader.ts
+++ b/packages/cli/src/services/dynamic-loader.ts
@@ -1,6 +1,12 @@
+import * as path from 'path'
+
 /* eslint-disable @typescript-eslint/no-explicit-any */
-export async function dynamicLoad(module: string): Promise<any> {
-  // We wrap the call to import here to ease testing dynamic module loading
+export async function dynamicLoadFile(filePath: string): Promise<any> {
+  // Import is always relative to current file, so we need to "relativize" the path
+  return dynamicLoadModule(path.relative(__dirname, filePath))
+}
+
+export async function dynamicLoadModule(packageName: string): Promise<any> {
   // The `import` expression returns a promise that will be rejected when the module is not found
-  return import(module)
+  return import(packageName)
 }
--- a/packages/cli/src/services/provider-service.ts
+++ b/packages/cli/src/services/provider-service.ts
@@ -1,5 +1,5 @@
 import { BoosterConfig, Logger, ProviderInfrastructure } from '@boostercloud/framework-types'
-import { dynamicLoad } from './dynamic-loader'
+import { dynamicLoadModule } from './dynamic-loader'
 
 export function assertNameIsCorrect(name: string): void {
   // Current characters max length: 37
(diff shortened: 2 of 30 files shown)


packages/cli/src/services/provider-service.ts
```diff
@@ -23,26 +24,61 @@ export function assertNameIsCorrect(name: string): void {
...
-    )
-  }
-  return method
 }
 
-export function deployToCloudProvider(config: BoosterConfig, logger: Logger): Promise<void> {
-  return supportedInfrastructureMethodOrDie('deploy', config)(config, logger)
+/**
+ * We load the infrastructure package dynamically here as a cli plugin to avoid
+ * including it among the user project's dependences.
```
javiertoledo, line 33:
```suggestion
 * including it among the user project's dependencies.
```

packages/cli/src/services/dynamic-loader.ts
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export function dynamicLoad(module: string): any {
+  return require(module)
+}
```
NickSeagull, line 4: I'd do this with `unknown` to force the caller to specify the type
```suggestion
export function dynamicLoad(module: string): unknown {
  return require(module)
}
```

packages/cli/test/services/dynamic-loader.test.ts
```diff
@@ -0,0 +1,12 @@
+import { dynamicLoad } from '../../src/services/dynamic-loader'
+import { expect } from '../expect'
+
+describe('dynamic-loader service', () => {
+  describe('dynamicLoad', () => {
+    it('behaves as require', () => {
+      const packageJson = dynamicLoad('../../package.json')
+
+      expect(packageJson.name).to.equal('@boostercloud/cli')
```
NickSeagull, line 9: Clever!

NickSeagull (approved):
Looks good!

javiertoledo:
Closing, as I'm working on this on #530

javiertoledo:
Reopening, as we're discarding #528 and #530 doesn't make sense anymore

alvaroloes:
I don't quite understand much the problem this PR is trying to resolve. The description says that we avoid the user adding the infrastructure package to their dependencies. How is this possible? In the end, the user needs to do an "npm install" and that dependency should be in the machine, correct? With Booster 0.10.x, if the user has the infrastructure packages in their devDependencies, everything works.

On the other hand, I see changes in this PR that I like a lot, like the `config.rockets` property 😃 . One question regarding that: if the rocket is only for a specific infrastructure provider, How is that handled? What would happen if I specify a rocket for AWS but then I deploy using Azure, for example?

javiertoledo:
> I don't quite understand much the problem this PR is trying to resolve. The description says that we avoid the user adding the infrastructure package to their dependencies. How is this possible? In the end, the user needs to do an "npm install" and that dependency should be in the machine, correct? With Booster 0.10.x, if the user has the infrastructure packages in their devDependencies, everything works.
> 
> On the other hand, I see changes in this PR that I like a lot, like the `config.rockets` property 😃 . One question regarding that: if the rocket is only for a specific infrastructure provider, How is that handled? What would happen if I specify a rocket for AWS but then I deploy using Azure, for example?

The main idea here is to treat the infrastructure packages as CLI plugins to break the current dependency cycle that requires the CLI to load the user's project to load the provider package to load the infrastructure package. This is just a first step towards the changes proposed in this Confluence page: https://agilemonkeys.atlassian.net/wiki/spaces/B/pages/1798504449/Booster+multi-language+support+iteration+2

Anyways, this is a work in progress and there are some loose ends yet.

If someone loads a rocket for the wrong provider, it will fail when you try to use it :trollface:

javiertoledo:
This code became a mess after the refactors made to adopt npm in #545. I've tried to fix it, but it's hard to make the integration tests pass and I suspect that, right now, it's going to be easier to reimplement the change than making this PR work, so I'm closing it.

PR #525, opened 2020-12-22
Fix/booster 503 remove real file operations from cli unit tests

davidverdu:
## Description
**CLI new** unit tests have been refactored to not write files on the file system, with the aim of run the test faster and without collateral effects. I mocked some **fs** methods and **init** function to avoid file creation during tests.

## Changes
- the test file has been refactored
- I added the **resources** folder to store assets used in our tests

## Checks
- [x] Project Builds


davidverdu:
we have the complete PR here: https://github.com/boostercloud/booster/pull/526

Issue #492, opened 2020-11-16
new:project fails if quotes ocurr in any of the fields


NickSeagull:
## Current Behavior

The quotes aren't being escaped, resulting in a malformed `package.json` file

## Expected behavior

Quotes are escaped

## Possible Solution

Perform a replace of the `'` and `"` characters with `\'` and `\".

#### Environment

- Booster version: 0.9.0

davidverdu:
Replacing ' was not necessary. Replacing " with \" fixes the bug

PR #526, opened 2020-12-23
Tests/502 cli validations, unit tests, fix double quote bug

davidverdu:
## Description
- Add new unit tests to CLI services and commands. 
- Existing CLI new unit tests have been refactored to not write files on the file system, with the aim of run the test faster and without collateral effects. I mocked some fs methods and init function to avoid file creation during tests.
- Adding some CLI validations to fix the following bugs: #535, #537
- Fixes the double quote bug on config prompt: #492

## Changes
- New unit test files are added to the cli test directory
- resources folder stores mock projects used in unit tests
- Adding validations to parsing.ts file
- Escaping " in Prompter class

## Checks
- [x] Project Builds


.gitignore
```diff
@@ -12,4 +12,6 @@ coverage
 scripts/mirroring.js
 .docs
 .booster
-*-integration-sandbox
\ No newline at end of file
+*-integration-sandbox
+
+!/packages/cli/resources/**/*
```
javiertoledo, line 17: Why do you added this rule? This folder should be included in the repo anyway, isn't it?

```suggestion
```

NickSeagull (changes requested):
I believe that there should **not** be a mock project for unit tests. Unit tests just check that some methods are called, and not that stuff is generated on the file system. That is responsibility of the integration tests :)

I'd remove the mock project completely, and would rely on the integration tests to do the checks that you want to do.

Also, try having small PRs for the next time addressing one issue at a time 😄

davidverdu:
I think that the best is to move the mock project tests I made to the integration tests folder, so I don't break the structure. It is my first PR in the project and I didn't notice the integration tests and unit tests separation in the project

javiertoledo (approved):
Amazing work @davidverdu, I left a few minor comments, and you'll probably need to rebase this branch to master and fix the places where you check for calls to yarn via `exec`, but everything else looks good to me.

javiertoledo:
> I believe that there should **not** be a mock project for unit tests. Unit tests just check that some methods are called, and not that stuff is generated on the file system. That is responsibility of the integration tests :)
> 
> I'd remove the mock project completely, and would rely on the integration tests to do the checks that you want to do.

@NickSeagull notice that these mock projects are there because he's unit-testing the generators comparing the result with the text in the mocks. I think that having these files there is easier to maintain than having strings in the tests themselves and I've seen this pattern in other frameworks too. What's important in unit tests is avoiding effectful operations in the code under test, what we have here is basically a way to store test data.

packages/cli/src/services/user-prompt.ts
```diff
@@ -3,10 +3,10 @@ import inquirer = require('inquirer')
 export default class Prompter {
   public async defaultOrPrompt(value: string | undefined | null, promptMessage: string): Promise<string> {
     if (value) {
-      return Promise.resolve(value)
+      return Promise.resolve(value.replace(/\"/g,"\\\""))
     } else {
       const res = await inquirer.prompt([{ name: 'value', type: 'input', message: promptMessage }])
-      return Promise.resolve(res['value'])
+      return Promise.resolve(res['value'].replace(/\"/g,"\\\""))
```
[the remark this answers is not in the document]
davidverdu: @javiertoledo This fixes this issue: https://github.com/boostercloud/booster/issues/492

.gitignore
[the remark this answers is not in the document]
davidverdu: the problem is that I need to commit the content inside `/packages/cli/resources/mock_project/dist` and `dist` is generally being ignored in .gitignore.

NickSeagull:
@javiertoledo in that case, I'd create a fixtures folder, but here the project included JavaScript files and other stuff.

packages/cli/src/services/generator/target/parsing.ts
```diff
@@ -7,12 +7,26 @@ export const parseName = (name: string): Promise<HasName> => Promise.resolve({ n
...
+        duplicated = field.name
+      }
+      usedFieldNames.push(field.name)
+    })
+    if (duplicated !== undefined) {
+      return Promise.reject(fieldDuplicatedError(duplicated))
+    }
+    return { fields }
+  })
+}
```
alvaroloes, line 25: Just a suggestion: if you work with promises, consider using an `async` function. This way, the code inside the function can throw normal errors or return values without the need of nesting everything inside promises, making it more readable. 

Also, as you are finding duplicates, you can leverage the native `Set` object that automatically removes duplicates for you and allows you to do set operations, like intersections. This is a suggestion for the code that combines the two things I have mentioned here:

```typescript
const parseFields = async (fields: Array<string>): Promise<HasFields> => { // <-- Notice the "async" prefix
  const parsedFields = fields.map(parseField) 
  const parsedFieldNames = parsedFields.map((field) => field.name) // Get only the field names
  const uniqueFieldNames = new Set(parsedFieldNames)               // Create a set of unique field names
  const duplicates = parsedFieldNames.filter((field) => !uniqueFieldNames.delete(field)) // This "intersection" extracts all duplicates
  if (duplicates.length > 0) {
      // This "throw" is converted automatically to a Promise.reject()
      // (BTW, we should change this to the correct error message)
      throw new Error("Duplicates: " + duplicates.join(', ')) 
  }
  return parsedFields // This "return" is automatically converted to a Promise.resolve()
}
```

packages/cli/test/commands/deploy.test.ts
```diff
@@ -141,4 +151,171 @@ describe('deploy', () => {
...
+      replace(oraLogger,'fail', fake.resolves({}))
+      replace(oraLogger, 'info', fake.resolves({}))
+      replace(oraLogger, 'start', fake.resolves({}))
+      replace(oraLogger, 'succeed', fake.resolves({}))
+    })
+
+    it('without flags', async () => {
+      await new Deploy.default([], {} as IConfig).run()
+
+      expect(childProcessPromise.exec).to.have.not.been.calledWithMatch('npx yarn clean && npx yarn compile')
```
alvaroloes, line 168: This should change not to `npm`

alvaroloes (commented):
Left some comments and the following question: Why we need to include a complete fake project including its `dist` folder?

.gitignore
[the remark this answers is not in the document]
javiertoledo: Maybe you can alter the `dist` rule to be more specific. Could something like `packages/*/dist/**/*` ignore the dist folder in the packages but not that one? Im not 100% sure about the syntax, but I bet there's something like that that works.

packages/cli/test/commands/new/entity.test.ts
```diff
@@ -0,0 +1,303 @@
...
+      })
+
+      it('creates Entity with a string field reducing PostCreated', async () => {
+        await new Entity([entityName, '--fields', 'title:string', '--reduces', 'PostCreated'], {} as IConfig).run()
+        const renderedEntity = Mustache.render(templates.entity, {
+          imports: reducingEntityImports,
+          name: entityName,
+          fields: [{ name: 'title', type: 'string' }],
+          events: [{ eventName: 'PostCreated' }],
+        })
```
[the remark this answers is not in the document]
davidverdu: I simplified the tests but in the end I didn't use `loadFixture`

.gitignore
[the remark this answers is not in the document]
davidverdu: I removed a lot of fixtures and this .gitignore line is no longer necessary

packages/cli/src/services/generator/target/parsing.ts
davidverdu: Good suggestions by the way. I need to study more in depth the await, async, Promise syntax. I have applied the changes @alvaroloes

davidverdu:
I have applied the suggested changes and improved the unit tests to make them more maintainable @NickSeagull @javiertoledo @alvaroloes . I have also rebased with `main`

packages/cli/test/commands/new/entity.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      })
+    })
+
+    describe('displays an error', () => {
+      it('with empty Entity name', async () => {
+        replace(console, 'error', fake.resolves({}))
+        await new Entity([], {} as IConfig).run()
+        expect(fs.outputFile).to.have.not.been.calledWithMatch(entitysRoot)
+        expect(console.error).to.have.been.calledWith(
+          "You haven't provided an entity name, but it is required, run with --help for usage"
```
alvaroloes, line 177: As error messages and logs could easily change without affecting the correctness of the application, I would suggest writing expectations on those messages using regular expressions that checks only the main part of the message, for example in this case it would be:
```suggestion
          /You haven't provided an entity name/
```
This way, it is less cumbersome to do code changes and refactors that only affect these messages.

What do you think? 

By the way, this message also applies to other similar expectations in this PR

alvaroloes (approved):
OMG! What an IMPRESSIVE work in tests @davidverdu 💥 

Thanks a lot!!! 🥇 

Approved. Just lest a suggestion to add some flexibility in the expectations about log/error messages

packages/cli/test/commands/new/entity.test.ts
davidverdu: Suggested changes applied! You are right!

NickSeagull (approved):
Thank you! 🙏

PR #528, opened 2020-12-23
Reverted changes to prune devDependencies using `yarn install --production`

javiertoledo:
The deployment process was running `yarn install --production` under the hoods with the intention of removing the devDependencies, but it also unexpectedly removed secondary dependencies (dependencies of our dependencies), making lambdas fail once deployed, and forcing us to explicitly add missing dependencies as direct project dependencies.

By doing this, we're bringing back the lambda size issues, so we won't generate a version when merging this PR. Users are currently using a workaround that consists of adding the infra package as a project dependency and use yarn, which seems to produce smaller packages.

I'm currently working on PR #524 to change the way infrastructure dependency is loaded, which should allow us to remove the dependencies related to infrastructure from user projects.

This change's main goal is to make integration tests pass in currently blocked PRs.

## Checks
- [x] Project Builds

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -6,16 +6,15 @@ import { Script } from '../common/script'
 import Brand from '../common/brand'
 import { logger } from '../services/logger'
 import { currentEnvironment, initializeEnvironment } from '../services/environment'
-import { installAllDependencies } from '../services/dependencies'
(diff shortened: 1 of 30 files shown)


javiertoledo:
It looks like we never documented the flag `--skipRestoreDependencies` for `boost deploy`, so no documentation changes are needed.

javiertoledo:
Closing, as I'm working on this in #530

PR #530, opened 2020-12-26
Refactored the way the CLI loads the infrastructure dependencies and rolled back the prune devDependencies feature

javiertoledo:
This is the combination of PRs #524 and #528 because I suspect that these changes are not going to be able to pass integration tests unless they work together. If this is true, I'll delete these other branches and move the descriptions here.


javiertoledo:
Closing this PR as we've reopened #524

PR #533, opened 2021-01-08
Add New Rocket: Upload file to S3 and new event will be persisted in the Event Store

rdoria1:
## Description
Add new Rocket:
- Create new Bucket with event notifications.
- Create new Lambda (triggered by event notification) that will persist a new event each time a new file is uploaded to the bucket. The event will have the s3URI of the file and its size

## Changes
Add new package for new rocket + upadte lerna.json and packge.json with new package.

## Checks
- [X] Project Builds


packages/rocket-uploads3-store-event-aws-infrastructure/.npmignore
```diff
@@ -0,0 +1 @@
+!src/lambdas/node_modules
```
juanjoman, line 1: @rdoria1 now thinking about this, I don' know if the github action can publish your rocket to the npm registry so maybe this won't work 100% as expected. For example the static rocket is published: https://www.npmjs.com/package/@boostercloud/rocket-static-sites-aws-infrastructure

juanjoman: Also for those others looking at this PR, this `.npmignore` file is set because the previous `.gitignore` (root folder) isn't allowing us to publish the lambda `node_modules`. We tried with another inner `.gitignore` and other solutions but it wasn't possible. We need to use `.npmignore` to whitelist the lambda node modules and publish using npm instead of yarn 😢

packages/rocket-uploads3-store-event-aws-infrastructure/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/boostercloud/booster.git"
+  },
+  "dependencies": {
+    "@aws-cdk/aws-dynamodb": "1.67.0",
+    "@aws-cdk/aws-lambda": "1.67.0",
+    "@aws-cdk/aws-lambda-event-sources": "1.67.0",
+    "@boostercloud/framework-types": "^0.9.3",
+    "aws-cdk": "1.67.0",
```
juanjoman, line 30: is it possible to just import @aws-cdk/core? I think this dependency only needs the Stack object

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
```diff
@@ -0,0 +1,44 @@
+import { AWSError } from 'aws-sdk'
+import { PutItemOutput } from 'aws-sdk/clients/dynamodb'
+
+export const handler = async (event: any): Promise<any> => {
+
+  const AWS = require('aws-sdk')
```
juanjoman, line 6: Can't you import this as `import * as AWS from { aws-sdk }`? 😮

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  const { v4: uuidv4 } = require('uuid');
+  const ddb = new AWS.DynamoDB.DocumentClient()
+
+  const bucketName = event.Records[0].s3.bucket.name
+  const objectKey = event.Records[0].s3.object.key
+  const fileSize = event.Records[0].s3.object.size
```
juanjoman, line 12: Super nitpicky: If you know the event content, could you change the type?
`async (event: <known_type>)`

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
```diff
@@ -0,0 +1,68 @@
...
+    const eventsStore = stack.node.tryFindChild('events-store') as Table
+
+    // Crete new Lambda function to be triggered once new file is uploaded in bucket
+    const processorFunction = new Function(stack, 'rocketS3Trigger' + params.entityTypeName, {
+      runtime: Runtime.NODEJS_12_X,
+      timeout: Duration.minutes(15),
+      memorySize: 1024,
+      handler: 'index.handler',
+      functionName: appName + '-' + params.entityTypeName.toLowerCase() +'-s3-rocket-trigger',
+      code: Code.fromAsset(path.join(__dirname, 'lambdas')),
```
juanjoman, line 36: Does this work without zipping the lambda folder? Just saying because of the node_modules folder.

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
```diff
@@ -0,0 +1,68 @@
...
+      events: [EventType.OBJECT_CREATED],
+    })
+    processorFunction.addEventSource(uploadEvent)
+
+    // Grant the Events handler lambda access to read from S3 bucket
+    const eventsHandlerLambda = stack.node.tryFindChild('events-main') as Function
+    eventsHandlerLambda.addToRolePolicy(
+      createPolicyStatement([sourceBucket.bucketArn, sourceBucket.bucketArn + '/*'], ['s3:*'])
+    )
+  }
```
juanjoman, line 62: Isn't this allowing all operations from S3? Shouldn't we restrict this a bit? I mean because of the `s3:*` but let me know if I'm wrong 🙏

juanjoman (changes requested):
Great idea for a rocket to be honest!! Just left some comments below 😛

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
rdoria1: Actually this policy is meant to give total control of the S3 bucket from the event handler. The code will be defined in the Booster app and we don't what the user will decide to do from the rocket implementation. This is why I decided to give them total access. Does that make sense @juanjoman ?

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
rdoria1: yes, it works fine without zipping I tested this in a real app

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
juanjoman: Amazing then!! :v:

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
rdoria1: @juanjoman I think we can yes, do we have any guidelines what is the preferred way for these imports? require vs from, * vs specific imports.
Asking because I don't really know them hehe :P

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
rdoria1: ah yes! I think the type is Records[], I will try :)

packages/rocket-uploads3-store-event-aws-infrastructure/.npmignore
rdoria1: yes I think you are right. As of now it won't compile because it can't find the nodes modules of the lambda. I think I need to add it OR I need to add an action form the rocket so that it it install the node modules of the lambda before compiling the rocket. What do you think? @juanjoman

juanjoman: Yeah for this case I think you can set the compile script like: `yarn compile <lambda_folder> && cp -R <origin> <destiny>`. The thing is that `lerna run compile` is running a `yarn compile` on the root of the rocket @rdoria1 .

rdoria1: Yeah actually I'll add 
`"postinstall": "yarn --cwd src/lambdas install",` to the package.json rocket and I think it will work :)

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
```diff
@@ -0,0 +1,68 @@
...
+  public static mountStack(params: AWSS3EventsLambdaParams, stack: Stack): void {
+    const appName = stack.stackName.split('-app')[0]
+
+    // Create new bucket to upload file
+    const sourceBucket = new Bucket(stack, 'sourceUploadBucket' + params.entityTypeName, {
+      bucketName: params.bucketName,
+      removalPolicy: RemovalPolicy.DESTROY,
+    })
+
+    const eventsStore = stack.node.tryFindChild('events-store') as Table
```
charlietfe, line 27: Why not pass the events-store table as a parameter? Or `events-store` is fixed and isn't configurable? Makes sense?

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  const s3uri = `s3://${bucketName}/${objectKey}`
```
charlietfe, line 13:
```suggestion
 const { bucket: { name }, object: { key, size } } = event.Records[0].s3
 const fileSize = size
 const s3uri = `s3://${name}/${key}`
```

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
charlietfe: There are some differences between how `import` and `require` load the libraries and also it depends on how the module was exported. https://blog.atomist.com/typescript-imports

charlietfe (approved):
Some comments, but in general LGTM.

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
charlietfe: Since you have the names of the resources inside the Booster config  object you could use that one passing by via rocket parameters:

```typescript
  config.appName = 'process-files'
  config.provider = Provider([
    {
      packageName: '@boostercloud/rocket-uploads3-store-event-aws-infrastructure',
      parameters: {
        bucketName: 'process-big-file-rocket',
        eventTypeName: 'BigFileAdded',
        entityTypeName: 'BigFile',
        eventStore: config.resourceNames.eventsStore,
      },
    }
  ])
})
```
And then in do this `const eventsStore = stack.node.tryFindChild(params.eventStore) as Table`

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export class S3NotificationEventStoreStack {
+  public static mountStack(params: AWSS3EventsLambdaParams, stack: Stack): void {
+    const appName = stack.stackName.split('-app')[0]
+
+    // Create new bucket to upload file
+    const sourceBucket = new Bucket(stack, 'sourceUploadBucket' + params.entityTypeName, {
+      bucketName: params.bucketName,
+      removalPolicy: RemovalPolicy.DESTROY,
+    })
```
javiertoledo, line 25: Ideally, an existing user should be able to upload files directly from their client application to S3. I didn't find a policy that allows this in here, so how does this work? Also, it would be ideal to integrate this module with Booster's role system, so we can grant upload permissions for a user role but not to another, or avoid anonymous users to upload a file. Any ideas around that? (We could perfectly leave this for a future PR)

packages/rocket-uploads3-store-event-aws-infrastructure/yarn.lock
```diff
@@ -0,0 +1,2043 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
javiertoledo, line 1: Notice that while this PR was open, we've switched to npm, so please rebase it to master and delete the yarn.lock file

javiertoledo (approved):
This is FANTASTIC, just left a few comments. One thing that we have not defined yet is how we're going to test rockets, we should prepare some tools to help rocket developers to build unit and integration tests for the things they create.

NickSeagull (changes requested):
Looks good overall! Instead of comments, I'd extract the code underneath them into functions with meaningful names, which makes code more understandable and manageable

packages/rocket-uploads3-store-event-aws-infrastructure/.npmignore
rdoria1: @juanjoman @javiertoledo I already rebase the branch with the newest changes and even with npm we have the same problem,the lambda folder won't compile nor install any dependencies. I think this is expected. The lambda folder is a sub folder of the package and by default it will only look into the package.json of the root folder. 

I added `"postinstall": "npm --prefix src/lambdas install",`

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
rdoria1: @charlietfe This was my initial implementation but I changed to the current one. I think exposing the event store name to the end user isn't a good idea. 

The main reason behind this reasoning is, this event store name can only have one valid value and only one. By exposing this as a param to the users, we are giving them the chance to make a mistake and put a different value meaning the rocket will fail. I really believe we shouldn't expose this value to them.

Also, I believe the event store name is an internal detail of the framework.

I would keep it as it is, thoughts? @javiertoledo @charlietfe @NickSeagull ?

lerna.json
```diff
@@ -12,7 +12,8 @@
     "packages/framework-core",
     "packages/cli",
     "packages/framework-integration-tests",
-    "packages/rocket-static-sites-aws-infrastructure"
+    "packages/rocket-static-sites-aws-infrastructure",
+    "packages/rocket-uploads3-store-event-aws-infrastructure"
```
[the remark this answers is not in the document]
rdoria1: @javiertoledo @NickSeagull That's what I thought initially but then the rocket is specific to AWS, it even has its name on it (it includes `aws-infrastructure` on the name) so I thought being specific to the tech stack made sense.

What do you think? @NickSeagull @javiertoledo

packages/rocket-uploads3-store-event-aws-infrastructure/yarn.lock
rdoria1: done :)

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
rdoria1: ` an existing user should be able to upload files directly from their client application to S3` => Not sure I'm following. As of now, when we deploy our booster project, we can navigate to the AWS console and then upload the file to be processed there.

` it would be ideal to integrate this module with Booster's role system` => Yes! Totally makes sense. Let's tackle this on a separate PR :)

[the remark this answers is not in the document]
rdoria1: I guess this is more a matter of styles, but I personally like comments better and I think they abstract the idea of your code without looking the code. 

In any case, I think we should define a coding style for CDK resources for upcoming PRs and maybe now is the time to discuss about this hehe.

Any thoughts? @charlietfe @juanjoman @javiertoledo @alvaroloes @NickSeagull ?

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
```diff
@@ -0,0 +1,68 @@
...
+      memorySize: 1024,
+      handler: 'index.handler',
+      functionName: appName + '-' + params.entityTypeName.toLowerCase() +'-s3-rocket-trigger',
+      code: Code.fromAsset(path.join(__dirname, 'lambdas')),
+      environment: {
+        EVENT_STORE_NAME:eventsStore.tableName,
+        ENTITY_TYPE_NAME: params.entityTypeName,
+        TYPE_NAME: params.eventTypeName,
+      },
+    })
```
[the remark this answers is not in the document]
rdoria1: ditto (see comment above)

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
rdoria1: great suggestion! :)

packages/rocket-uploads3-store-event-aws-infrastructure/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@boostercloud/rocket-uploads3-store-event-aws-infrastructure",
+  "version": "0.9.5699",
```
NickSeagull, line 3: The minor version looks very weird, should be probably according to the Booster version

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    bucket: { name },
+    object: { key, size },
+  } = event.Records[0].s3
+  const fileSize = size
+  const s3uri = `s3://${name}/${key}`
+
+  const boosterEvent = {
+    s3uri: s3uri,
+    fileSize: fileSize,
+  }
```
NickSeagull, line 18: Maybe this can be a function like this?

```suggestion
const boosterEvent = s3EventToBoosterEvent(event)

// ...
// somewhere else in the module
function s3EventToBoosterEvent(event: /* type of the event */): {s3uri: string, fileSize: string} {
  const {
    bucket: { name },
    object: { key, size },
  } = event.Records[0].s3
  const fileSize = size
  const s3uri = `s3://${name}/${key}`

  return {
    s3uri,        // you can omit setting the value if the variable for the value has the same name as the key
    fileSize,
  }
}
```

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    },
+  }
+
+  await ddb
+    .put(params, function(err: AWSError, data: PutItemOutput) {
+      if (err) {
+        console.log('Error', err)
+      } else {
+        console.log('Success', data)
+      }
```
NickSeagull, line 41: I'd make this a bit more verbose for the sake of the users, while at the same time, removing some `if`:
```suggestion
      const logData = 
          err
          ? ['An error ocurred while performing a PutItem operation', err]
          : ['Success performing PutItem', data]
      console.log('[ROCKET#upload-s3-store-event]', ...logData)
```

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
NickSeagull: Agree with @rdoria1

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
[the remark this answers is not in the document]
NickSeagull: But then a function name is the same as adding the comment, plus, the fact that it is a machine (compiler/interpreter) delimited way of grouping code of the same responsibility. For the same reason, while code evolves, comments are much harder to maintain. Ending up in stuff like this 😅

not basmati rice

I believe that code comments are useful for explaining design decisions, e.g. in this same PR the decision of not passing the `eventStore` as a parameter, but not for anything else (of course, there are exceptions, but very minimal).

I think that we should avoid comments here as well, code is still code, be it CDK or not. If YAML had functions, I'd probably use them there 😂

NickSeagull (dismissed):

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
```diff
@@ -0,0 +1,44 @@
+import { AWSError, DynamoDB } from 'aws-sdk'
+import { PutItemOutput } from 'aws-sdk/clients/dynamodb'
+
+export const handler = async (event: any): Promise<any> => {
```
juanjoman, line 4: Just missing this type above `(event: any)`

juanjoman (approved):
LGTM, just left the type thing as a suggestion!

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
NickSeagull: Which makes me think, perhaps it should be `unknown`, rather than `any`?

juanjoman (changes requested):
Oh I just noticed that this rocket doesn't have a readme, can you add one please? 🙏

lerna.json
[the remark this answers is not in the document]
javiertoledo: From the user perspective, why does that matter? When naming rockets, I'd rather focus on naming the feature more than the technologies under it. This is "the rocket that enables file uploads in AWS"; AWS is the "environment" here, not a specific implementation detail. I think that it makes sense to think about rockets in this way. Anyway, Rocket authors will be able to name them in any way they want, so we won't enforce this way of naming things in any case!

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
javiertoledo: Gotcha, we don't need to add this functionality in this PR, but I think that it would make a lot of sense to integrate this rocket with Booster's roles to also provide an API that someone could use to upload a file from a web browser. As it is today, it's restricted to users that can authenticate in the AWS console.

packages/rocket-uploads3-store-event-aws-infrastructure/.npmignore
alvaroloes: @rdoria1 @juanjoman Would this bug be automatically fixed if the rocket were in a separate repo?

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
alvaroloes: @rdoria1 Now that the `config` object is passed to all rockets (the PR was merged), you can get now the name from it 😃

lerna.json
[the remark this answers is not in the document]
alvaroloes: Also, @rdoria1 the idea is to make rockets provider un-aware by creating different implementations for the providers. So the name would be generic, let's say "fileuploader". Then each specific provider would have its package in the same way we have structured the framework packages:
- "packages/rocket-fileuploader-aws" (If this rocket adds some functions for the runtime, they would go in this package)
- "packages/rocket-fileuploader-aws-infrastructure"
- "packages/rocket-fileuploader-local"
- "packages/rocket-fileuploader-local-infrastructure"
- "packages/rocket-fileuploader-types" (If the rocket define types that are shared among the different providers, they'd go here)

Etc. The idea is to always have in mind the "provider unaware" nature of the framework. This is crucial for the local provider for example: people would like to test this rocket first in their local machines.

For more information about the rockets naming convention, @javiertoledo wrote some information in the docs: https://github.com/boostercloud/booster/tree/main/docs#naming-recommendations

packages/rocket-uploads3-store-event-aws-infrastructure/package.json
alvaroloes: hahaha, what happened here?

packages/rocket-uploads3-store-event-aws-infrastructure/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "description": "Booster rocket to upload file in S3 and store event in Event Store when it happens.",
```
alvaroloes, line 4: I would also change this description not to be provider-specific

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
alvaroloes: There are types for every AWS event, there must be something like `S3FileUploaded`, for example. 

Regarding the response, I think there must be one. If not, we can use there `unknown` as nick suggeted

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const {
+    bucket: { name },
+    object: { key, size },
+  } = event.Records[0].s3
+  const fileSize = size
+  const s3uri = `s3://${name}/${key}`
+
+  const boosterEvent = {
+    s3uri: s3uri,
```
alvaroloes, line 16: Could you please change this to `fileURI`? The idea is not to expose provider-aware names.

Also, Could you please create an exported type (typescript interface) for this so that people can use it in the definitions of their events?

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  await ddb
+    .put(params, function(err: AWSError, data: PutItemOutput) {
+      if (err) {
+        console.log('Error', err)
+      } else {
+        console.log('Success', data)
+      }
+    })
+    .promise()
```
alvaroloes, line 43: When you use `await`, the result would be returned if there is success and an error would be _thrown_ when there is a problem. Could I suggest changing the code like this?
```suggestion
  try {
    await ddb.put(param).promise()
  } catch(e) {
    console.log('[ROCKET#fileuploader] An error occurred while performing a PutItem operation: ', e)
  }
```
The non-callback style is the one we are using in Booster and it seems a bit more readable

alvaroloes (commented):
Left some comments. Again, thanks a lot for this rocket! It is simple yet very powerful!

packages/rocket-uploads3-store-event-aws-infrastructure/.npmignore
rdoria1: @alvaroloes I didn't get that one. 

So we have an embedded package.json for the lambda we are defining which is why we need to run an npm install on that subfolder as well (not only the main one for the rocket) 
I think this solution is fine. Thoughts?

lerna.json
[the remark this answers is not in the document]
rdoria1: @javiertoledo @alvaroloes  that makes sense I will rename everything then :)

packages/rocket-uploads3-store-event-aws-infrastructure/src/s3-notification-event-store.ts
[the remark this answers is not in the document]
rdoria1: hahaha I agree with you @NickSeagull 

I've mainly decided to use comments in this specific case for 2 reasons

- Since we have infra code I would expect this not too change (or very little) compared to business logic code
- CDK documentation is really crap in most cases so having comments would help people explore te code 

If you don't mind I will post this in Slack and I'd like to know what everyone thinks hehe.

packages/rocket-uploads3-store-event-aws-infrastructure/package.json
rdoria1: @NickSeagull @alvaroloes 
hahaha, I've published a LOT of local versions to test the rocket before publushing the PR :P 
So a question for you guys, Does it have to be the newest one? I thought Github action would update this to the new one? Not sure how it works

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
rdoria1: yep makes sense

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
rdoria1: @juanjoman @alvaroloes @NickSeagull 

- Believe it or not I could not find any reference what kind of type should I use for the S3 event to Lambda event 😢 

- All the references codes when receiving an S3 notification are not returning any value which I think makes sense. We don't control the response since it's automatically triggered from S3. We can't do anything with that response. 

- I can't return a Promise<unknown> the error is get is `A function whose declared type is neither 'void' nor 'any' must return a value.` Sorry still a noob in TS not sure what to do here.

packages/rocket-uploads3-store-event-aws-infrastructure/.npmignore
juanjoman: Oh I think @alvaroloes's question is if we would need this `.npmignore` in an external repo. I think we could avoid it if we specify on the `.gitignore` that we specifically don't want the root `node_modules` right? @rdoria1

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
juanjoman: Hmm I'm not sure about the input (event param) but for the return type, I think it would be `Promise<void>` if you don't return that await, and `Promise<DynamoDB.PutItemOutput>` 🤔

packages/rocket-fileuploader-aws-infrastructure/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+This package is a configurable Booster rocket to add an Object Storage(S3 bucket) to your Booster Application. 
+Every time a new object is uploaded, a new event will be persisted in the Events Store will data like the fileUri and file size.
+
+## Usage
+
+Install this package as a dependency in your Booster project.
+
+```sh
+npm install @boostercloud/rocket-static-sites-aws-infrastructure
```
juanjoman, line 11:
```suggestion
npm install --save-dev @boostercloud/rocket-static-sites-aws-infrastructure
```

packages/rocket-fileuploader-aws-infrastructure/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@boostercloud/rocket-fileuploader-aws-infrastructure",
+  "version": "0.9.5699",
```
juanjoman, line 3:
```suggestion
  "version": "0.11.2",
```

packages/rocket-fileuploader-aws-infrastructure/package.json
```diff
@@ -0,0 +1,56 @@
...
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/boostercloud/booster.git"
+  },
+  "dependencies": {
+    "@aws-cdk/aws-dynamodb": "1.67.0",
+    "@aws-cdk/aws-lambda": "1.67.0",
+    "@aws-cdk/aws-lambda-event-sources": "1.67.0",
+    "@boostercloud/framework-types": "^0.11.1",
```
juanjoman, line 29: I think we can also move this one to devDependencies:
```suggestion
    "@boostercloud/framework-types": "0.11.2",
```

packages/rocket-fileuploader-aws-infrastructure/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "postinstall": "npm --prefix src/lambdas install",
+    "clean": "rimraf ./dist tsconfig.tsbuildinfo",
+    "prepack": "tsc -b tsconfig.json",
+    "test": "nyc --extension .ts mocha --forbid-only \"test/**/*.test.ts\""
+  },
+  "bugs": {
+    "url": "https://github.com/boostercloud/booster/issues"
+  },
+  "devDependencies": {
+    "@boostercloud/framework-provider-aws-infrastructure": "^0.11.1",
```
juanjoman, line 47:
```suggestion
    "@boostercloud/framework-provider-aws-infrastructure": "0.11.2",
```

juanjoman (approved):
LGTM!! Just left some minor suggestions 🙏

packages/rocket-fileuploader-aws-infrastructure/README.md
rdoria1: Synced offline, we need it as dependency because of lambda code

packages/rocket-uploads3-store-event-aws-infrastructure/src/lambdas/index.ts
alvaroloes: Yep! What Juanjo said @rdoria1

packages/rocket-fileuploader-aws-infrastructure/README.md
```diff
@@ -0,0 +1,67 @@
+# File Uploader Booster Rocket for AWS
+
+This package is a configurable Booster rocket to add an Object Storage(S3 bucket) to your Booster Application. 
+Every time a new object is uploaded, a new event will be persisted in the Events Store will data like the fileUri and file size.
```
alvaroloes, line 4:
```suggestion
Every time a new object is uploaded, a new event will be persisted in the Events Store with data like the fileUri and file size.
```

packages/rocket-fileuploader-aws-infrastructure/README.md
alvaroloes: Could you please tell the rest why? From my point of view, the lambda code gets deployed on deployment time, so we should not need the rocket code in runtime anymore ❔

alvaroloes (approved):
Approved, just left two more comments

PR #542, opened 2021-01-13
Fix/509 missing rimraf dependency cli package

davidverdu:
## Description
**rimraf** is used in `yarn clean` script. If it's not installed globally, yarn fails. I included **rimraf** as a cli `package.json` dependency. I included it as a global binary command, so yarn execution doesn't fail.

## Changes
- Update `packages/cli/package.json` to add the dependency

## Checks
- [x] Project Builds
 
## Additional information
- I'm not sure if this is the best solution. Adding **rimraf** in the bin package.json property will install the command globally and it may collide with another software that installs the same dependency globally. Rimraf is likely to be commonly used in other projects. An alternative is creating our own delete script and require the **rimraf** and **glob** dependencies to perform the deletion without declaring **rimraf** as a global binary

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -8,7 +8,8 @@
     "access": "public"
   },
   "bin": {
-    "boost": "./bin/run"
+    "boost": "./bin/run",
+    "rimraf": "./node_modules/rimraf/bin.js"
   },
   "bugs": "https://github.com/boostercloud/booster/issues",
   "dependencies": {
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -27,6 +27,7 @@
     "inquirer": "^7.0.0",
     "mustache": "^3.0.1",
     "ora": "^3.4.0",
+    "rimraf": "^3.0.2",
     "tslib": "^1"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -9486,7 +9486,7 @@ rimraf@^2.5.4, rimraf@^2.6.2, rimraf@^2.6.3:
   dependencies:
     glob "^7.1.3"
 
-rimraf@^3.0.0:
+rimraf@^3.0.0, rimraf@^3.0.2:
   version "3.0.2"
   resolved "https://registry.yarnpkg.com/rimraf/-/rimraf-3.0.2.tgz#f1a5402ba6220ad52cc1282bac1ae3aa49fd061a"
   integrity sha512-JZkJMZkAGFFPP2YqXZXPbMlMBgsxzE8ILs4lMIX/2o0L9UBw9O/Y3o6wFw/i9YLapcUJWwqbi3kdxIPdC62TIA==


davidverdu:
Another alternative: Why don't we create the `boost clean`command?

packages/cli/package.json
```diff
@@ -8,7 +8,8 @@
     "access": "public"
   },
   "bin": {
-    "boost": "./bin/run"
+    "boost": "./bin/run",
+    "rimraf": "./node_modules/rimraf/bin.js"
```
NickSeagull, line 12: I don't think that this should be exposed to the user at all. Especially because they might have `rimraf` already installed. I don't see the issue, why are we exposing it like this?

davidverdu: @NickSeagull by just adding the dependency in package.json I didn't manage to make `yarn clean` and `boost deploy -e production` work (rimraf command is required to be a global command to make those commands work as they are now). I totally agree that adding rimraf to bin is smell code. I suggest what I mentioned in Additional information or adding in the documentation that rimraf should be installed by the user before using booster. Nonetheless, yarn is about to disappear and maybe we can handle this by running `npx rimraf ...`, which probably will be able to get the package correctly

javiertoledo (approved):
For me this is fine, but would it make sense to remove the dependency completely and do the scripts differently? Having to install a dependency to perform a `rm -rf` is not a bit overkill?

davidverdu:
@javiertoledo I would just remove this explicit call to rimraf from a npm script. Maybe we can solve this by creating `boost clean` command (which. could call rimraf and glob internally, without the need of having rimraf as a global dependency). What do you think?

alvaroloes (commented):
@davidverdu I have two questions/proposals:
1. We don't need to add `rimraf` globally so that it is available when we do something like `npm run clean`. We can have it in the project local dependencies and call that command with `npx npm run clean`. Anyway, I'm almost sure with the latest changes in master, this is not needed and it is already working.
2. the `rimraf` dependency should be added to the `devDependencies` section, as it is only used when developing the framework.
3. When you add `rimraf` as devDependency, it already creates a binary shortcut in the "node_modules/.bin" folder, so you don't need to add it manually in the package.json "bin" section. Check the image below:
image

Right now in master, I'm not finding any problem with "rimraf", so it is likely that if you rebase, this problem is gone and nothing needs to be done 😃

davidverdu:
OK @alvaroloes thanks for your review. I'll check it and if it's not necessary, I'll close the PR without merging

NickSeagull:
@javiertoledo `rimraf` ensures cross platform compatibility. `rm -rf` doesn't work on Windows

javiertoledo:
> @javiertoledo `rimraf` ensures cross platform compatibility. `rm -rf` doesn't work on Windows

Somehow I knew you would answer that :trollface:

alvaroloes:
This is already fixed by https://github.com/boostercloud/booster/pull/576 using `npx` ;-)

davidverdu:
Ok @alvaroloes I close the PR then

PR #545, opened 2021-01-13
Change to NPM

alvaroloes:
## Description
This PR makes the necessary changes to use only lerna + npm and initiate the moving away from yarn. 
Deployments made on this branch using the `integration/deploy` scrips results in lambdas of only **2,8MB instead of 70MB** 

Please! follow these steps when checking out this branch:

_Things you must ensure_
- Clean all compiled code (`lerna run clean`)
- Delete all "node_modules" folders (`lerna clean -y`)
- Delete the "node_modules" folder in the root of your project
- Delete all “.lock” files (yarn.lock, npm.lock, package-lock.json) from all packages
- Verify you have installed the **npm version 6 (any subversion)**

_Extra steps_
- Install “lerna” globally `npm install -g lerna`
- `lerna bootstrap`
- `lerna run compile`

From now on, you can keep working as usual

_Advice that should be hardcoded in our brains_

Always use fixed versions in your "package.json" file. For example, use “1.2.3” instead of “^1.2.3”. The reason is that libraries, although they try to adhere to the semantic versioning and not to break compatibility with minor version updates, this is not always true, especially in Typescript. An example was the `aws-sdk` that added some new methods to an interface and the compilation was broken because the “thing” we were assigning to a variable of that interface didn’t implement those methods.

This is done automatically if you use "lerna" to add the dependencies. For example `lerna add uuid packages/framework-core`. That command will install uuid to the "framework-core" package using the latest version and writing the exact version in the corresponding package.json file.

## Checks
- [x] Project Builds
 
## Additional information

There are three other PRs that depend on this change and are waiting for this PR to become stable to fix their tests.
* #547 -> This fixes an issue that prevented framework packages under development to be deployed during integration tests. If we run into this kind of issues while fixing this PR integration tests, it could make sense to merge #547 with #545 before trying to solve this again in this branch.
* #546 -> Minor changes after a previous spike trying to make npm work. We might need the changes in the tests made in #546, so it could make sense to merge it with this branch.
* #524 -> This one doesn't really depend on this branch, but chances are that we need npm to work before being able to finish #524, so we've rebase it to this PR.


javiertoledo (approved):
Looking awesome already!

packages/framework-provider-azure-infrastructure/package.json
```diff
@@ -32,7 +32,8 @@
     "copyfiles": "^2.3.0",
     "ms-rest-azure": "^3.0.0",
     "mustache": "^4.0.1",
-    "needle": "^2.5.0"
+    "needle": "^2.5.0",
+    "uuid": "^8.3.2"
```
juanjoman, line 36: What's the difference between this `uuid` and `@types/uuid` from devDependencies below?

`"uuid": "^8.3.2"` vs `"@types/uuid": "^8.3.0",`

Also mustache:
`"mustache": "^4.0.1"` vs `"@types/mustache": "^4.1.0"`

juanjoman (approved):
LGTM! Just left a comment below and also have a question: shouldn't we re-upload the `-lock.json` files to the repo but generated with npm?

adriangmweb (approved):
Should you change as well the yarn add link references in this file?
https://github.com/boostercloud/booster/blob/b7a082b409ac61120b2fb334500053c39709c7da/packages/framework-integration-tests/integration/helper/depsHelper.ts#L15

Regarding the npm fixed versions, if it still an issue in the future we could add a `.npmrc` file to the project with the `save-exact=true` flag on it. 

Apart from that LGTM! Thanks for this 🥇

adriangmweb:
Should you change as well the `yarn add link` references in this file? https://github.com/boostercloud/booster/blob/b7a082b409ac61120b2fb334500053c39709c7da/packages/framework-integration-tests/integration/helper/depsHelper.ts#L15

packages/framework-provider-aws/package.json
```diff
@@ -38,7 +38,7 @@
     "@types/aws-lambda": "^8.10.48",
     "@types/faker": "^4.1.12",
     "@types/get-installed-path": "^4.0.0",
-    "aws-sdk": "^2.771.0",
+    "aws-sdk": "2.771.0",
```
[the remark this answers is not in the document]
alvaroloes: This is something that we should take into account much more closely from now on: use strict versions, and, as you say, use the same version everywhere.

I think we were using the same version in both packages, so I will change this. Thanks!

Maybe there is a way to parameterize the `package.json` files so that we can use variables for common version numbers...

packages/framework-provider-azure-infrastructure/package.json
alvaroloes: "uuid" is the runtime dependency. It is the package itself. However, it is a package done in Javascript, not in Typescript, so when compiling we get errors because Typescript don't know anything about the types of the things defined in "uuid". The solution for that is to provide an extra package with all the types definitions (just the definitions, they go in a separate .d.ts file). And those packages are normally under the @types scope.

So when using a package in a Typescript project that was developed using Javascript, you need to also include the type definition package, which in this case is `@types/uuid` (the same with mustache)

**Regarding your question about `-lock.json` files**

 I think you are correct. I will upload them after fixing the tests ;-)

alvaroloes:
@adriangmweb 
- I haven't changed those `yarn link` commands because @javiertoledo has fixed that differently (or already used npm) in his PRs
- What does `--save-exact` mean?

adriangmweb:
> @adriangmweb
> 
> * I haven't changed those `yarn link` commands because @javiertoledo has fixed that differently (or already used npm) in his PRs
> * What does `--save-exact` mean?

That flag makes npm save the exact version in the `package.json`, so instead of `^2.1.1` it saves the dependency as `2.1.1` automatically when performing a `npm install`

more info: https://dev.to/daleljefferson/how-to-save-exact-npm-package-versions-10fm

alvaroloes:
Hey! that's amazing! I'll do that, thanks!

NickSeagull (approved):
Lovely! Remember to change the documentation as well

davidverdu (approved):

PR #547, opened 2021-01-13
Fix: deploy framework packages under development during integration tests

javiertoledo:
## Description

We've been having issues with local and aws integration tests because the published versions were used or deployed instead of the versions under development. This should unblock some PRs that are blocked because integration tests don't pass (as they're using old versions.

> These changes were rescued from a #528, which is now closed, as switching to npm miraculously fixed the "prune devDependencies" process 🎉 

## Changes

* Replaced calls to `yarn add link:...` by `npm install file:...` to install the actual packages in the project's `node_modules` instead of just symlinking them
* Renamed the function that 

## Checks
- [ ] Project Builds
 
## TODO

This change depends on #545, so we'll need to fix the tests in that branch before finishing and merging this code.
- [ ] Fix unit and integration tests

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -22,7 +22,10 @@
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen",
       "protocol": "inspector",
-      "runtimeArgs": ["--preserve-symlinks"]
+      "runtimeArgs": ["--preserve-symlinks"],
+      "env": {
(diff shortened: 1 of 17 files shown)


alvaroloes:
I think this is already fixed in version `0.10.0` right?

javiertoledo:
> I think this is already fixed in version `0.10.0` right?

Yes, closing this one

PR #555, opened 2021-01-15
541 refactor GraphQLGenerator class

### LINKED ISSUE #541: Consider refactoring `GraphQLGenerator` class to make sure that it's used consistently
## Description

The `GraphQLGenerator` class is used to build the GraphQL schema from the Booster configuration. It's currently used from two different places using slightly different ways to build it:

_packages/framework-core/src/booster-graphql-dispatcher.ts_
```typescript
this.graphQLSchema = new GraphQLGenerator(config, commandDispatcher, this.readModelDispatcher).generateSchema()
```

_packages/framework-core/src/booster-subscribers-notifier.ts_
```typescript
this.graphQLSchema = new GraphQLGenerator(
      config,
      new BoosterCommandDispatcher(config, logger),
      new BoosterReadModelDispatcher(config, logger)
    ).generateSchema()
```

This could lead to future issues if we decide to change one of the usages but forget about the other. We want to make sure that the schema is generated in a 100% consistent way.

## Possible Solution

One possible way to solve this could be having a builder function somewhere or transform this to a singleton, but the goal of this task would be to study the code more deeply to come up with a good solution.

davidverdu:
## Description
Refactor GraphQLGenerator class to access it in a uniform way. Applying Singleton pattern. Since config: BoosterConfig is declared as static in Booster class, all classes that access config access the same instance of BoosterConfig. Taking that into account, the design has been simplified.

Fixes #541 

## Changes
- I made this class initialized via a Singleton getInstance() method
- Client classes get the GraphQLGenerator singleton instance to use it
- Update unit tests

image

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-core/src/booster-graphql-dispatcher.ts
+++ b/packages/framework-core/src/booster-graphql-dispatcher.ts
@@ -26,7 +26,7 @@ export class BoosterGraphQLDispatcher {
 
   public constructor(private config: BoosterConfig, private logger: Logger) {
     this.readModelDispatcher = new BoosterReadModelDispatcher(config, logger)
-    this.graphQLSchema = GraphQLGenerator.getInstance(config, logger).generateSchema()
+    this.graphQLSchema = GraphQLGenerator.build(config, logger).generateSchema()
     this.boosterTokenVerifier = new BoosterTokenVerifier(config)
     this.websocketHandler = new GraphQLWebsocketHandler(
       config,
--- a/packages/framework-core/src/booster-graphql-dispatcher.ts
+++ b/packages/framework-core/src/booster-graphql-dispatcher.ts
@@ -10,7 +10,6 @@ import {
 import { GraphQLSchema, DocumentNode, ExecutionResult, GraphQLError } from 'graphql'
 import * as graphql from 'graphql'
 import { GraphQLGenerator } from './services/graphql/graphql-generator'
-import { BoosterCommandDispatcher } from './booster-command-dispatcher'
 import { BoosterReadModelDispatcher } from './booster-read-model-dispatcher'
 import { GraphQLResolverContext, graphQLWebsocketSubprotocolHeaders } from './services/graphql/common'
 import { NoopReadModelPubSub } from './services/pub-sub/noop-read-model-pub-sub'
@@ -27,9 +26,7 @@ export class BoosterGraphQLDispatcher {
 
   public constructor(private config: BoosterConfig, private logger: Logger) {
     this.readModelDispatcher = new BoosterReadModelDispatcher(config, logger)
-    const commandDispatcher = new BoosterCommandDispatcher(config, logger)
-
-    this.graphQLSchema = new GraphQLGenerator(config, commandDispatcher, this.readModelDispatcher).generateSchema()
+    this.graphQLSchema = GraphQLGenerator.getInstance(config, logger).generateSchema()
     this.boosterTokenVerifier = new BoosterTokenVerifier(config)
     this.websocketHandler = new GraphQLWebsocketHandler(
       config,
--- a/packages/framework-core/src/services/graphql/graphql-generator.ts
+++ b/packages/framework-core/src/services/graphql/graphql-generator.ts
@@ -21,17 +21,14 @@ export class GraphQLGenerator {
   private readonly subscriptionGenerator: GraphQLSubscriptionGenerator
   private readonly typeInformer: GraphQLTypeInformer
 
-  private static singleton: GraphQLGenerator | undefined = undefined
+  private static singleton: GraphQLGenerator | undefined
   
   public static getInstance(config: BoosterConfig, logger: Logger): GraphQLGenerator {
-    if (this.singleton === undefined) {
-      this.singleton = new GraphQLGenerator(
-        config,
-        new BoosterCommandDispatcher(config, logger),
-        new BoosterReadModelDispatcher(config, logger)
-      )
-    }
-    return this.singleton;
+    return this.singleton ?? new GraphQLGenerator(
+      config,
+      new BoosterCommandDispatcher(config, logger),
+      new BoosterReadModelDispatcher(config, logger)
+    )
   }
 
   private constructor(
--- a/packages/framework-core/src/services/graphql/graphql-generator.ts
+++ b/packages/framework-core/src/services/graphql/graphql-generator.ts
@@ -34,7 +34,7 @@ export class GraphQLGenerator {
     return this.singleton;
   }
 
-  public constructor(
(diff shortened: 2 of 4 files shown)


NickSeagull:
Remember to add the text "Fixes #541" to your Pull Request description so GitHub can close the issue automatically. You can apply this for any issue number 😄

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -20,7 +21,20 @@ export class GraphQLGenerator {
...
+  public static getInstance(config: BoosterConfig, logger: Logger): GraphQLGenerator {
+    if (this.singleton === undefined) {
+      this.singleton = new GraphQLGenerator(
+        config,
+        new BoosterCommandDispatcher(config, logger),
+        new BoosterReadModelDispatcher(config, logger)
+      )
+    }
+    return this.singleton;
+  }
```
NickSeagull, line 35: A simplification:

* Nullable fields are initializied to `undefined` by default
* You can use the `??` operator to return another value in case the left hand side operator of it is null or undefined

```suggestion
  private static singleton: GraphQLGenerator | undefined
  
  public static getInstance(config: BoosterConfig, logger: Logger): GraphQLGenerator {
    return this.singleton ?? new GraphQLGenerator(
      config,
      new BoosterCommandDispatcher(config, logger),
      new BoosterReadModelDispatcher(config, logger)
    )
  }
```

NickSeagull (changes requested):
Thanks for this! Just a small suggestion

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -20,7 +21,18 @@ export class GraphQLGenerator {
   private readonly subscriptionGenerator: GraphQLSubscriptionGenerator
   private readonly typeInformer: GraphQLTypeInformer
 
-  public constructor(
+  private static singleton: GraphQLGenerator | undefined
+  
+  public static getInstance(config: BoosterConfig, logger: Logger): GraphQLGenerator {
+    this.singleton = this.singleton ?? new GraphQLGenerator(
```
davidverdu, line 27: @NickSeagull I had to update your suggestion because it was always creating a new instance of the class. With this change, Singleton behavior is working now

NickSeagull: Lovely, thanks!!

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -20,7 +21,18 @@ export class GraphQLGenerator {
   private readonly subscriptionGenerator: GraphQLSubscriptionGenerator
   private readonly typeInformer: GraphQLTypeInformer
 
-  public constructor(
+  private static singleton: GraphQLGenerator | undefined
+  
+  public static getInstance(config: BoosterConfig, logger: Logger): GraphQLGenerator {
```
javiertoledo, line 26: This is great, the only thing I'd change is that I think we're using the word `build` instead of `getInstance` in other singletons in the codebase, so I'd suggest to rename it. What do other @boostercloud/booster-core members think about this?

```suggestion
  public static build(config: BoosterConfig, logger: Logger): GraphQLGenerator {
```

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -27,9 +26,7 @@ export class BoosterGraphQLDispatcher {
 
   public constructor(private config: BoosterConfig, private logger: Logger) {
     this.readModelDispatcher = new BoosterReadModelDispatcher(config, logger)
```
javiertoledo, line 28: 🤔 I see that a similar thing is happening here with the `readModelDispatcher` and later with the `commandDispatcher`. Now that we're not using it to initialize the `GraphQLGenerator`, we're now using two separate instances that are initialized with the same values, would it make sense to make them a singleton class too?

packages/framework-core/test/services/graphql/graphql-generator.test.ts
```diff
@@ -1,38 +1,53 @@
...
+    it('should create the instance', () => {
+      const instance = GraphQLGenerator.getInstance(mockConfig, mockLogger)
+
+      expect(instance).to.not.be.undefined
+      expect(instance).to.be.an.instanceOf(GraphQLGenerator)
+    })
+
+    it('should return always the same instance', () => {
+      const instance1 = GraphQLGenerator.getInstance(mockConfig, mockLogger)
+      const instance2 = GraphQLGenerator.getInstance(mockConfig, mockLogger)
```
javiertoledo, line 41: Given that neither the config or the logger ever change, maybe we should explore a way to make these params disappear. I think that the config should definitely become a singleton, and the logger can be a config option (not in this PR anyway).

One nice advantage of this is that we could allow the user to set an alternate logger class in their Booster config to easily support logging to external platforms like Splunk or Papertrail...

javiertoledo (approved):
Very nice change! Make sure to rebase it to master and see all integration tests passing before merging in addition to the unit tests checks.

packages/framework-core/src/booster-graphql-dispatcher.ts
davidverdu: I think we can make another PR to refactor more classes into Singleton. I'll create the issue in GitHub

packages/framework-core/test/services/graphql/graphql-generator.test.ts
davidverdu: You are right, we can add more refactoring here. I'll create the issue for future improvements

packages/framework-core/src/services/graphql/graphql-generator.ts
NickSeagull: Agree

packages/framework-core/test/services/graphql/graphql-generator.test.ts
NickSeagull: @javiertoledo what you describe is addressed with the [Reader pattern](https://dev.to/riccardo_cardin/and-monads-for-almost-all-the-reader-monad-1ife), sadly TypeScript syntax doesn't support working comfortably with it. I'll _eventually_ work on a fix...

davidverdu: I have created an issue for this future refactoring https://github.com/boostercloud/booster/issues/566

davidverdu:
I have refactored getInstance() into buid() method and rebased with master. Could you review the PR?

NickSeagull (approved):
Thank you a lot! 🙏

PR #556, opened 2021-01-15
Feature: Rocket Cognito Auth

charlietfe:
## Description
This PR contains a Booster Rocket to create an Authentication API using AWS Cognito Service. The idea behind this change is the deprecation of the current auth implementation inside Booster core.

Initially, we are offering the same features we had in the Booster core but with 2 operational modes:

  * **One time password/Passwordless** to avoid using passwords and use **only phone number** as usernames. This mode will send use SMS as a notification mode during the authentication/authorization process. To use this mode the following configuration must be specified in the `config.ts` file inside the user project: 

```typescript
  config.appName = 'my-shop-demo'
  config.provider = Provider([
    {
      packageName: '@boostercloud/rocket-auth-aws-infrastructure',
      parameters: {
        mode: 'Passwordless',
      },
    },
  ])
})
```

  * **Username/Password** using **email or phone (depending on the booster role definition in the user app)** as username and password in a classic style. This mode allows the users to set password policies. With this mode the password policies could be configured as well:

```typescript
  config.appName = 'my-shop-demo'
  config.provider = Provider([
    {
      packageName: '@boostercloud/rocket-auth-aws-infrastructure',
      parameters: {
        mode: 'UserPassword',
        passwordPolicy: {
          minLenght: 8,
          requireDigits: true,
          requireSymbols: false,
          requireUppercase: true,
          requireLowercase: true,
        }
      },
    },
  ])
})
```

The rocket will expose the following endpoints:

 `POST /sign-in`

 `POST /sign-up`
 `POST /sign-up/confirm`
 `POST /sign-up/resend-code`

 `POST /token` **(passwordless)**
 `POST /token/refresh`
 `POST /token/revoke`

 `POST /password/forgot` **(user/password)**
 `POST /password/change` **(user/password)**

## Changes

All the above endpoints will hit the corresponding lambdas which contain the specific logic for each Cognito use case. All the lambdas are located in the [lambdas](https://github.com/boostercloud/booster/tree/feature%2Frocket-cognito-auth/packages/rocket-auth-aws-infrastructure/src/lambdas) directory

In addition to that, to implement the `Passwordless` mode I had to study how the custom Cognito auth flow works, because this flow involves an internal protocol between three lambdas called `challenges` as described [here](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-challenge.html). Here you can see a diagram of how this protocol works:

lambda-challenges1

The Passwordless mode only works with phone numbers as usernames, which is a common pattern in Mobile Development. Taking that into account, all the interaction between the users and the API to confirm accounts and get authorizations will be done using SMS messages **which may increase the cost of using the API in this way.**

## Checks
- [X] Project Builds
 
## Notes

In the future, we'll add more features like registering events inside Booster on each auth action and we'll allow the user to configure more Cognito options like third party providers, confirmation messages, and so on.


packages/rocket-auth-aws-infrastructure/src/lambdas/challenge-answer.ts
```diff
@@ -0,0 +1,31 @@
+import { CognitoIdentityServiceProvider } from 'aws-sdk'
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda'
+import { errorResponse, okResponse } from './response'
+
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  try {
+    const params = JSON.parse(event.body!)
```
juanjoman, line 7: Just curious, what's this '!' at the end of `event.body`??

packages/rocket-auth-aws-infrastructure/src/lambdas/response.ts
```diff
@@ -0,0 +1,23 @@
...
+export const errorResponse = (e: AWSError): APIGatewayProxyResult => {
+  return response(httpStatusCodeFor(e), {
+    error: {
+      type: e.code,
+      message: e.message,
+    },
+  })
+}
+
+export const okResponse = (data: object): APIGatewayProxyResult => {
```
juanjoman, line 21: Is this function needed? Can't we call to `response(200, data)` directly?

packages/rocket-auth-aws-infrastructure/src/lambdas/sign-up.ts
```diff
@@ -0,0 +1,70 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { CognitoIdentityServiceProvider } from 'aws-sdk'
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda'
+import { errorResponse, okResponse } from './response'
+import { InvalidParameterError } from '@boostercloud/framework-types'
+import * as validator from 'validator'
+
+const validateParams = (params: any): void => {
```
juanjoman, line 8: Can we set this type?

juanjoman (changes requested):
Wow, amazing rocket!! 💯 . Just left a couple of suggestions below 😛

packages/rocket-auth-aws-infrastructure/README.md
```diff
@@ -0,0 +1,382 @@
+# Authentication Booster Rocket for AWS
+
+This package is a configurable Booster rocket to add an authentication api based on Cognito to your Booster applications.
```
NickSeagull, line 3:
```suggestion
This package is a configurable Booster rocket to add an authentication API based on Cognito to your Booster applications.
```

packages/rocket-auth-aws-infrastructure/README.md
```diff
@@ -0,0 +1,382 @@
...
+
+## Usage
+
+Install this package as a dev dependency in your Booster project (It's a dev dependency because it's only used during deployment, but we don't want this code to be uploaded to the project lambdas)
```
NickSeagull, line 7:
```suggestion
Install this package as a `devDependency` in your Booster project (It's a `devDependency` because it's only used during deployment, so we don't want this code to be uploaded to the project lambdas)
```

packages/rocket-auth-aws-infrastructure/README.md
```diff
@@ -0,0 +1,382 @@
...
+## Configuration Options
+
+```typescript
+{
+  config: BoosterConfig                      // Required
+  passwordPolicy?: {                         // Optional, all values are set to true by default.
+    minLength?: number                       // Minimum length, which must be at least 6 characters but fewer than 99 character
+    requireDigits: boolean                   // Require numbers
+    requireLowercase: boolean                // Require lowercase letters
+    requireSymbols: boolean                  // Require a special characters
```
NickSeagull, line 44:
```suggestion
    requireSymbols: boolean                  // Require special characters
```

packages/rocket-auth-aws-infrastructure/README.md
```diff
@@ -0,0 +1,382 @@
...
+    requireUppercase: boolean                // Require uppercase letters
+  }
+  mode: 'Passwordless' | 'UserPassword'      // One time password the user must be a phone number | User Password mode, the user must be a valid email.
```
NickSeagull, line 47:
```suggestion
  mode: 'Passwordless' | 'UserPassword'      // If Passwordless, the user must be a phone number. If UserPassword, the user must be a valid email.
```

packages/rocket-auth-aws-infrastructure/README.md
```diff
@@ -0,0 +1,382 @@
...
+  "error": {
+    "type": "CodeMismatchException",
+    "message": "Invalid verification code provided, please try again."
+  }
+}
+```
+
+### Resend Sign-up confirmation code
+
+If for some reason the confirmationCode never reach your email or your phone via SMS you can ask the api to resend a new one.
```
NickSeagull, line 177:
```suggestion
If for some reason the confirmation code never reaches your email, or your phone via SMS, you can ask the API to resend a new one.
```

packages/rocket-auth-aws-infrastructure/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@boostercloud/rocket-auth-aws-infrastructure",
+  "version": "0.9.4",
```
NickSeagull, line 3: Note that we're on `0.10.0` now

packages/rocket-auth-aws-infrastructure/src/auth-stack.ts
```diff
@@ -0,0 +1,336 @@
...
+      allowHeaders: ['*'],
+      allowOrigins: Cors.ALL_ORIGINS,
+      allowMethods: ['POST', 'OPTIONS'],
+    }
+
+    // sign-in
+    const signInResource = authResource.addResource('sign-in', { defaultCorsPreflightOptions })
+    AuthStack.addIntegration('sign-in', stack, params, userPool, userPoolClientId, signInResource, 'sign-in.handler', [
+      'cognito-idp:InitiateAuth',
+    ])
```
NickSeagull, line 160: Maybe this can be a function called `createSignInResource` ?

packages/rocket-auth-aws-infrastructure/src/auth-stack.ts
```diff
@@ -0,0 +1,336 @@
...
+    AuthStack.addIntegration(
+      'sign-up-resend-code',
+      stack,
+      params,
+      userPool,
+      userPoolClientId,
+      resource,
+      'resend-confirmation-code.handler',
+      ['cognito-idp:ResendConfirmationCode']
+    )
```
NickSeagull, line 201: Maybe this can be a function called `createSignUpResources` ?

packages/rocket-auth-aws-infrastructure/src/auth-stack.ts
```diff
@@ -0,0 +1,336 @@
...
+      userPoolClientId,
+      resource,
+      'refresh-token.handler',
+      ['cognito-idp:InitiateAuth']
+    )
+
+    resource = tokenResource.addResource('revoke', { defaultCorsPreflightOptions })
+    AuthStack.addIntegration('revoke-token', stack, params, userPool, userPoolClientId, resource, 'sign-out.handler', [
+      'cognito-idp:GlobalSignOut',
+    ])
```
NickSeagull, line 236: Maybe this can be a function called `createTokenResources` ?

packages/rocket-auth-aws-infrastructure/src/auth-stack.ts
```diff
@@ -0,0 +1,336 @@
...
+        'change-password',
+        stack,
+        params,
+        userPool,
+        userPoolClientId,
+        resource,
+        'confirm-forgot-password.handler',
+        ['cognito-idp:ConfirmForgotPassword']
+      )
+    }
```
NickSeagull, line 265: Maybe this can be a function called `createTokenResources` ?

packages/rocket-auth-aws-infrastructure/src/auth-stack.ts
```diff
@@ -0,0 +1,336 @@
...
+          responseParameters: allowedOriginHeaderForCors,
+        },
+        {
+          statusCode: '500',
+          responseParameters: allowedOriginHeaderForCors,
+        },
+      ],
+    }
+
+    const authLambda = createLamba(stack, `${AuthStack.rocketArtifactsPrefix(params)}-${name}`, handler, {
```
NickSeagull, line 300: Typo
```suggestion
    const authLambda = createLambda(stack, `${AuthStack.rocketArtifactsPrefix(params)}-${name}`, handler, {
```

packages/rocket-auth-aws-infrastructure/src/auth-stack.ts
```diff
@@ -0,0 +1,336 @@
...
+        actions: actions,
+        resources: [userPool.userPoolArn],
+      })
+    )
+  }
+
+  private static printOutput(stack: Stack, userPoolId: string, authApi: RestApi): void {
+    new CfnOutput(stack, 'AuthApiEndpoint', {
+      value: authApi.url + 'auth',
+      description: 'Auth api endpoint',
```
NickSeagull, line 321:
```suggestion
      description: 'Auth API endpoint',
```

packages/rocket-auth-aws-infrastructure/src/auth-stack.ts
```diff
@@ -0,0 +1,336 @@
...
+    })
+
+    const issuer = `https://cognito-idp.${stack.region}.${stack.urlSuffix}/${userPoolId}`
+
+    new CfnOutput(stack, 'AuthApiIssuer', {
+      value: issuer,
+      description: 'Auth api JWT issuer',
```
NickSeagull, line 328:
```suggestion
      description: 'Auth API JWT issuer',
```

packages/rocket-auth-aws-infrastructure/src/auth-stack.ts
```diff
@@ -0,0 +1,336 @@
...
+    })
+
+    new CfnOutput(stack, 'AuthApiJwksUri', {
+      value: issuer + '/.well-known/jwks.json',
+      description: 'Auth api JKWS uri',
```
NickSeagull, line 333:
```suggestion
      description: 'Auth API JKWS URI',
```

packages/rocket-auth-aws-infrastructure/src/lambdas/challenge-answer.ts
NickSeagull: It's to "force" it

Here, `event.body` is nullable, so TS will complain about the fact that it is nullable and you need to write an `if` to make sure that it exists and react otherwise. If you write `event.body!`, you are telling TS to forget about the nullable part, so you "force" it into non-nullable.

The drawback of using that operator, and why I dislike its usage, is that it will `throw` if it is indeed `undefined`, but here it is okay to use it, because @charlietfe wrapped everything into a `try`/`catch`

packages/rocket-auth-aws-infrastructure/src/lambdas/challenge-create.ts
```diff
@@ -0,0 +1,29 @@
+import { CognitoUserPoolEvent } from 'aws-lambda'
+import { SNS } from 'aws-sdk'
+
+const randomSecretCode = (min = 100000, max = 999999): string => {
+  return Math.floor(Math.random() * (max - min + 1) + min).toString()
+}
```
NickSeagull, line 6: Can you add a doc comment to this please?

packages/rocket-auth-aws-infrastructure/src/lambdas/challenge-define.ts
```diff
@@ -0,0 +1,32 @@
...
+    // The user provided the right answer; succeed auth
+    event.response.issueTokens = true
+    event.response.failAuthentication = false
+  } else {
+    // The user did not provide a correct answer yet; present challenge
+    event.response.issueTokens = false
+    event.response.failAuthentication = false
+    event.response.challengeName = 'CUSTOM_CHALLENGE'
+  }
+  return event
```
NickSeagull, line 31: Maybe extracting into functions?

```suggestion
  if (onlyAcceptingCustomChallenges(event)) {
    return failAuth(event)
  } else if (userProvidedWrongAnswerThreeTimes(event)) {
    return failAuth(event)
  } else if (userProvidedRightAnswer(event)) {
    return succeedAuth(event)
  } else { // The user did not provide a correct answer yet
    return presentChallenge(event)
  }
  return event
```

packages/rocket-auth-aws-infrastructure/src/lambdas/challenge-verify.ts
```diff
@@ -0,0 +1,11 @@
+import { CognitoUserPoolEvent } from 'aws-lambda'
+
+export const handler = async (event: CognitoUserPoolEvent): Promise<CognitoUserPoolEvent> => {
+  const expectedAnswer = event.request.privateChallengeParameters?.secretLoginCode ?? ''
+  if (event.request.challengeAnswer === expectedAnswer) {
+    event.response.answerCorrect = true
+  } else {
+    event.response.answerCorrect = false
+  }
```
NickSeagull, line 9: Why not
```suggestion
  const { challengeAnswer } = event.request
  event.response.answerCorrect = challengeAnswer === expectedAnswer
```

packages/rocket-auth-aws-infrastructure/src/lambdas/response.ts
NickSeagull: Its to give the same abstraction level

Imagine some code that goes like:

```ts
if (something) {
  errorResponse(...)
} else {
  response(200, ...)
}
```

It's okay, but the `else` branch is slightly low level than the `if` branch. So its good to hide the code implementation detail in order to make it slightly more abstract:

```ts
if (something) {
  errorResponse(...)
} else {
  okResponse(...)
}
```

packages/rocket-auth-aws-infrastructure/src/lambdas/sign-in.ts
```diff
@@ -0,0 +1,37 @@
...
+    return okResponse({
+      accessToken: initAuthResponse.AuthenticationResult?.AccessToken,
+      idToken: initAuthResponse.AuthenticationResult?.IdToken,
+      expiresIn: initAuthResponse.AuthenticationResult?.ExpiresIn,
+      refreshToken: initAuthResponse.AuthenticationResult?.RefreshToken,
+      tokenType: initAuthResponse.AuthenticationResult?.TokenType,
+    })
+  } catch (e) {
+    return errorResponse(e)
+  }
```
NickSeagull, line 36: I see that many of these lambdas are nearly equal, perhaps you could create a wrapper higher order function in order to remove duplication?

packages/rocket-auth-aws-infrastructure/src/lambdas/sign-up.ts
NickSeagull: Yes please 🙏

NickSeagull (changes requested):
Thank you a lot! 🙏 Left some suggestions

packages/rocket-auth-aws-infrastructure/src/lambdas/sign-in.ts
charlietfe: Good idea @NickSeagull, I'll review all the code again and I will apply some suggestions. I'll let you know.

javiertoledo (commented):
I left a couple of comments to discuss

packages/rocket-auth-aws-infrastructure/src/index.ts
```diff
@@ -0,0 +1,9 @@
+import { InfrastructureRocket } from '@boostercloud/framework-provider-aws-infrastructure'
+import { AuthStack, AWSAuthRocketParams } from './auth-stack'
+
+const AWSAuth = (params: AWSAuthRocketParams): InfrastructureRocket => ({
+  mountStack: AuthStack.mountStack.bind(null, params),
+  unmountStack: AuthStack.unmountStack?.bind(null, params),
+})
```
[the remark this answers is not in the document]
charlietfe: @javiertoledo I missed the booster config, the rest is fine at least for this use case. Anyway, let's gather feedback from other rockets to create a better interface if needed.

packages/rocket-auth-aws-infrastructure/src/lambdas/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "validator": "^13.1.1",
+    "@boostercloud/framework-types": "0.9.4",
+    "tslib": "2.0.1"
+  },
+  "bundledDependencies": [
+    "validator",
+    "@boostercloud/framework-types",
+    "tslib"
+  ]
+}
```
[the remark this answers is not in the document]
charlietfe: @javiertoledo Totally agreed, but for some reason, I had to upload the `dist` folder with all lambdas dependencies in order to work. I didn't find a better way but definitely, it's something we need to discuss because this was one of the pain points I found creating this rocket.

packages/rocket-auth-aws-infrastructure/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "test": "nyc --extension .ts mocha --forbid-only \"test/**/*.test.ts\""
+  },
+  "bugs": {
+    "url": "https://github.com/boostercloud/booster/issues"
+  },
+  "devDependencies": {
+    "@types/aws-lambda": "8.10.48",
+    "aws-sdk": "2.764.0",
+    "@boostercloud/framework-provider-aws-infrastructure": "^0.9.4",
+    "@boostercloud/framework-types": "^0.9.4",
```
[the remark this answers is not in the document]
charlietfe: Fixed [here](https://github.com/boostercloud/booster/pull/556/commits/dfc0c4014bcdfd27086220b92a91a30fe6acd1af)

packages/rocket-auth-aws-infrastructure/src/lambdas/challenge-define.ts
charlietfe: Fixed [here](https://github.com/boostercloud/booster/pull/556/commits/dfc0c4014bcdfd27086220b92a91a30fe6acd1af)

packages/rocket-auth-aws-infrastructure/src/lambdas/challenge-verify.ts
charlietfe: Fixed [here](https://github.com/boostercloud/booster/pull/556/commits/dfc0c4014bcdfd27086220b92a91a30fe6acd1af)

packages/rocket-auth-aws-infrastructure/package.json
charlietfe: Fixed [here](https://github.com/boostercloud/booster/pull/556/commits/dfc0c4014bcdfd27086220b92a91a30fe6acd1af)

packages/rocket-auth-aws-infrastructure/src/index.ts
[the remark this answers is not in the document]
javiertoledo: I think that someone made the change to make it available, it must be in some of the open PRs

packages/rocket-auth-aws-infrastructure/src/lambdas/package.json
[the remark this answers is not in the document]
javiertoledo: One idea we discussed is that rockets like this are provided in two (or more) separate packages (See the [naming recommendation section in the docs](https://github.com/boostercloud/booster/tree/master/docs#naming-recommendations)):

In this case:

* `rocket-auth-aws-infrastructure`: Provides all the code that is imported dynamically from the CLI to provision and deploy the infrastructure.
* `rocket-auth-aws`: Provides extensions to the Booster app, so you could provide the lambdas as functions. The only problem is that you'd need to re-export them in the App index doing something like this:

```typescript
// index.ts from a Booster application
import { Booster } from '@boostercloud/framework-core'
export {
  Booster,
  boosterEventDispatcher,
  boosterPreSignUpChecker,
  boosterServeGraphQL,
  boosterNotifySubscribers,
  boosterTriggerScheduledCommand,
} from '@boostercloud/framework-core'

// Re-export the functions provided by the rocket
export * from '@boostercloud/rocket-auth-aws'

Booster.start(__dirname)
```

Making the end user re-export the lambdas provided by the rockets is a bit messy, I wonder id we could do something more "standarized", like having a lambda registry object or something like that. We're already exporting the `Booster` object, which is a proxy for the `BoosterConfig` object. If we register all lambdas in a new `lambda` property, I think we could change the Booster template to export only the `Booster` object like this:

```typescript
import { Booster } from '@boostercloud/framework-core'
export { Booster } from '@boostercloud/framework-core'

Booster.start(__dirname)
```

And then from the infrastructure packages, referencing the functions as `Booster.lambda.boosterServeGraphQL` instead of just `boosterServeGraphQL`. The advantage of this, if it works, is that rockets could register any number of lambdas they want in an standardized way without hacks, and as a side effect, we could get rid of the weird static functions in the `framework-core` package (https://github.com/boostercloud/booster/blob/master/packages/framework-core/src/booster.ts#L125-L143)

What do you think @boostercloud/booster-core?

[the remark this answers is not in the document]
NickSeagull: Love the proposal! But sadly what you show here, won't work:

```ts
import { Booster } from '@boostercloud/framework-core'
export { Booster } from '@boostercloud/framework-core'

Booster.start(__dirname)
```

As AWS requires you to have static functions in the global scope in order to call them from lambdas :(((((((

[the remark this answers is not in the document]
rdiaz82: I see a potential issue here, because in this paeticular case `auth-aws` needs to interact with `auth-aws-infrastructure` because the infrastructure needs to create thec ambda itself in the stack and the package needs the lambda code available as a folder or a zip file. Am I missing something?

packages/rocket-auth-aws-infrastructure/src/index.ts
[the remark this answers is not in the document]
charlietfe: Yeah, now with the latest changes I'm able to use the config.

packages/rocket-auth-aws-infrastructure/src/lambdas/challenge-create.ts
charlietfe: Done!

packages/rocket-auth-aws-infrastructure/src/lambdas/package.json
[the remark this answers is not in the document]
javiertoledo: @NickSeagull could we inject them in the global scope as they're "registered" somehow?

```typescript
// Would this even work??
global['nameOfTheLambda'] = leFunction
```

packages/rocket-auth-aws-infrastructure/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "prepack": "tsc -b tsconfig.json",
+    "test": "nyc --extension .ts mocha --forbid-only \"test/**/*.test.ts\""
+  },
+  "bugs": {
+    "url": "https://github.com/boostercloud/booster/issues"
+  },
+  "devDependencies": {
+    "aws-sdk": "2.764.0",
+    "@boostercloud/framework-provider-aws-infrastructure": "^0.11.2",
+    "@boostercloud/framework-types": "^0.11.2"
```
juanjoman, line 49:
```suggestion
    "@boostercloud/framework-provider-aws-infrastructure": "0.11.2",
    "@boostercloud/framework-types": "0.11.2"
```

juanjoman (approved):
LGTM! 👏 Just left a super small suggestion

javiertoledo (approved):
Good work with the tests. We've not resolved [this discussion about lambda registering](https://github.com/boostercloud/booster/pull/556#discussion_r566739406) yet, but it probably doesn't belong here, but in a new issue. I'll create an issue and approve this change.

javiertoledo:
What's stopping us from merging this change?

packages/rocket-auth-aws-infrastructure/src/lambdas/package.json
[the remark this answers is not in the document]
alvaroloes: @javiertoledo We investigated a lot this matter but never thought of that approach 😳 

If it works, it would be amazing!

alvaroloes (approved):
What an AMAZING work @charlietfe 👏 

Seriously, love it. Perfectly integrated with the framework-core roles... 😭  (of happiness)

You already had more than the required approvals, I would say ship it!!

One question! How integration tests are run? I guess when this lives in its own repo, they will be run automatically, but I guess that now we need to run them manually, correct? Or they are run along with the core tests?

Issue #494, opened 2020-11-17
Document that for contributing Node v12.19 is required


NickSeagull:
With version v15, unit tests fail in unexpected ways

PR #560, opened 2021-01-18
doc: Add comment in contributing file indicating that Node v12.19 is …

Moneyba:
…required

## Description
Add a comment in the contributing file indicating that Node v12.19 is required.

## Changes
<Changes made> 

## Checks
- [ ] Project Builds

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -143,6 +143,8 @@ To start contributing to the project you would need to set up the project in you
   - `./scripts/check-all-the-things.sh` on Linux and MacOS
   - `.\scripts\check-all-the-things.ps1` on Windows
 
+**Important:** Node v12 is the minimum version required.
+
 ### Understanding the "lerna monorepo" approach and how dependencies are structured in the project
 
 The Booster Framework project is organized following the ["lerna monorepo"](https://lerna.js.org/) structure. There are several "package.json" files and each one has its purpose with regard to the dependencies you include on them:


CONTRIBUTING.md
```diff
@@ -141,6 +141,7 @@ To start contributing to the project you would need to set up the project in you
   - `./scripts/check-all-the-things.sh` on Linux and MacOS
   - `.\scripts\check-all-the-things.ps1` on Windows
 
+**Important:** Node v12.19 is required.
```
rdiaz82, line 144: I would say that this is the minimim required version, because it sounds that we are supporting just this version

rdiaz82 (changes requested):
I left a nit-picky comment

CONTRIBUTING.md
javiertoledo: Single-click suggestion:
```suggestion
**Important:** Node version v12.19 or greater is required.
```

javiertoledo: It looks like that's actually the case :trollface: https://github.com/boostercloud/booster/issues/494

javiertoledo (approved):

CONTRIBUTING.md
alvaroloes: I have node v14 and everything runs well.

In any case, I would add the "minimum" part and then try to fix the problems with node 15

javiertoledo: Yeah, it's weird that something works only under a specific version, ideally it should work starting in some specific version and in all the next...

Moneyba:
I finally made the suggested change :)

javiertoledo:
Alright! I'll merge this right away, thanks @Moneyba!

PR #561, opened 2021-01-18
Fake provider for integration tests

javiertoledo:
## Description

Current provider integration tests are pure black-box tests where one application is actually deployed to the provider and run through the Booster and provider's APIs to check that everything goes as expected. These tests are great but there's one thing they don't easily allow: Debugging the code as it's run in the provider.

## Changes

In this PR I introduce the concept of the "Fake provider", which is just a JS module that simulates the different operations that are provider-specific. This should allow us to run an integration tests suite locally, to simulate how the framework would behave in a lambda function. This will help us to debug sneaky things like how Booster builds the different objects in the `BoosterConfig` object under different environment circumstances, which are hard to debug with the current testing schemas.

## Checks
- [ ] Project Builds

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -98,3 +98,15 @@ jobs:
       - run: npx lerna bootstrap
       - run: npx lerna run compile
       - run: npx lerna run integration/local --stream
+
+  fake-integration-tests:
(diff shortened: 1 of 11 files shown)


javiertoledo:
Closing, as we're not doing this yet

PR #563, opened 2021-01-20
feature: Print the full GraphQL endpoint after a successful deploy

Moneyba:
## Description
Print the full GraphQL endpoint after a successful deploy

## Changes
<Changes made> 

## Checks
- [ ] Project Builds

### DIFF
--- a/packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/application-stack.ts
+++ b/packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/application-stack.ts
@@ -47,12 +47,12 @@ export class ApplicationStackBuilder {
 
     new CfnOutput(stack, 'httpURL', {
       value: rootAPI.url,
-      description: 'The base URL for all the auth endpoints and for sending GraphQL mutations and queries',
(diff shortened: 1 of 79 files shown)


packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/application-stack.ts
```diff
@@ -50,6 +50,11 @@ export class ApplicationStackBuilder {
       description: 'The base URL for all the auth endpoints and for sending GraphQL mutations and queries',
     })
 
+    new CfnOutput(stack, 'graphQL', {
```
alvaroloes, line 53: I would name this output with the same style as other URLs, `graphqlURL` for consistency (Other outputs that contain a URL are "httpURL" and "websocketURL")

alvaroloes (commented):
Just left a comment

NickSeagull (approved):
Thank you! 🙏

javiertoledo (approved):
LGTM, but consider @alvaroloes's comment, I think it makes sense to keep consistency

PR #570, opened 2021-01-20
Fix for lambda size problem (and probably others related to dependencies)

alvaroloes:
## Description of the problem
We were having a lot of problems when deploying because:
- We were including dev dependencies in the deployed lambdas, making them weight too much and getting errors
- To fix this, we were installing production dependencies and removing the dev ones before deployment, but then we were missing dependencies that were really needed at deployment time.
- To fix the above, we tried to install some of those dependencies globally, but we discovered that there were more dependencies than expected and people started to also have problems with this.

## Description of this PR's solution
Apply the same technique we were applying in the integration tests by creating a sandbox project. In brief:
- When you call `boost deploy`, before doing the actual deployment, we create a sandbox project (it is a copy of the project with only the strictly needed files)
- Go inside that sandbox project, and **install only the production deployment**
- Leave the rest of the deployment process run as always.

## Why this works?
When NPM wants to find a dependency, it first searches inside `./node_modules` folder. If not found, then it goes up one directory level and searches inside `../node_modules`. If not found again, searches inside `../../node_modules`, if not, inside `../../../node_modules`, etc.
The sandboxed project only contains the production dependencies, which means they don't include the _required_ `*-infrastructure` packages. But there is no problem: they will be found one level up, in the original non-sandboxed project, which contains all de devDependencies.

So thanks to this solution, we:
- Only deploy what is strictly needed for runtime
- Can have as many devDependencies as we want
- The deployment works correctly

## Important notice
With this solution, we **don't need to install globally the infrastructure dependencies**, but **they must be listed in the `devDependencies` section of the "package.json" file

## Checks
- [x] Project Builds

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -1,12 +1,11 @@
 import { Command, flags } from '@oclif/command'
 import { deployToCloudProvider } from '../services/provider-service'
-import { compileProjectAndLoadConfig, DEPLOYMENT_SANDBOX } from '../services/config-service'
+import { cleanProjectFiles, compileProjectAndLoadConfig } from '../services/config-service'
 import { BoosterConfig, Logger } from '@boostercloud/framework-types'
 import { Script } from '../common/script'
 import Brand from '../common/brand'
 import { logger } from '../services/logger'
 import { currentEnvironment, initializeEnvironment } from '../services/environment'
-import { removeSandboxProject } from '../common/sandbox'
 
 const runTasks = async (
   skipRestoreDependencies: boolean,
@@ -15,7 +14,7 @@ const runTasks = async (
 ): Promise<void> =>
   Script.init(`boost ${Brand.dangerize('deploy')} [${currentEnvironment()}] 🚀`, compileAndLoad)
     .step('Deploying', (config) => deployer(config, logger))
-    .step('Cleaning up deployment files', async () => removeSandboxProject(DEPLOYMENT_SANDBOX))
+    .step('Cleaning up deployment files', cleanProjectFiles)
     .info('Deployment complete!')
     .done()
 
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -3,7 +3,7 @@ import { deployToCloudProvider } from '../services/provider-service'
 import {
   cleanProductionSandbox,
   compileProjectAndLoadConfig,
-  createProductionSandbox,
+  createDeploymentSandbox,
 } from '../services/config-service'
 import { BoosterConfig, Logger } from '@boostercloud/framework-types'
 import { Script } from '../common/script'
@@ -36,9 +36,8 @@ export default class Deploy extends Command {
     const { flags } = this.parse(Deploy)
 
     if (initializeEnvironment(logger, flags.environment)) {
-      const deploymentProjectPath = await createProductionSandbox()
-      process.chdir(deploymentProjectPath) // The deployment will work inside the production sandbox
-      await runTasks(compileProjectAndLoadConfig(), deployToCloudProvider)
+      const deploymentProjectPath = await createDeploymentSandbox()
+      await runTasks(compileProjectAndLoadConfig(deploymentProjectPath), deployToCloudProvider)
     }
   }
 }
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -1,7 +1,7 @@
 import { Command, flags } from '@oclif/command'
 import { deployToCloudProvider } from '../services/provider-service'
 import {
-  cleanProductionSandbox,
+  cleanDeploymentSandbox,
   compileProjectAndLoadConfig,
   createDeploymentSandbox,
 } from '../services/config-service'
@@ -17,7 +17,7 @@ const runTasks = async (
 ): Promise<void> =>
   Script.init(`boost ${Brand.dangerize('deploy')} [${currentEnvironment()}] 🚀`, compileAndLoad)
     .step('Deploying', (config) => deployer(config, logger))
-    .step('Cleaning up deployment files', cleanProductionSandbox)
+    .step('Cleaning up deployment files', cleanDeploymentSandbox)
     .info('Deployment complete!')
     .done()
 
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -1,6 +1,10 @@
 import { Command, flags } from '@oclif/command'
 import { deployToCloudProvider } from '../services/provider-service'
-import { cleanProjectFiles, compileProjectAndLoadConfig } from '../services/config-service'
+import {
+  cleanProductionSandbox,
+  compileProjectAndLoadConfig,
+  createProductionSandbox,
+} from '../services/config-service'
 import { BoosterConfig, Logger } from '@boostercloud/framework-types'
 import { Script } from '../common/script'
 import Brand from '../common/brand'
@@ -13,7 +17,7 @@ const runTasks = async (
 ): Promise<void> =>
   Script.init(`boost ${Brand.dangerize('deploy')} [${currentEnvironment()}] 🚀`, compileAndLoad)
     .step('Deploying', (config) => deployer(config, logger))
-    .step('Cleaning up deployment files', cleanProjectFiles)
+    .step('Cleaning up deployment files', cleanProductionSandbox)
     .info('Deployment complete!')
     .done()
 
@@ -32,6 +36,8 @@ export default class Deploy extends Command {
     const { flags } = this.parse(Deploy)
 
     if (initializeEnvironment(logger, flags.environment)) {
+      const deploymentProjectPath = await createProductionSandbox()
+      process.chdir(deploymentProjectPath) // The deployment will work inside the production sandbox
       await runTasks(compileProjectAndLoadConfig(), deployToCloudProvider)
     }
   }
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -1,12 +1,12 @@
 import { Command, flags } from '@oclif/command'
 import { deployToCloudProvider } from '../services/provider-service'
-import { compileProjectAndLoadConfig } from '../services/config-service'
+import { compileProjectAndLoadConfig, DEPLOYMENT_SANDBOX } from '../services/config-service'
 import { BoosterConfig, Logger } from '@boostercloud/framework-types'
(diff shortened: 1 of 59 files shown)


packages/cli/src/commands/deploy.ts
```diff
@@ -28,21 +30,15 @@ export default class Deploy extends Command {
       char: 'e',
       description: 'environment configuration to run',
     }),
-    skipRestoreDependencies: flags.boolean({
-      char: 's',
-      description: 'skips restoring dependencies after deployment',
-    }),
```
alvaroloes, line 34: No need for this anymore

packages/cli/src/commands/deploy.ts
```diff
@@ -28,21 +30,15 @@ export default class Deploy extends Command {
...
     const { flags } = this.parse(Deploy)
 
     if (initializeEnvironment(logger, flags.environment)) {
-      await runTasks(
-        flags.skipRestoreDependencies,
-        compileProjectAndLoadConfig({ production: true }),
-        deployToCloudProvider
-      )
+      const deploymentProjectPath = await createProductionSandbox()
+      process.chdir(deploymentProjectPath) // The deployment will work inside the production sandbox
```
alvaroloes, line 40: These two lines are the main change of this PR

packages/cli/src/common/sandbox.ts
```diff
@@ -0,0 +1,29 @@
+import { copyFileSync, mkdirSync, readdirSync, rmdirSync } from 'fs'
+import * as path from 'path'
+
```
alvaroloes, line 3: The functions of this file are copied and pasted from `packages/framework-integration-tests/integration/helper/fileHelper.ts`, as I don't know how to share common functionality like this inside a lerna monorepo (without creating an extra package, of course).

Does anybody have an idea to share helper function accross packages in a lerna monorepo?

NickSeagull: Essentially creating a package :(

NickSeagull (approved):
LGTM

rdiaz82 (approved):
LGTM!!

The only thing that I see with this PR is that we need a mechanism now to add folders to the `.deploy` from rockets. Currently, we have the rocket interface with:
```typescript
 export interface InfrastructureRocket {
  mountStack: (stack: Stack) => void
  unmountStack?: (utils: RocketUtils) => void
}
```
probably we can add another method to the interface to add folders to the `.deploy` folder during the deployment. In fact, right now we are breaking the static site rocket because the rocket is doing this:
```typescript
const rootPath = params.rootPath ?? './public'
```
and now that folder is unaccessible from the `.deploy` folder. 

**Anyway this is a task a that can be addressed in other PR because it is another issue.**

davidverdu:
LGTM. Nonetheless, Should the new:project `package.json` template be updated accordingly, just to consider the dependencies as you have explained for those who create new projects?

davidverdu (approved):

adriangmweb (approved):
LGTM!!

packages/cli/src/services/config-service.ts
```diff
@@ -4,18 +4,24 @@ import { exec } from 'child-process-promise'
...
 import { currentEnvironment } from './environment'
-import { pruneDevDependencies } from './dependencies'
+import { createSandboxProject, removeSandboxProject } from '../common/sandbox'
+import { installProductionDependencies } from './dependencies'
 
-type CompileAndLoadOptions = {
-  production: boolean
+export const DEPLOYMENT_SANDBOX = '.deploy'
+
+export async function createProductionSandbox(): Promise<string> {
```
javiertoledo, line 12: I think we should change the name of this function, right now in Booster you can define the name of your environment however you want and we should keep this code generic. Suggestion:

```suggestion
export async function createDeploymentSandbox(): Promise<string> {
```

packages/cli/src/services/config-service.ts
```diff
@@ -4,18 +4,24 @@ import { exec } from 'child-process-promise'
...
+  const sandboxRelativePath = createSandboxProject(DEPLOYMENT_SANDBOX)
+  await installProductionDependencies(sandboxRelativePath)
+  return sandboxRelativePath
 }
 
-export async function compileProjectAndLoadConfig(opts?: CompileAndLoadOptions): Promise<BoosterConfig> {
-  const userProjectPath = process.cwd()
+export async function cleanProductionSandbox(): Promise<void> {
```
javiertoledo, line 18:
```suggestion
export async function cleanDeploymentSandbox(): Promise<void> {
```

packages/cli/src/services/config-service.ts
javiertoledo: Would it make sense to move these sandbox management functions to their own file? It's weird that they are in `config-service.ts`

packages/cli/src/commands/deploy.ts
```diff
@@ -1,21 +1,23 @@
 import { Command, flags } from '@oclif/command'
 import { deployToCloudProvider } from '../services/provider-service'
-import { compileProjectAndLoadConfig } from '../services/config-service'
+import {
+  cleanProductionSandbox,
+  compileProjectAndLoadConfig,
+  createProductionSandbox,
```
javiertoledo, line 6: I wonder if sandbox-related operations belong to this scope, they could perfectly be used from the `compileProjectAndLoadConfig` process without managing them from here, they're a bit "low level" to be interesting to the user 🤔

packages/cli/src/common/sandbox.ts
javiertoledo: If you want a 100% clean solution, these functions should be in a module that is imported from all the other packages anyway like `framework-types`, and renaming `framework-types` as `framework-common`. This is a refactor I'd love to do, because I've never liked the idea of separating types in a module just because they're  types...

But we're talking about code that is used both from production code (the CLI) and the integration tests. I think that we can safely assume that integration tests are going to be "always" executed in a context where you have access to this file, no matter where it is, so you can delete the copy in the integration test package and just import this with an extra serve of `'..'` in your path. As soon as the file is reachable it can be used and it's not a big deal to break the package structure from integration tests because any changes that make it stop working will be detected by the compiler anyway 😜

javiertoledo (changes requested):
Great work! I left a few comments for you to consider

packages/cli/src/commands/deploy.ts
alvaroloes: I tried that! But it didn't work. The reason is that the Booster application configuration expects all files to be in CWD **when we do the  "require" of the project files**, and we can't specify a different CWD in a `require` call. I will try to try again to see what we can do

packages/cli/src/common/sandbox.ts
alvaroloes: Good suggestions, thanks guys! Maybe I try to create a private package called "common" or "util", because this is starting to occur many times. The problem with that is that the package will be published and it doesn't make much sense to publish this. 
I think I'll go with the Javi approach. Let's see

packages/cli/src/services/config-service.ts
alvaroloes: 🤣  I had that name but then I changed it! Anyway, you convinced me to use it now. Thanks!

Regarding moving these to their own file: they are! the generic version of the functions are in the file `sandbox.ts`. These functions are just wrappers that use some parameters that are specific to the scope of the `config-service` responsibility.

alvaroloes:
@rdiaz82 yep! that's an important point. I will create another issue for that because merging this is a bit in a hurry

@davidverdu Thanks for pointing that! I'll review that file to see if something need to change. Thanks!

packages/cli/src/commands/deploy.ts
alvaroloes: @javiertoledo I finally was able to fixed this. It required more changes than expected, but as a result, now there is much less code that depends on the global `proces.cwd()`. One of the flagrant cases was that the code paths for the lambas were set using `proces.cwd()` which could cause several problems.

Fixed here: f2b535ba39349e5d8006e9d4afefb5009f1cf4c2

packages/cli/src/services/config-service.ts
alvaroloes: Fixed here: fc385d3a4b4154a586aa6bf9b832fbe5a42a0846

packages/cli/src/commands/deploy.ts
alvaroloes: I first wrote it in the `compileAndLoadConfig`, but I had quickly to change it because of the following reasons:
- Many commands use the `compileAndLoadConfig`, but only the `deploy` command needs to create a sandbox.
- That function works with a project, and we first need to create the project, which in this case is done with `createProductionSandbox`

packages/cli/src/common/sandbox.ts
alvaroloes: Done here: 392efbd4d349b5132defa024aa3a968736b1a6ef

packages/framework-integration-tests/src/index.ts
```diff
@@ -8,4 +8,4 @@ export {
   boosterTriggerScheduledCommand,
 } from '@boostercloud/framework-core'
 
-Booster.start()
+Booster.start(__dirname)
```
NickSeagull, line 11: Do we have now to change the project generator to have this as well?

alvaroloes: Yes @NickSeagull I already change the template. And another yes: this is a breaking change, but a really easy one to solve. I will let everybody know (I will create a release note for this)

packages/cli/src/commands/nuke.ts
```diff
@@ -56,7 +56,7 @@ export default class Nuke extends Command {
     const { flags } = this.parse(Nuke)
     if (initializeEnvironment(logger, flags.environment)) {
       await runTasks(
-        askToConfirmRemoval(new Prompter(), flags.force, compileProjectAndLoadConfig()),
+        askToConfirmRemoval(new Prompter(), flags.force, compileProjectAndLoadConfig(process.cwd())),
```
javiertoledo, line 59: Do we really need to call `process.cwd()` here and pass the result as a value to `compileProjectAndLoadConfig`? Calling to it from `compileProjectAndLoadConfig` should make no difference. AFAIK, the process doesn't change over the whole program unless you call to `chdir`

javiertoledo (approved):
Alright! Left a minor comment

packages/cli/src/commands/nuke.ts
alvaroloes: I think I get your point. Yes, we need to pass now the directory we want to compile, as `compileProjectAndLoadConfig` is called in other places in the code (when deploying, for example) with a different path than CWD (the path of the blackbox project). I did this so that we don't need to use `process.chdir` 😃

PR #571, opened 2021-01-21
[558] added pagination to save events

rdiaz82:
## Description
Implemented the pagination in save events. Now the events are saved in batchs of 25 elements. 

## Changes
Modify `event-adapter.ts` adding the pagination algorithm and move the current save events function into another one just for save the batch
Added test in `events-adapter.test.ts` to check that the pagination is working. The case for less than 25 elements are covered with the previous tests in the file.  
## Checks
- [x] Project Builds
- [ ] Updated documentation accordingly (Not necessary for this PR)


packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -88,7 +88,25 @@ export async function storeEvents(
...
   logger: Logger
 ): Promise<void> {
-  // TODO: Manage query pagination and db.batchWrite limit of 25 operations at a time
+  const batchItemsNumbers = 25
+  let processedItems = 0
+  while (Math.floor((eventEnvelopes.length - processedItems) / batchItemsNumbers) != 0) {
+    const currentBatch = eventEnvelopes.slice(processedItems, processedItems + batchItemsNumbers)
+    await persistBatch(logger, currentBatch, config, dynamoDB)
+    processedItems += batchItemsNumbers
+  }
```
NickSeagull, line 97: Can you extract the `while` condition into a separate function, and the body as well? It's a bit difficult to read what's going on here

packages/framework-provider-aws/test/library/events-adapter.test.ts
```diff
@@ -130,6 +130,35 @@ describe('the events-adapter', () => {
...
+          id: 'id',
+        },
+        typeName: 'EventName',
+        entityTypeName: 'EntityName',
+        requestID: 'requestID',
+        createdAt: 'once',
+      }
+      const eventEnvelopes = []
+      let numberEvents = 51
+      while (numberEvents--) eventEnvelopes.push(event)
```
NickSeagull, line 151: This can be
```suggestion
      const eventEnvelopes = Array.from({length: 51}, () => event)
```

NickSeagull (changes requested):

packages/framework-provider-aws/test/library/events-adapter.test.ts
rdiaz82: nice!!

packages/framework-provider-aws/src/library/events-adapter.ts
rdiaz82: I ended refatoring the wway it was done :P

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -88,7 +88,37 @@ export async function storeEvents(
...
+function getBatchFromEnvelopes(
+  eventEnvelopes: EventEnvelope[],
+  processedItems: number,
+  batchSize: number
+): EventEnvelope[] {
+  if (!validBatchSize(eventEnvelopes, processedItems, batchSize)) {
+    return eventEnvelopes.slice(processedItems)
+  }
+  return eventEnvelopes.slice(processedItems, processedItems + batchSize)
+}
```
NickSeagull, line 114: Now that I think on it, why not doing it the following way:

1. Separate in chunks
2. Iterate and persist

Something like

```suggestion
  const batches = inChunksOf(25, eventEnvelopes)
  for (const batch of batches) {
    await persistBatch(logger, batch, config, dynamoDB)
  }
}

function inChunksOf<TElement>(n: number, arr: Array<TElement>): Array<TElement> {
  var result = []
  while (arr.length) {
    result.push(arr.splice(0, chunk_size))
  }
  return result
}

function validBatchSize(eventEnvelopes: EventEnvelope[], processedItems: number, batchSize: number): boolean {
  return Math.floor((eventEnvelopes.length - processedItems) / batchSize) != 0
}
```

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -88,7 +88,37 @@ export async function storeEvents(
   config: BoosterConfig,
   logger: Logger
 ): Promise<void> {
-  // TODO: Manage query pagination and db.batchWrite limit of 25 operations at a time
+  const batchSize = 25
```
javiertoledo, line 91: Is there any constant or function in the SDK that brings us this value? Having this hardcoded here feels a bit fragile...

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -88,7 +88,37 @@ export async function storeEvents(
...
+  let processedItems = 0
+  let currentBatch = getBatchFromEnvelopes(eventEnvelopes, processedItems, batchSize)
+  while (currentBatch.length != 0) {
+    await persistBatch(logger, currentBatch, config, dynamoDB)
+    processedItems += currentBatch.length
+    currentBatch = getBatchFromEnvelopes(eventEnvelopes, processedItems, batchSize)
+  }
```
javiertoledo, line 98: This is one of those rare scenarios where a `do..while` could make sense:

```suggestion
  do {
    currentBatch = getBatchFromEnvelopes(eventEnvelopes, processedItems, batchSize)
    await persistBatch(logger, currentBatch, config, dynamoDB)
    processedItems += currentBatch.length
  } while (currentBatch.length > 0)
```

I'm also paranoid with strict comparations with indexes, so I'd use a `>` instead of `!=`

packages/framework-provider-aws/src/library/events-adapter.ts
javiertoledo: Using @NickSeagull's implementation, with some extra, you could even send all chunks in parallel (not sure if that makes sense TBH):

```typescript
const batchSize = 25
Promise.allSettled(
  inChunksOf(batchSize, eventEnvelopes) // Builds the chunks
    .map((batch) => persistBatch(logger, batch, config, dynamoDB)) // Maps them to promises
)  // Promise.allSettled awaits for all of them

function inChunksOf<TElement>(n: number, arr: Array<TElement>): Array<TElement> {
  let result = []
  while (arr.length) {
    result.push(arr.splice(0, chunk_size))
  }
  return result
}
```

packages/framework-provider-aws/test/library/events-adapter.test.ts
```diff
@@ -130,6 +130,33 @@ describe('the events-adapter', () => {
...
+      const eventEnvelopes = Array.from({ length: 51 }, () => event)
+
+      const fakeBatchWrite = fake.returns({
+        promise: fake.resolves(''),
+      })
+      const fakeDynamo: DocumentClient = { batchWrite: fakeBatchWrite } as any
+
+      await Library.storeEvents(fakeDynamo, eventEnvelopes, config, fakeLogger)
+
+      expect(fakeBatchWrite).to.be.calledThrice
```
javiertoledo, line 158: Would it make sense to check that each call contains the right number of items and that they're in the expected order?

javiertoledo (commented):
Great change, I added a few comments to the discussion

packages/framework-provider-aws/src/library/events-adapter.ts
rdiaz82: @javiertoledo I really like your approach sending all chunks in parallel but I'm not sure about the order because if we are saving all the events in parallel we can't guarantee in which order are they saved in the event store.  What do you think?

javiertoledo: Good point. Events are always read in order, using the timestamps as a sort key, so if we're generating the timestamps within this loop, then we could easily lose the right order. If we could generate the timestamps before and make the inserts in parallel, I think it won't matter in which order they are inserted. Anyway, it's a micro-optimization and it's hard to predict what will happen as we're also relying on event insertion to trigger event handlers, snapshooting, etc. So 100% agree that it's better to avoid parallelizing this at this time :trollface:

packages/framework-provider-aws/src/library/events-adapter.ts
alvaroloes: Agree here

packages/framework-provider-aws/src/library/events-adapter.ts
alvaroloes: Love the Nicks suggestion. Regarding the timestams, they are generated in the `core` so, here, when we already have the event envelopes, it would be save to store in parallel. BUT! 100% agree that we should forget now about that optimization too

packages/framework-provider-aws/test/library/events-adapter.test.ts
alvaroloes: Agree here too. It would be really dangerous that, because of rounding problems, we forget to persist some events

alvaroloes (approved):
Pre-approving now. But please check the comments, especially the one about a more detailed test ;-)

NickSeagull:
I'm picking up this PR as it's priority

juanjoman (changes requested):
Overall LGTM, but I'd love to see some integration tests 👀

NickSeagull:
Makes sense, thanks @juanjoman

NickSeagull (approved):
PR ready to review!

juanjoman (approved):
LGTM!! :v:

PR #573, opened 2021-01-21
New GraphQL filters

adriangmweb:
## Description
This new way of filtering is based on the Amplify one, which is way less limiting than the previous one.

## Changes
* The core `searcher` function has been changed to the new format. The old filter system is still compatible until we update the rest of the providers.
* The AWS `searcher-adapter` has been modified to translate these new filters to DynamoDB queries.
* Unit tests for AWS `searcher-adapter` has been added to ensure stability across all the different scenarios.

## Checks
- [x] Project Builds
- [x] Updated documentation accordingly - Just basic documentation. The rest will be added when the GraphQL schema gets updated in a new PR.
 
## Additional information
This is working just for retrieving and filtering ReadModels inside commands, events, etc. The GraphQL schema will be changed in a future PR.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1565,7 +1565,7 @@ Booster GraphQL API also provides support for real-time updates using subscripti
 
 #### Getting and filtering read models data at code level
 
-Booster allows you to get your read models data in your commands, handlers, etc. Using the `Booster.readModel` method.
+Booster allows you to get your read models data in your commands handlers and event handlers using the `Booster.readModel` method.
 
 For example, you can filter and get the total number of the products that meet your criteria in your commands like this:
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,99 +3,99 @@
 <!-- toc -->
 
 - [Introduction](#introduction)
-  * [What is Booster?](#what-is-booster)
-  * [Booster Principles](#booster-principles)
-  * [Why use Booster](#why-use-booster)
+  - [What is Booster?](#what-is-booster)
+  - [Booster Principles](#booster-principles)
+  - [Why use Booster](#why-use-booster)
 - [Getting started](#getting-started)
-  * [Installing Booster](#installing-booster)
-    + [Prerequisites](#prerequisites)
-    + [Installing the Booster CLI](#installing-the-booster-cli)
-  * [Your first Booster app in 10 minutes](#your-first-booster-app-in-10-minutes)
-    + [1. Create the project](#1-create-the-project)
-    + [2. First command](#2-first-command)
-    + [3. First event](#3-first-event)
-    + [4. First entity](#4-first-entity)
-    + [5. First read model](#5-first-read-model)
-    + [6. Deployment](#6-deployment)
-    + [7. Testing](#7-testing)
-    + [8. Removing the stack](#8-removing-the-stack)
-    + [9. More functionalities](#9-more-functionalities)
+  - [Installing Booster](#installing-booster)
+    - [Prerequisites](#prerequisites)
+    - [Installing the Booster CLI](#installing-the-booster-cli)
+  - [Your first Booster app in 10 minutes](#your-first-booster-app-in-10-minutes)
+    - [1. Create the project](#1-create-the-project)
+    - [2. First command](#2-first-command)
+    - [3. First event](#3-first-event)
+    - [4. First entity](#4-first-entity)
+    - [5. First read model](#5-first-read-model)
+    - [6. Deployment](#6-deployment)
+    - [7. Testing](#7-testing)
+    - [8. Removing the stack](#8-removing-the-stack)
+    - [9. More functionalities](#9-more-functionalities)
 - [Booster architecture](#booster-architecture)
-  * [1. Command and command handlers](#1-command-and-command-handlers)
-    + [Commands naming convention](#commands-naming-convention)
-    + [Creating a command](#creating-a-command)
-    + [The command handler function](#the-command-handler-function)
-    + [Authorizing a command](#authorizing-a-command)
-    + [Submitting a command](#submitting-a-command)
-    + [Scheduling a command](#scheduling-a-command)
-    + [Creating a scheduled command](#creating-a-scheduled-command)
-  * [2. Events](#2-events)
-    + [Events naming convention](#events-naming-convention)
-    + [Creating events](#creating-events)
-    + [Registering events in the event store](#registering-events-in-the-event-store)
-    + [Events ordering](#events-ordering)
-  * [3. Event handlers](#3-event-handlers)
-    + [Creating an event handler](#creating-an-event-handler)
-    + [Registering events from an event handler](#registering-events-from-an-event-handler)
-    + [Reading entities from event handlers](#reading-entities-from-event-handlers)
-  * [4. Entities and reducers](#4-entities-and-reducers)
-    + [Entities naming convention](#entities-naming-convention)
-    + [Creating entities](#creating-entities)
-    + [The reducer function](#the-reducer-function)
-    + [Eventual consistency](#eventual-consistency)
-  * [5. Read models and projections](#5-read-models-and-projections)
-    + [Read models naming convention](#read-models-naming-convention)
-    + [Creating a read model](#creating-a-read-model)
-    + [The projection function](#the-projection-function)
-    + [Authorizing read models](#authorizing-read-models)
-    + [Querying a read model](#querying-a-read-model)
-    + [Getting real-time updates for a read model](#getting-real-time-updates-for-a-read-model)
+  - [1. Command and command handlers](#1-command-and-command-handlers)
+    - [Commands naming convention](#commands-naming-convention)
+    - [Creating a command](#creating-a-command)
+    - [The command handler function](#the-command-handler-function)
+    - [Authorizing a command](#authorizing-a-command)
+    - [Submitting a command](#submitting-a-command)
+    - [Scheduling a command](#scheduling-a-command)
+    - [Creating a scheduled command](#creating-a-scheduled-command)
+  - [2. Events](#2-events)
+    - [Events naming convention](#events-naming-convention)
+    - [Creating events](#creating-events)
+    - [Registering events in the event store](#registering-events-in-the-event-store)
+    - [Events ordering](#events-ordering)
+  - [3. Event handlers](#3-event-handlers)
+    - [Creating an event handler](#creating-an-event-handler)
+    - [Registering events from an event handler](#registering-events-from-an-event-handler)
+    - [Reading entities from event handlers](#reading-entities-from-event-handlers)
+  - [4. Entities and reducers](#4-entities-and-reducers)
+    - [Entities naming convention](#entities-naming-convention)
+    - [Creating entities](#creating-entities)
+    - [The reducer function](#the-reducer-function)
+    - [Eventual consistency](#eventual-consistency)
+  - [5. Read models and projections](#5-read-models-and-projections)
+    - [Read models naming convention](#read-models-naming-convention)
+    - [Creating a read model](#creating-a-read-model)
+    - [The projection function](#the-projection-function)
+    - [Authorizing read models](#authorizing-read-models)
+    - [Querying a read model](#querying-a-read-model)
+    - [Getting real-time updates for a read model](#getting-real-time-updates-for-a-read-model)
 - [Features](#features)
-  * [Authentication and Authorization](#authentication-and-authorization)
-    + [Authentication API](#authentication-api)
-  * [Custom Authentication](#custom-authentication)
-    + [JWT Configuration](#jwt-configuration)
-  * [GraphQL API](#graphql-api)
-    + [Relationship between GraphQL operations and commands and read models](#relationship-between-graphql-operations-and-commands-and-read-models)
-    + [How to send GraphQL request](#how-to-send-graphql-request)
-    + [Get GraphQL schema from deployed application](#get-graphql-schema-from-deployed-application)
-    + [Sending commands](#sending-commands)
-    + [Reading read models](#reading-read-models)
-    + [Subscribing to read models](#subscribing-to-read-models)
-    + [Using Apollo Client](#using-apollo-client)
-    + [Authorizing operations](#authorizing-operations)
-    + [The GraphQL over WebSocket protocol](#the-graphql-over-websocket-protocol)
-  * [Cloud native](#cloud-native)
-    + [Configure your provider credentials](#configure-your-provider-credentials)
-    + [Deploy your project](#deploy-your-project)
-    + [Application outputs](#application-outputs)
-    + [Delete your cloud stack](#delete-your-cloud-stack)
+  - [Authentication and Authorization](#authentication-and-authorization)
+    - [Authentication API](#authentication-api)
+  - [Custom Authentication](#custom-authentication)
+    - [JWT Configuration](#jwt-configuration)
+  - [GraphQL API](#graphql-api)
+    - [Relationship between GraphQL operations and commands and read models](#relationship-between-graphql-operations-and-commands-and-read-models)
+    - [How to send GraphQL request](#how-to-send-graphql-request)
+    - [Get GraphQL schema from deployed application](#get-graphql-schema-from-deployed-application)
+    - [Sending commands](#sending-commands)
+    - [Reading read models](#reading-read-models)
+    - [Subscribing to read models](#subscribing-to-read-models)
+    - [Using Apollo Client](#using-apollo-client)
+    - [Authorizing operations](#authorizing-operations)
+    - [The GraphQL over WebSocket protocol](#the-graphql-over-websocket-protocol)
+  - [Cloud native](#cloud-native)
+    - [Configure your provider credentials](#configure-your-provider-credentials)
+    - [Deploy your project](#deploy-your-project)
+    - [Application outputs](#application-outputs)
+    - [Delete your cloud stack](#delete-your-cloud-stack)
 - [Going deeper with Booster](#going-deeper-with-booster)
-  * [Contributing](#contributing)
-  * [Framework Core](#framework-core)
-  * [Framework Types](#framework-types)
-  * [Framework integration tests](#framework-integration-tests)
-  * [Providers](#providers)
-    + [framework-provider-aws-\*](#framework-provider-aws-)
-    + [framework-provider-local-\*](#framework-provider-local-)
-    + [framework-provider-kubernetes-\*](#framework-provider-kubernetes-)
-    + [framework-provider-azure-\*](#framework-provider-azure-)
-  * [Configuration and environments](#configuration-and-environments)
-    + [Booster configuration](#booster-configuration)
-    + [Environments](#environments)
-  * [Extending Booster with Rockets!](#extending-booster-with-rockets)
-    + [Naming recommendations](#naming-recommendations)
-    + [Infrastructure extensions](#infrastructure-extensions)
-    + [Runtime extensions (Not yet implemented)](#runtime-extensions-not-yet-implemented)
-    + [Deploy and init hooks (Not yet implemented)](#deploy-and-init-hooks-not-yet-implemented)
+  - [Contributing](#contributing)
+  - [Framework Core](#framework-core)
+  - [Framework Types](#framework-types)
+  - [Framework integration tests](#framework-integration-tests)
+  - [Providers](#providers)
+    - [framework-provider-aws-\*](#framework-provider-aws-)
+    - [framework-provider-local-\*](#framework-provider-local-)
+    - [framework-provider-kubernetes-\*](#framework-provider-kubernetes-)
+    - [framework-provider-azure-\*](#framework-provider-azure-)
+  - [Configuration and environments](#configuration-and-environments)
+    - [Booster configuration](#booster-configuration)
+    - [Environments](#environments)
+  - [Extending Booster with Rockets!](#extending-booster-with-rockets)
+    - [Naming recommendations](#naming-recommendations)
+    - [Infrastructure extensions](#infrastructure-extensions)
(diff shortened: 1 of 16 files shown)


packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
```diff
@@ -27,43 +28,46 @@ function filterReadModel(readModel: Record<string, any>, filters?: Record<string
...
+        case 'contains':
+          if (!contains(readModelPropValue, value)) return false
+          break
+        case 'not-contains':
+          if (contains(readModelPropValue, value)) return false
+          break
+        case 'begins-with':
+          if (!beginWith(readModelPropValue, value as string)) return false
+          break
+      }
```
NickSeagull, line 70: This is a great opportunity to extract this into a function. By the way, note the `console.log`, not sure if you meant to leave it there

packages/framework-provider-aws/src/library/searcher-adapter.ts
```diff
@@ -9,7 +10,7 @@ export async function searchReadModel(
   config: BoosterConfig,
   logger: Logger,
   readModelName: string,
-  filters: Record<string, Filter<any>>
+  filters: FilterFor<any>
```
NickSeagull, line 13: Can we use `unknown` instead of `any` here?

adriangmweb: What would be the difference here @NickSeagull? I'm still learning some things about TS :)

NickSeagull: The `unknown` type forces the programmer to cast the type to the appropriate type when its needed, while `any` can be passed around freely:

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
```diff
@@ -28,41 +29,43 @@ function filterReadModel(readModel: Record<string, any>, filters?: Record<string
...
       case 'not-contains':
-        if (contains(readModelPropValue, values[0])) return false
+        if (contains(readModelPropValue, value)) return false
         break
       case 'begins-with':
-        if (!beginWith(readModelPropValue, values[0] as string)) return false
+        if (!beginWith(readModelPropValue, value as string)) return false
         break
     }
   }
```
javiertoledo, line 71: After the previous changes you can also remove the `return true` at the end of the method.

packages/framework-provider-aws/src/library/searcher-adapter.ts
```diff
@@ -32,62 +31,150 @@ export async function searchReadModel(
...
+        break
+      case 'and':
+      case 'or':
+        for (const filter of filters[propName] as Array<FilterFor<any>>) {
+          Object.assign(attributeValues, buildExpressionAttributeValues(filter as FilterFor<any>))
+        }
+        break
+      default:
+        Object.assign(attributeValues, buildAttributeValue(propName, filters[propName]))
+        break
```
javiertoledo, line 158: This switch is pretty similar to the one in `buildExpressionAttributeNames`. It makes me wonder if we could abstract this somehow to make both functions simpler 🤔 (Not a requirement for this PR though)

packages/framework-provider-aws/src/library/searcher-adapter.ts
```diff
@@ -32,62 +31,150 @@ export async function searchReadModel(
...
+        Object.assign(attributeNames, buildExpressionAttributeNames(filters[propName] as FilterFor<any>))
+        break
+      case 'and':
+      case 'or':
+        for (const filter of filters[propName] as Array<FilterFor<any>>) {
+          Object.assign(attributeNames, buildExpressionAttributeNames(filter as FilterFor<any>))
+        }
+        break
+      case 'includes':
+        break
```
javiertoledo, line 129: What happens with this case? If we're planning to implement it in the future it could make sense to leave a TODO here:

```suggestion
      case 'includes':
        // TODO: To be implemented in the future. More details in issue #XYZ
        break
```

packages/framework-provider-aws/test/library/searcher-adapter.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        ExpressionAttributeNames: { '#id': 'id' },
+        ExpressionAttributeValues: {
+          ':id_0': '3',
+          ':id_1': '333',
+        },
+      }
+      const filters: FilterFor<Product> = {
+        id: { contains: '3' },
+        not: { id: { eq: '333' } },
+      }
```
javiertoledo, line 107: Pretty cool DSL. Let's make sure it's properly documented before merging this PR, otherwise nobody will use this feature!

packages/framework-provider-local/src/library/searcher-adapter.ts
```diff
@@ -1,4 +1,4 @@
-import { Filter } from '@boostercloud/framework-types'
+import { FilterOld } from '@boostercloud/framework-types'
```
javiertoledo, line 1: Why renaming this type as `FilterOld`?

packages/framework-types/src/searcher.ts
```diff
@@ -33,13 +41,14 @@ export class Searcher<TObject> {
    * @param property The property the filter will act upon
    * @param operation The filter operation.
    * @param values The values for the filter. Depending on the operation, you can specify here one or many values
+   * @deprecated Use "filter" instead
```
javiertoledo, line 44: Why deprecating it if we're changing the name anyway producing breaking changes? Would it make sense to remove this code entirely? What would be the consequences? Is other part of the system using it? Is this to avoid making the PR too long? If that's the case, make sure you add an issue to remove this deprecated code in the near future!

docs/README.md
```diff
@@ -3,99 +3,100 @@
...
-    + [Performing GraphQL requests](#performing-graphql-requests)
-    + [Inspect Database information](#inspect-database-information)
-  * [Booster examples](#booster-examples)
+  - [Running Booster applications locally](#running-booster-applications-locally)
+    - [Local development prerequisites](#local-development-prerequisites)
+    - [Starting your application](#starting-your-application)
+    - [Performing Auth requests](#performing-auth-requests)
+    - [Performing GraphQL requests](#performing-graphql-requests)
+    - [Inspect Database information](#inspect-database-information)
+  - [Booster examples](#booster-examples)
```
javiertoledo, line 99: It's weird that you changed the format in the whole TOC. Are you using the linter config in `.markdownlintrc`? If you are, there might be some extra options we should add to the config to make sure we all use the same.

javiertoledo (changes requested):
Nice changes, just left a few comments to discuss. There are two points that I'd consider completing before merging:
1. Extensive documentation about the filtering DSL with all the available keywords and their meaning (We could agree on doing this in a different PR if we create an issue). 
2. We should also make sure that the DSL is the same for the API and for the in-handler interface (it's hard to see in the code review without digging much more in the code)
3. I'd consider removing the deprecated `filterOld` function and related code. As we're changing the function name, it's a breaking change anyway, and users could keep working with the current version until they make the changes for the new one...

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
```diff
@@ -4,6 +4,7 @@ import {
   ReadModelInterface,
   ReadModelPropertyFilter,
   Instance,
+  Operation,
```
alvaroloes, line 7: @adriangmweb This file should also be adapted to the new filter structure, but I see it keeps the same old structure, so subscriptions won't work properly.

_Maybe you want to do this in another PR? If so, please create the issue_

Context: When you subscribe to a ReadModel, you use the same filter as when querying them. We need to implement the behavior of those filters (as you did in AWS) but, this time, by filtering the array of read models in memory, in pure javascript (instead of with AWS SQL).

The GraphQL engine will call this code with an array of ReadModels of a specific type that are eligible to send to the user who subscribed, but we need to filter them using the filters the user specified in the Subscriptoin opreation.

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
alvaroloes: Take into account the previous comment marked with ⚠️ . The read model needs to pass _all the filters_

packages/framework-provider-aws/src/library/searcher-adapter.ts
alvaroloes: I _think_ this is OK, because, after the include operation, there is only a value. There are no attribute names that needs to be accounted for in the AWS query.

Anyway, if that is the case, I agree that it would be great to have a comment explaining that.

packages/framework-provider-aws/src/library/searcher-adapter.ts
```diff
@@ -32,62 +31,150 @@ export async function searchReadModel(
   return result.Items ?? []
 }
 
-function buildFilterExpression(filters: Record<string, Filter<any>>): string {
+function buildExpression(filters: FilterFor<any>): any {
+  const expresion = buildFilterExpression(filters)
+  const names = buildExpressionAttributeNames(filters)
+  resetAttrCount()
+  const values = buildExpressionAttributeValues(filters)
+  resetAttrCount()
```
alvaroloes, line 39: I see you call `resetAttrCount()` after `buildExpressionAttributeNames`, but that function doesn't use the global array that is reset. I guess you wanted to call it after `buildFilterExpression`.

Anyway, I think that using a global array (to this file) and knowing when I should reset it (it is not clear) is dangerous and can lead to bugs. Let me make you a proposal:

Instead of having a global array and resetting it after calling each of these two functions, _Why not creating two different empty arrays and passing them as parameters (this wouldn't be weird as the functions are recursive, so it is a common pattern to pass initial values as parameters)?_

The code would end up like this:
```suggestion
  const expresion = buildFilterExpression(filters, [])
  const names = buildExpressionAttributeNames(filters)
  const values = buildExpressionAttributeValues(filters, [])
```

This would erase possible subtle bugs when modifying this code and misplacing those "reset" calls or forget to add them, as it is not clear why we need to call them right here.

Check the other comments where I give hints of the other needed changes for this proposal

packages/framework-provider-aws/src/library/searcher-adapter.ts
```diff
@@ -32,62 +31,150 @@ export async function searchReadModel(
...
+  return {
+    FilterExpression: expresion,
+    ExpressionAttributeNames: names,
+    ExpressionAttributeValues: values,
+  }
+}
+
+function buildFilterExpression(filters: FilterFor<any>): string {
```
alvaroloes, line 47: **Related to the proposal to delete the global `attrCount` array:**

```suggestion
function buildFilterExpression(filters: FilterFor<any>, usedPlaceholders: Array<string>): string {
```

Also, this parameter should be passed to the different recursive calls we have here

packages/framework-provider-aws/src/library/searcher-adapter.ts
```diff
@@ -32,62 +31,150 @@ export async function searchReadModel(
...
+      switch (propName) {
+        case 'not':
+          return `NOT (${buildFilterExpression(filter as FilterFor<any>)})`
+        case 'and':
+        case 'or':
+          return `(${(filter as Array<FilterFor<any>>)
+            .map((arrayFilter) => buildFilterExpression(arrayFilter))
+            .join(` ${propName} `)})`
+        default:
+          return buildOperation(propName, filter)
```
alvaroloes, line 59: **Related to the proposal to delete the global `attrCount` array:**

```suggestion
          return buildOperation(propName, filter, usedPlaceholders)
```

packages/framework-provider-aws/src/library/searcher-adapter.ts
```diff
@@ -32,62 +31,150 @@ export async function searchReadModel(
...
+      }
+    })
     .join(' AND ')
 }
 
-function buildOperation(propName: string, filter: Filter<any>): string {
+function buildOperation(propName: string, filter: Operation<any> = {}): string {
   const holder = placeholderBuilderFor(propName)
```
alvaroloes, line 66: **Related to the proposal to delete the global `attrCount` array:**
```suggestion
  const holder = placeholderBuilderFor(propName, usedPlaceholders)
```

packages/framework-provider-aws/src/library/searcher-adapter.ts
```diff
@@ -32,62 +31,150 @@ export async function searchReadModel(
...
+    })
+    .join(' AND ')
 }
 
-function placeholderBuilderFor(propName: string): (valueIndex: number) => string {
-  return (valueIndex: number) => `:${propName}_${valueIndex}`
+const attrCount: string[] = []
+function resetAttrCount(): void {
+  attrCount.length = 0
+}
```
alvaroloes, line 105: **Related to the proposal to delete the global `attrCount` array:**

This code would not be needed anymore

packages/framework-provider-aws/src/library/searcher-adapter.ts
```diff
@@ -32,62 +31,150 @@ export async function searchReadModel(
...
+function placeholderBuilderFor(propName: string): (valueIndex: number, valueSubIndex?: number) => string {
+  return (valueIndex: number, valueSubIndex?: number) => {
+    const placeholder = `:${propName}_${valueIndex}` + (typeof valueSubIndex === 'number' ? `_${valueSubIndex}` : '')
+    if (attrCount.includes(placeholder)) return placeholderBuilderFor(propName)(valueIndex + 1)
+    attrCount.push(placeholder)
+    return placeholder
+  }
 }
```
alvaroloes, line 113: **Related to the proposal to delete the global `attrCount` array:**

Just using the parameter, instead of the global array:

```suggestion
function placeholderBuilderFor(propName: string): (valueIndex: number, valueSubIndex?: number, usedPlaceholders: Array<string>) => string {
  return (valueIndex: number, valueSubIndex?: number) => {
    const placeholder = `:${propName}_${valueIndex}` + (typeof valueSubIndex === 'number' ? `_${valueSubIndex}` : '')
    if (usedPlaceholders.includes(placeholder)) return placeholderBuilderFor(propName, undefined, usedPlaceholders)(valueIndex + 1)
    usedPlaceholders.push(placeholder)
    return placeholder
  }
}
```

packages/framework-provider-aws/src/library/searcher-adapter.ts
```diff
@@ -32,62 +31,150 @@ export async function searchReadModel(
...
+          }
+        })
+        break
+    }
   }
   return attributeNames
 }
 
-function buildExpressionAttributeValues(filters: Record<string, Filter<any>>): ExpressionAttributeValueMap {
+function buildExpressionAttributeValues(filters: FilterFor<any>): ExpressionAttributeValueMap {
```
alvaroloes, line 143: Do the same changes that we did in "buildFilterExpression" regarding removing the attrCoung global array

packages/framework-provider-aws/test/library/searcher-adapter.test.ts
alvaroloes: 1000% agree. Now is the time to document this thoroughly

packages/framework-provider-aws/test/library/searcher-adapter.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      const filters: FilterFor<any> = {
+        id: { [unknownOperator]: 'test' },
+      }
+
+      await expect(searchReadModel(database, config, logger, readModelName, filters)).to.be.eventually.rejectedWith(
+        `Operator "${unknownOperator}" is not supported`
+      )
+    })
+  })
+})
```
alvaroloes, line 202: LOVE THE TESTS!

packages/framework-provider-local/src/library/searcher-adapter.ts
alvaroloes: I think he did that to avoid the need of changing all the providers at once to the new filters. So he left the old filters in place and used only the new ones in AWS

packages/framework-types/src/envelope.ts
```diff
@@ -42,11 +43,7 @@ export interface ReadModelRequestEnvelope extends Envelope {
   filters?: Record<string, ReadModelPropertyFilter>
 }
 
-export interface ReadModelPropertyFilter {
-  operation: string
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  values: Array<any>
-}
+export type ReadModelPropertyFilter = FilterFor<any>
```
alvaroloes, line 46: I'm not sure if it won't give you problems, but Could you please try to change this to `FilterFor<unknown>`? This way we avoid `any` which is a bit dangerous in Typescript.

If you find too many problems after the change, forget it

alvaroloes (commented):
Left some commange @adriangmweb Thanks a lot for this!!

packages/framework-provider-local/src/library/searcher-adapter.ts
javiertoledo: 🤔 and that's why we need to move the core features to separate rockets too...

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
```diff
@@ -28,41 +29,43 @@ function filterReadModel(readModel: Record<string, any>, filters?: Record<string
...
+  }
+  return true
+}
 
+function filterByOperation(filter: ReadModelPropertyFilter, readModelPropValue: any): boolean {
+  for (const [operation, value] of Object.entries(filter as Operation<any>)) {
     switch (operation) {
       case '=':
-        if (readModelPropValue !== values[0]) return false
+        if (readModelPropValue !== value) return false
```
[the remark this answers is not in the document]
javiertoledo: Oh, I didn't notice the `for` at the beginning of the function. Still, you could extract the switch to a new function and apply the suggested changes to make this code a bit more readable. Anyway, this was just a suggestion, @adriangmweb decides!

docs/README.md
```diff
@@ -1565,6 +1568,40 @@ For more information about queries and how to use them, please check the [GraphQ
...
+        },
+      ],
+    })
+
+    const result = await searcher.search()
+    return { count: result.length }
+  }
+}
+```
+
```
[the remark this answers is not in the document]
adriangmweb: Completely agree with this!

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
adriangmweb: This won't actually be able to test until we got the GraphQL new filter working, right? I'll create the issue to solve it along with the new schema

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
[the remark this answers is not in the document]
adriangmweb: I'll be working on refactoring this file once I start working on the new GraphQL schema so it can be properly tested. I created an issue for this https://github.com/boostercloud/booster/issues/600

packages/framework-provider-aws/src/library/searcher-adapter.ts
adriangmweb: This is because when there is an includes operation it shouldn't enter in the `default` case but get ingnored instead.

packages/framework-provider-aws/test/library/searcher-adapter.test.ts
adriangmweb: I think it would make more sense to document it on the GraphQL new schema branch.

docs/README.md
adriangmweb: I haven't change anything but if I'm the only one experiencing this, I'll love to compare our rules to find what I'm missing 😅

packages/framework-provider-aws/src/library/searcher-adapter.ts
adriangmweb: This is awesome @alvaroloes! 💯  I just implemented it and even improve it a little, I'll show you shortly!

packages/framework-provider-aws/src/library/searcher-adapter.ts
javiertoledo: Then I'd add a comment explaining that or someone could end up removing that code thinking that's useless..

packages/framework-provider-aws/test/library/searcher-adapter.test.ts
javiertoledo: Well, you have to document two things: how you use it from a handler and how you'll use it when it's available from the API. I'd add the docs on each PR to make sure it's not forgotten later

adriangmweb: Wouldn't be enough with the current `Getting and filtering read models data at code level` section I added on the docs?

packages/framework-types/src/searcher.ts
adriangmweb: Because it's still used in the rest of the providers, so it's renamed in order to be able to make them know it's an old and deprecated functionality. I just created a new issue explaining that we should migrate all providers to the new system before removing it. https://github.com/boostercloud/booster/issues/601

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
adriangmweb: Here's the issue https://github.com/boostercloud/booster/issues/602

javiertoledo (approved):
Approving, as all my concerns have been either addressed or moved to new issues. Please make sure that documentation is addressed in the short term, or this feature won't be used.

davidverdu (approved):
LGTM, provided that some documentation is added in the future. These are amazing changes! (nonetheless, there are things that appear to be complex for me at first glance)

alvaroloes (commented):
I have now realized that I had this review pending to submit from the previous week 🤦

Issue #523, opened 2020-12-21
Unable to get the GraphQL API schema


rdiaz82:
# Bug Report

## Current Behavior

GraphQL schema is not working. In any graphQL client if you try to get the schema you get an error 405 (method not allowed). In spite of this error, the GRaphQL API is working because I have checked and I was able to perform a mutation and a query without any problems.

## Expected behavior

The expected behavior is to get the graphQL API schema.

## Possible Solution

Checking in the API gateway I have seen that the API only has enabled the POST method, So I guess that the problem could be there. During the deployment, the cloud formation template probably is not setting the API gateway for GET operations. 
![Screenshot 2020-12-21 at 23 50 35](https://user-images.githubusercontent.com/5313797/102832912-a0403180-43e7-11eb-919e-318199993bb7.png)
### Environment

- Booster version: 0.9.1
- Node/npm version: node 12.16.1 / npm 6.13.4
- OS: OSX 10.15.3

javiertoledo:
@alvaroloes what do you think about it? This is either a bug or we need to clarify this in the documentation...

rdiaz82:
I would say that this was working in the past, I wouldn't put my hand over the fire but I have seen schemas in Postwoman.

The only thing that I doubt is If I saw that schema using the kubernetes provider  or using the aws provider.

NickSeagull:
Yes, this was working with AWS, in fact I was using it last month 😱

rdiaz82:
could happen that it is fixed because @davidverdu I think that has been able to get it. @davidverdu could you confirm that this is working?

davidverdu:
@rdiaz82 I haven't touched anything related to GraphQL. Nonetheless, I'm about to refactor the GraphQLGenerator class as is proposed in this issue (https://github.com/boostercloud/booster/issues/541). If there is an error maybe I can try to solve it, but I need more context to do that.

javiertoledo:
AFAIK when you connect to your API using Postwoman or other GraphQL client, you can see the schema. Are we sure that we want to be able to get the schema using an actual `GET` operation? I'm not sure this is standard TBH

NickSeagull:
Schemas are retrieved by `POST` ing the `IntrospectQuery` GraphQL query, which is a special GraphQL query that returns the schema, I don't think it should be `GET`ted

davidverdu:
Definitely it works with `POST`. I've found that the bug occurs when no readModel is defined in your project. I'm working on the FIX

PR #575, opened 2021-01-22
Fix/523 unable to get graphql schema

davidverdu:
## Description

This is a bug fix that solves the problem that occurs when user deploys a project without ReadModels or without Commands. In those cases the generated GraphQL schema contained validation errors and the Lambda crashed. Now the system behaves this way:
- when there is no commands, the system returns no mutations
- When there is no read model, the system returns a _dummy query. It is mandatory for a GraphQL schema to always contain a query

Fixes: #523 

## Changes
- I updated the framework-core to take into account these edge cases
- I updated framework-core unit tests

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-core/src/services/graphql/graphql-generator.ts
+++ b/packages/framework-core/src/services/graphql/graphql-generator.ts
@@ -16,9 +16,9 @@ import { BoosterReadModelDispatcher } from '../../booster-read-model-dispatcher'
 import { GraphQLResolverContext } from './common'
 
 export class GraphQLGenerator {
-  private readonly queryGenerator: GraphQLQueryGenerator
+  private readonly queryGenerator: GraphQLQueryGenerator | undefined
   private readonly mutationGenerator: GraphQLMutationGenerator
-  private readonly subscriptionGenerator: GraphQLSubscriptionGenerator
+  private readonly subscriptionGenerator: GraphQLSubscriptionGenerator | undefined
   private readonly typeInformer: GraphQLTypeInformer
 
   private static singleton: GraphQLGenerator | undefined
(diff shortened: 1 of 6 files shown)


packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -37,12 +38,18 @@ export class GraphQLQueryGenerator {
...
     const filterQueries = this.generateFilterQueries()
+    const fields = {...byIDQueries, ...filterQueries}
+    if (Object.keys(fields).length === 0) {
+      return new GraphQLObjectType({
+        name: 'Query',
+        fields: {
+          _dummy: { type: GraphQLString }
+        }
+      })
+    }
```
davidverdu, line 49: It is mandatory to always return a query, even though is a silly query. Empty queries violates GraphQL validations

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -55,7 +55,7 @@ export class GraphQLGenerator {
       this.queryGenerator,
       this.subscriptionByIDResolverBuilder.bind(this),
       this.subscriptionResolverBuilder.bind(this)
-    )
+    )
```
NickSeagull, line 58: Note these spaces, remember to always run `check-all-the-things` before submitting the PR so the linter fixes these things

packages/framework-core/src/services/graphql/graphql-mutation-generator.ts
```diff
@@ -9,8 +9,11 @@ export class GraphQLMutationGenerator {
     private readonly mutationResolver: ResolverBuilder
   ) {}
 
-  public generate(): GraphQLObjectType {
+  public generate(): GraphQLObjectType | undefined {
```
NickSeagull, line 12: Why do we have to return `undefined`?

packages/framework-core/src/services/graphql/graphql-query-generator.ts
NickSeagull: Perhaps we can give it a meaningful name? Like `NoQueriesDefined`?

packages/framework-core/src/services/graphql/graphql-generator.ts
davidverdu: OK I'll check it

packages/framework-core/src/services/graphql/graphql-query-generator.ts
davidverdu: I'll change its name accordingly

packages/framework-core/src/services/graphql/graphql-mutation-generator.ts
davidverdu: If we return a GraphQLObjectType with name 'Mutation' with empty fields, GraphQL library throws a validation error. By sending undefined to the GraphQLGenerator it creates the schema without mutations and without triggering a validation error.

davidverdu:
@NickSeagull I applied your suggested changes. Could you check it again?

alvaroloes (approved):

NickSeagull (approved):
Thank you a lot! 🙏

PR #579, opened 2021-01-26
Feature: Remove Auth Infra From Core

charlietfe:
## Description

In this PR, I've removed all the auth related code from the core and now we don't have any dependency on Cognito. 

## Changes

- Modified the Booster Config to accept a `publicKey` property with the value of the public certificate key, instead of the `jwks.json` file used in the `jwksUri`. This will avoid hitting an endpoint during tests. All the tests now are self contained. 
- Created the function `getTokenForUser`  inside `utils.ts` to generate tokens for a user with a specific role. This function will use the private key file to sign the tokens.
- Removed all the auth tests which depend on Cognito and authentication endpoints
- Removed all Cognito helpers in `utils.ts`.
- Modified test project config to use the new `publicKey` property in the tokenVerifier configuration, which uses a local public certificate included in the `assets/certs` folder. 

```
  /* We use an automatically generated app name suffix to allow
   * running integration tests for different branches concurrently.
   */
  const appNameSuffix = process.env.BOOSTER_APP_SUFFIX ?? 'default'

  // The app suffix must be copied to the test app lambdas
  config.env['BOOSTER_APP_SUFFIX'] = appNameSuffix

  config.appName = 'my-store-' + appNameSuffix
  config.provider = AWS.Provider()
  config.assets = ['assets']
  config.tokenVerifier = {
    issuer: 'booster',
  }
})
```

## Checks
- [X] Project Builds


packages/framework-integration-tests/integration/providers/aws/utils.ts
```diff
@@ -597,3 +597,26 @@ export async function waitForIt<TResult>(
     return doWaitFor()
   }
 }
+
+// This helper will create a valid token using a real private key for testing
+// the tokens will be validate against the public keyset uri
+// located in: https://booster-integration-tests.s3.amazonaws.com/.well-known/jkws.json
+export const getTokenForUser = (email: string, role: string): string => {
```
adriangmweb, line 604: Awesome! ❤️

adriangmweb (approved):
Thanks, it's a huge change 🙏🏻.

I'll lock/unlock the conversation to run integration tests

NickSeagull (approved):
Lovely!

juanjoman (approved):
Amazing 🙌

packages/framework-integration-tests/src/config/config.ts
```diff
@@ -28,4 +28,8 @@ Booster.configure('production', (config: BoosterConfig): void => {
 
   config.appName = 'my-store-' + appNameSuffix
   config.provider = AWS.Provider()
+  config.tokenVerifier = {
+    jwksUri: 'https://booster-integration-tests.s3.amazonaws.com/.well-known/jkws.json',
```
javiertoledo, line 32: Depending on an external resource in a hardcoded URL scares me... maybe we could generate it dynamically, it's a little more work now, but it could save a lot of headaches in the future...

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/api-stack-velocity-templates.ts
```diff
@@ -1,79 +0,0 @@
...
-               }`,
-  },
-  signOut: {
-    request: `#set($root = $input.path('$'))
-              {
-                "AccessToken":"$root.accessToken" 
-              }`,
-    response: '{}',
-  },
-} as const
```
javiertoledo, line 79: Happy to get rid of these!

javiertoledo (approved):
Just left a few comments

.github/workflows/integration-tests.yml
```diff
@@ -55,6 +56,7 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.BOT_AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.BOT_AWS_SECRET_ACCESS_KEY }}
           AWS_REGION: us-east-1
+          BOOSTER_RSA_PRIVATE_KEY: ${{ secrets.BOOSTER_RSA_PRIVATE_KEY }}
```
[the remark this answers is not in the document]
alvaroloes: I agree. If not possible, at least I would change the name to leave it clear  that it is only used for integration tests

alvaroloes (approved):
Approved, but please check first Javi's comments.

Oh and BTW:

image
image

packages/framework-integration-tests/src/config/config.ts
```diff
@@ -29,5 +31,10 @@ Booster.configure('production', (config: BoosterConfig): void => {
 
   config.appName = 'my-store-' + appNameSuffix
   config.provider = AWS.Provider()
-  config.assets = ['assets', 'assetFile.txt']
+  config.assets = ['assets']
+  config.tokenVerifier = {
+    issuer: 'booster',
+    // Read the content of the public RS256 cert, used to sign the JWT tokens
+    publicKey: fs.readFileSync(path.join(__dirname, '..', '..', 'assets', 'certs', 'public.key'), 'utf8'),
```
charlietfe, line 38: This is the reck to perform the tests without hitting any endpoint. Even more if your auth provider doesn't provides  you a jwksUri, and instead provides you the public cert you could use it as well. Both options are available.

PR #581, opened 2021-01-27
feature: Print the full GraphQL endpoint after a successful deploy

javiertoledo:
## Description

I've reopened PR #563 here to fix a bad rebase, there are no code changes.

## Checks
- [x] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -409,12 +409,8 @@ Additionally, you can use the `--skipInstall` flag if you want to skip installin
 > Let's break down the command we have just executed:
 >
 > - `boost` is the Booster CLI executable
-> - `new:project` is the "subcommand" part. In this case, it is composed of two parts separated by a colon.
-        >   The first part, `new`, means that we want to generate a new resource. The second part, `project`, indicates which
-        >   kind of resource we are interested in. Other examples are `new:command`, `new:event`, etc. We'll see a bunch of them later.
-> - `boosted-blog` is a "parameter" for the subcommand `new:project`. Flags and parameters are optional and
-        >   their meaning and shape depend on the subcommand you used. In this case, we are specifying the name of the project
-        >   we are creating.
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -395,28 +395,28 @@ If you prefer to specify each parameter without following the instructions, you
...
 
 > Booster CLI commands follow this structure: `boost <subcommand> [<flags>] [<parameters>]`.
 > Let's break down the command we have just executed:
 >
 > - `boost` is the Booster CLI executable
 > - `new:project` is the "subcommand" part. In this case, it is composed of two parts separated by a colon.
->   The first part, `new`, means that we want to generate a new resource. The second part, `project`, indicates which
->   kind of resource we are interested in. Other examples are `new:command`, `new:event`, etc. We'll see a bunch of them later.
+    >   The first part, `new`, means that we want to generate a new resource. The second part, `project`, indicates which
+    >   kind of resource we are interested in. Other examples are `new:command`, `new:event`, etc. We'll see a bunch of them later.
```
javiertoledo, line 414: It looks like there's a small indentation issue here that makes these greater than symbols render in the page:

Screen Shot 2021-01-28 at 20 11 46

Here's a one-click patch:

```suggestion
>   The first part, `new`, means that we want to generate a new resource. The second part, `project`, indicates which
>   kind of resource we are interested in. Other examples are `new:command`, `new:event`, etc. We'll see a bunch of them later.
```

Another option could be deleting the newlines to join lines 413, 414, and 412; they're rendered as a single paragraph anyway, but applying this patch should be fine too.

docs/README.md
```diff
@@ -395,28 +395,28 @@ If you prefer to specify each parameter without following the instructions, you
...
 > - `boosted-blog` is a "parameter" for the subcommand `new:project`. Flags and parameters are optional and
->   their meaning and shape depend on the subcommand you used. In this case, we are specifying the name of the project
->   we are creating.
+    >   their meaning and shape depend on the subcommand you used. In this case, we are specifying the name of the project
+    >   we are creating.
```
javiertoledo, line 417: Same issue here

```suggestion
>   their meaning and shape depend on the subcommand you used. In this case, we are specifying the name of the project
>   we are creating.
```

javiertoledo (commented):
Good work @Moneyba, just take a look to the comments to fix a minor indentation issue.

PR #583, opened 2021-01-28
Make Booster config public

juanjoman:
## Description
Creating a getter for the configuration to use methods like `readModelNameFromResourceName` from command handlers, event handlers, etc.

## Changes
Created a `getConfiguration` method so we can do things like this on handlers:

`const config = Booster.getConfiguration()`
`const myTableName = config.readModelNameFromResourceName('myTable')`
`sdk.dynamoDB.describeTable({ mytableName }, () => { ... })`

## Checks
- [X] Project Builds
- [] Updated documentation accordingly ->> Not sure if this should be documented

### DIFF
--- a/packages/framework-core/src/booster.ts
+++ b/packages/framework-core/src/booster.ts
@@ -37,6 +37,10 @@ export class Booster {
     configurator(this.config)
   }
 
+  public static getConfiguration(): BoosterConfig {
+    return this.config
+  }
+
   /**
    * Allows to configure the Booster project.
    *
--- a/packages/framework-core/src/booster.ts
+++ b/packages/framework-core/src/booster.ts
@@ -27,7 +27,7 @@ import { BoosterScheduledCommandDispatcher } from './booster-scheduled-command-d
 export class Booster {
   public static readonly configuredEnvironments: Set<string> = new Set<string>()
   private static logger: Logger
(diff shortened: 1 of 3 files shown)


alvaroloes (commented):
@juanjoman Now that I see this, I think it would be much simpler if we just make the `config` field "public" and "readonly" (it already is readonly), which exactly equivalent to what you have done, but this way it is Typescript the one which creates the getter for you :smiley:

packages/framework-core/src/booster.ts
```diff
@@ -37,6 +37,10 @@ export class Booster {
     configurator(this.config)
   }
 
+  public static getConfiguration(): BoosterConfig {
+    return this.config
+  }
+
```
alvaroloes, line 43: What I meant is to remove this getter, and make the `config` getter like this:
```typescript
  *public* static readonly config = new BoosterConfig(checkAndGetCurrentEnv())
```

juanjoman: done!!

alvaroloes (approved):
🎉 :shipit:

charlietfe (approved):
:shipit:

NickSeagull (approved):
Thank you a lot! 🙏

PR #574, opened 2021-01-22
Backup rocket

juanjoman:
## Description
This rocket allows the user to backup all DynamoDB tables from your Booster Framework project. With the current implementation, you can set both on-demand or point-in-time recovery, depending on your needs.

### On-demand
More info: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/BackupRestore.html

We use [AWS Backup](https://aws.amazon.com/backup/?nc1=h_ls) to manage DynamoDB on-demand backups. This is because on-demand backups are not configurable through the DynamoDB CDK.

**Disclaimer:** I let this service configured with a very basic setup. In future versions, we can add much more detailed features like encryption, cold storage management, different retention times, etc. 

### Point-in-time recovery
More info: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/PointInTimeRecovery.html

We modify the DynamoDB table properties and set the [pointInTimeRecovery](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-dynamodb.Table.html#pointintimerecovery) param to true.

--- 
**Note:** Point-in-time recovery option is the best one to [export data to S3](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DataExport.HowItWorks.html). In case that we wanted to do this with the on-demand option, we'd have to implement additional features. Also, and at least for now, the `export to S3` feature can only be done through [AWS SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB.html#exportTableToPointInTime-property), [AWS CLI, or AWS Console](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DataExport.Requesting.html).

**Note 2:** Just in case you're interested, these are the [recommended practices from AWS to export DynamoDB Tables to S3](https://aws.amazon.com/premiumsupport/knowledge-center/back-up-dynamodb-s3/).

---
## Checks
- [X] Project Builds
- [X] Project passes tests and checks
- [X] Updated documentation accordingly
- [x] Manually tested

PR #586, opened 2021-01-29
Backup rocket

juanjoman:
## Description
This rocket allows the user to backup all DynamoDB tables from your Booster Framework project. With the current implementation, you can set both on-demand or point-in-time recovery, depending on your needs.

### On-demand
More info: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/BackupRestore.html

We use [AWS Backup](https://aws.amazon.com/backup/?nc1=h_ls) to manage DynamoDB on-demand backups. This is because on-demand backups are not configurable through the DynamoDB CDK.

**Disclaimer:** I let this service configured with a very basic setup. In future versions, we can add much more detailed features like encryption, cold storage management, different retention times, etc. 

### Point-in-time recovery
More info: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/PointInTimeRecovery.html

We modify the DynamoDB table properties and set the [pointInTimeRecovery](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-dynamodb.Table.html#pointintimerecovery) param to true.

--- 
**Note:** Point-in-time recovery option is the best one to [export data to S3](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DataExport.HowItWorks.html). In case that we wanted to do this with the on-demand option, we'd have to implement additional features. Also, and at least for now, the `export to S3` feature can only be done through [AWS SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB.html#exportTableToPointInTime-property), [AWS CLI, or AWS Console](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DataExport.Requesting.html).

**Note 2:** Just in case you're interested, these are the [recommended practices from AWS to export DynamoDB Tables to S3](https://aws.amazon.com/premiumsupport/knowledge-center/back-up-dynamodb-s3/).

---
## Checks
- [X] Project Builds
- [x] Manually tested


juanjoman:
Merging since it was already approved here: https://github.com/boostercloud/booster/pull/574

PR #589, opened 2021-01-29
Revert "Backup rocket"

juanjoman:
Creating this just in case we can't solve the integration test issue in master (taking a look)


juanjoman:
It finally worked, closing

Issue #564, opened 2021-01-20
Implement boost clean cli command


davidverdu:
## Description

This command will clean the project build files, as if you run `rimraf dist/`. That way we can avoid requiring rimraf as a global dependency

## Possible Solution

Create a new command class in the cli package that runs that operation in the folder

PR #604, opened 2021-02-05
Feature/565&564 boost build and boost clean

### LINKED ISSUE #565: Implement boost build cli command
## Description

The command `boost build` should build the project without deploying the project in the cloud. That way, the programmer can see if the project compiles correctly.

## Possible Solution

Create a new Command class in the cli package

davidverdu:
## Description
I have created the `boost build` command, which compiles the entire project without deploying. This will allow programmers to check if the code compiles without deploying to any environment. 
I have also created the `boost clean` command, which cleans the built project.

Closes #565 and #564 

## Changes
- I created the Build Command class and its unit and integration tests
- I created the Clean Command class and its unit and integration tests

## Checks
- [x] Project Builds

### DIFF
--- a/packages/cli/src/services/config-service.ts
+++ b/packages/cli/src/services/config-service.ts
@@ -26,13 +26,7 @@ export async function compileProjectAndLoadConfig(userProjectPath: string): Prom
   return readProjectConfig(userProjectPath)
 }
 
-export async function checkAndCompileProject(userProjectPath: string): Promise<BoosterConfig> {
-  await checkItIsABoosterProject(userProjectPath)
-  await compileProject(userProjectPath)
-  return new BoosterConfig('build')
-}
-
-async function compileProject(projectPath: string): Promise<void> {
+export async function compileProject(projectPath: string): Promise<void> {
   try {
     await exec('npm run clean && npm run compile', { cwd: projectPath })
   } catch (e) {
--- a/packages/cli/src/services/config-service.ts
+++ b/packages/cli/src/services/config-service.ts
@@ -26,6 +26,12 @@ export async function compileProjectAndLoadConfig(userProjectPath: string): Prom
   return readProjectConfig(userProjectPath)
 }
 
+export async function checkAndCompileProject(userProjectPath: string): Promise<BoosterConfig> {
+  await checkItIsABoosterProject(userProjectPath)
+  await compileProject(userProjectPath)
+  return new BoosterConfig('build')
+}
+
 async function compileProject(projectPath: string): Promise<void> {
   try {
     await exec('npm run clean && npm run compile', { cwd: projectPath })
--- a/packages/cli/src/services/config-service.ts
+++ b/packages/cli/src/services/config-service.ts
@@ -34,6 +34,14 @@ export async function compileProject(projectPath: string): Promise<void> {
   }
 }
 
+export async function cleanProject(projectPath: string): Promise<void> {
+  try {
+    await exec('npm run clean', { cwd: projectPath })
+  } catch (e) {
+    throw wrapExecError(e, 'Error cleaning project')
+  }
+}
+
 function readProjectConfig(userProjectPath: string): Promise<BoosterConfig> {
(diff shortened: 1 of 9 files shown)


davidverdu:
@boostercloud/booster-core  I don't know if you prefer `boost compile` in place of `boost build`. I'm open to suggestions.

@alvaroloes I build the project in its folder. I'm not sure If I should do it in a .build sandbox instead, as in deployments. What's your view on this doubts I have?

packages/cli/src/services/config-service.ts
```diff
@@ -26,6 +26,12 @@ export async function compileProjectAndLoadConfig(userProjectPath: string): Prom
   return readProjectConfig(userProjectPath)
 }
 
+export async function checkAndCompileProject(userProjectPath: string): Promise<BoosterConfig> {
+  await checkItIsABoosterProject(userProjectPath)
+  await compileProject(userProjectPath)
+  return new BoosterConfig('build')
```
javiertoledo, line 32: I think you were forced to return this config object here because of the current interface, but it's weird that you need to do this. I think that you can take advantage from the abstract implementation of the `Script` class to do something like this:

```typescript
Script.init(`boost ${Brand.dangerize('build')} 🚀`, Promise.resolve(process.cwd()))
  .step('Checking project structure', checkItIsABoosterProject)
  .step('Building project', compileProject)
  .info('Build complete!')
  .done()
```

In this case, the type `TContext` will be resolved as `string` and the Script object will be resolved as `Script<string>`, making the action types in the `checkItIsABoosterProject` and `compileProject` match (`(ctx: string) => Promise<void>`. Doing this (and if it works, because I've not tried) you won't need this extra `checkAndCompileProject` function and the final solution will look cleaner. If you try, let me know if you liked this alternate solution and if it worked!

javiertoledo (dismissed):
Nice change! Just left a suggestion for you to consider.

javiertoledo (changes requested):
I changed my review because I'd also add an integration test in this PR before merging it.

packages/cli/src/services/config-service.ts
davidverdu: OK @javiertoledo . The way you suggest is more convenient

davidverdu:
@javiertoledo I added the integration tests and applied your suggested changes. Could you review it again? (I also did a rebase)

juanjoman (approved):
LGTM!! Nice command 🚀

davidverdu:
Hi @javiertoledo @juanjoman ! I added the `boost clean` command here, since it was easier than creating a new PR because clean depends on build. Could you check this change too?

packages/cli/src/services/config-service.ts
```diff
@@ -26,14 +26,22 @@ export async function compileProjectAndLoadConfig(userProjectPath: string): Prom
   return readProjectConfig(userProjectPath)
 }
 
-async function compileProject(projectPath: string): Promise<void> {
+export async function compileProject(projectPath: string): Promise<void> {
   try {
     await exec('npm run clean && npm run compile', { cwd: projectPath })
```
juanjoman, line 31: Now that we are adding the `boost clean` command, do we need the `npm run clean` here too? 😮

juanjoman (approved):
LGTM! Left a comment below :v:

packages/cli/src/services/config-service.ts
davidverdu: @juanjoman I think the `npm run clean` is run before compile just in case, to avoid compilation issues. Compile was implemented that way when I started with the project. `boost clean` is just a nice to have but not necessary as you mentioned.

javiertoledo (approved):
The tests look great! I'm just missing a paragraph for the new `boost clean` command in the documentation.

packages/cli/src/services/config-service.ts
juanjoman: Awesome, super approved then :P

davidverdu:
I added some notes in the documentation.

PR #606, opened 2021-02-08
Api for reading events

alvaroloes:
****NOTE:** Don't get overwhelmed by the high file count of this PR**. Most of those changes are:
- Minor name changes that were spread to many files 
- Tests, tests and tests!

## Description
This PR creates an API for reading events in GraphQL. There will be two kinds of queries:
- Read events **by entity**: allows you to get events for a specific "entity name" (like `Cart`) and, if desired, with a specific "entity ID" (like `abc-123-56435-etc`)
- Read events **by event type**: allows you to get events with a specific event type

Both queries allow extra time filtering by providing the parameters "from" or "to". These are the two new GraphQL queries in the schema:
```graphql
  eventsByEntity(entity: String!, entityID: ID, from: String, to: String): [EventQueryResponse]
  eventsByType(type: String!, from: String, to: String): [EventQueryResponse]
```

#### Authorization
Each Entity decorator can have a parameter called `authorizeReadEvents` (we expect to have "authorizeWriteEvents" in the future), whose values are exactly the same as the one in Commands or ReadModels `authorize` parameter. Something like:
```typescript
@Entity({
  authorizeReadEvents: ["Admin", "iPhoneApp", "AndroidApp"] // List of roles or "all"
})
class MyEntity { ... }
```

If the `@Entity` has no "authorizeReadEvents" field, then **nobody is allowed to read its events**

So, when you do a call to the new API methods `eventsByEntity` or `eventsByType`, Booster will first check if the events you want to read belongs to an Entity your role has permissions over.

## Extra things fixed in this PR
While doing integration tests, several complex bugs showed up. That's the reason why there is extra code in this PR not completely related to its purpose. I will talk more about those things in inline comments, but as a summary, they were the following:
- We were relying on the `throw` behavior of `Promise.allSettled`, but this method never throws.
- When events are stored by sending concurrent requests, it could be the case that 2 or more of those events ended up with exactly the same sorting key (the timestamp), causing that only one of those events were saved (losing the others)
- Some integration tests perform `COUNT` operations or dynamoDB `scan` operations without having the pagination into account, causing false negatives when there were many events in the result set.

## Checks
- [x] Project Builds
- ~[ ] Updated documentation accordingly~ - Will be worked out in another PR -

## Proposal and discussion
<details>
<summary><em>Click here if you want to read the proposal and the reasons that lead to this PR. Forget if you just want to review</em></summary>
This is the first approach for an API to read events. I have found several challenges that I will list here so that you can understand better the decisions and trade-offs I needed to come up with:
### Authorization: 
I didn't find a practical solution to add authorization options to each event or entity so that a specific user can only read some events or events related to some entities. Reasoning:
- What would happen if you submit a query for reading all the events of an entity and you have permission to read only some of them? Should the query fail? Should we only return those you are allowed to read? 
- What would happen if you submit a query to read all the events of a specific type and you have permission for that, but you don't have permissions set at the entity level?

The answers to those questions weren't too satisfactory so the approach I followed is the following: The developer and set in the configuration which roles are allowed to query events. So depending on whether you are allowed or not, **you can read any event or none**
Eventually and after more thought, I started to like this approach more and more. Let me know what you think

### DynamoDB limitations
DynamoDB is an _amazing_ DB engine for the purposes it was created. And flexibility in querying was not one of them 😒 
The events are all stored in a table with a primary key index tuned for the kind of query the framework does: retrieve events based on entity name and entity ID ordered by date. And this goes ultra-fast.
Now we want to query events based on other criteria. There are two ways to do this:
1. Scan the full table and then filter events. But this table is the one containing events for _everything_ that happens in your system. So this option is already discarded
2. Create an index for every possible query the user might do. Performance-wise this is doable, but we can't create indexes for every possible permutation of the fields of the events.

There two more limitations:
- Indexes in DynamoDB are not like the equivalent in the relational DB world, where you can specify any number of fields to be indexed. In Dynamo, you can only specify one field for the partition key and another one for the sorting key.
- When you query an index, you are **forced to use an equality operation over the partition key**. So if your partition key is `eventType`, you can't do a query like `WHERE eventType IN ("typeA", "typeB", "typeC")`, you can only do `WHERE eventType = "typeA"`, for example.

Eventually, I opted for approach number 2 but limiting the index to the most common queries an application can do to read events (for this I was also influenced by the requirements of other projects). So you can query events:
- By type
- By entity
- By entity and entity ID
In all queries, you can also pass a time range to narrow down the results.

I think this approach will suit most use cases. Time will tell.

### GraphQL limitations
My initial idea was to have a new GraphQL query called `events`, and, depending on the parameters you pass, internally we can choose which kind of query the user is doing: either by entity, or by event type. However, due to the lack of support of union **_input_** types ([they will probably be added  in the next GraphQL version](https://github.com/graphql/graphql-spec/pull/733)) I couldn't create such API (I could, but the awesome GraphQL validation would be lost). So I ended up creating two new GraphQL queries `eventsByEntity` and `eventByType`.

It ended up being quite nice: when you do the query, you can't specify an Entity or Event type that doesn't exist in your application (In this regard I LOVE GraphQL)

---------------------- END OF LIMITATIONS -------------------------

After this odyssey, the result is not bad at all and this new API will open a lot of "doors" for may projects!
</details>

- Also, the comments in this PR related to the discussion of the proposal has been hidden with "outdated" reason, so that you can focus on the review. You can always read them by expanding them.


javiertoledo:
I haven't read the code yet, but I want to share some thoughts after reading the detailed PR description. It would have been nice to have the opportunity to see this as a proposal/RFC to discuss about it before seeing a specific implementation, but what I read looks good so far. I'll just leave a few notes for you (and the rest of the team) to consider:

* In a perfect world, a change like this should have been provided as a rocket in preparation to next steps in the framework design. We haven't discussed this deeply yet, but we're currently considering migrating all existing functionality to rockets to make Booster more composable.
* If I understood correctly, you're proposing an all-or-nothing schema that you set globally. That sounds pretty dangerous to me, and having to set this up separately in the config breaks the framework consistency a little bit. I think that you should be able to use the same DSL that we used for Read Models and Commands in Entities and Events. In order to interpret the queries, I think you already has the solution in the way you implemented the queries if you understand permissions as "permissions to use the queries": permissions at event level, allows you to use the `eventsByType` query for that type, and permissions at entity level, allows you to use `eventsByEntity` for that entity. If the developer messes up with the permissions, using the `eventsByEntity` will throw all events, no matter the individual permissions. I think that doing this is understandable and consistent with the semantics, because our auth system is based on the white-listing principle, giving you access to a resource doesn't mean that you don't have access to others, it only means that you have access to that specific resource.
* I actually like having two separate methods to query events via type or entity, I think that it just makes sense!
* Regarding DynamoDB limitations, we've hitting this stone for a while already with Read Model queries. Maybe it's time to start thinking about complementing the current DynamoDB implementation with dedicated indexing services like [Amazon CloudSearch](https://aws.amazon.com/cloudsearch/) or even replicating the DynamoDB tables in a different database technology that is more query-friendly like Aurora Serverless or something like that.
* No need to do this in this PR, but just wanted to mention that while the implementation over GraphQL could make sense to use it from a client application, it could also make sense to provide lower-level REST and/or websocket interfaces that allow polling the event streams or subscribe to new events in a stream.
* Apart from that, and again, not necessarily as part of this PR, I think that it could make a lot of sense to provide a direct endpoint/graphQL method to inject RAW events too, as we're seeing that in many applications, we're creating commands that doesn't do any validation or processing beyond inserting an event with the same command structure.

Anyway, it's good to finally see progress with this functionality, I'll review the code ASAP!

alvaroloes:
Thanks a lot for the comments @javiertoledo !
Let me comment them:

> I haven't read the code yet, but I want to share some thoughts after reading the detailed PR description. It would have been nice to have the opportunity to see this as a proposal/RFC to discuss about it before seeing a specific implementation, but what I read looks good so far. 

Agree here, but when I was thinking about this, I didn't know if anything of what I was thinking was possible, so I tried some implementation to figure things out. It turned out that I needed to change a lot of things because of the limitations I found.
Consider this as an "RFC with a proposed implementation"
> In a perfect world, a change like this should have been provided as a rocket in preparation to next steps in the framework design. We haven't discussed this deeply yet, but we're currently considering migrating all existing functionality to rockets to make Booster more composable.ç

Sure! For me, the whole GraphQL API should be in a rocket. We need to evolve more the rockets handling so that we can add things to the core, runtime, etc.

> If I understood correctly the auth part, (...) I think you already has the solution in the way you implemented the queries if you understand permissions as "permissions to use the queries": permissions at event level, allows you to use the eventsByType query for that type, and permissions at entity level, allows you to use eventsByEntity for that entity. If the developer messes up with the permissions, using the eventsByEntity will throw all events, no matter the individual permissions. I think that doing this is understandable and consistent with the semantics, because our auth system is based on the white-listing principle, giving you access to a resource doesn't mean that you don't have access to others, it only means that you have access to that specific resource.

🤔 I agree we can put more thought in the auth part, but I don't like too much the proposal you are telling because it is kind of "superfluous". What I mean is that if I write, for example:
```typescript
@Event({
   authorize: [Admin]
})
class MyEvent {}
```
It means that only Admins can access the information of that event and should not be visible to other roles (at least that is how it is done with ReadModels and commands), but this is not true. If you do a queryByEntity, you will be able to get all the "MyEvent" events you want. This could be misleading but it can be considered.

There other concerns that I had in mind when decided to follow the current approach: I have analyzed EventStore event queries and some comments from our clients and it is very likely that this event query mechanism will need to be expanded to allow more flexibility in the filters: 

- "get all events with the same requestID" (this would be ultra-useful)
- "get events of type A, B, and C that belongs to entity X with ID=Y"
- "get all events that happened between this and this time"
- 
When we allow those queries (probably by having a copy of the events in a separate DB engine) I would expect to have only one query "events" with filters (this was my initial idea, but due to dynamo limitations I needed to split them). In this scenario, the auth mechanism you are proposing would be really complex to apply.

> I actually like having two separate methods to query events via type or entity, I think that it just makes sense!

Glad you liked it! But take into account that this imposes some restrictions on the queries you can do. For example, you can't get all the events during a period of time (we would need that for other projects). 
Anyway, it is a good starting point.

> Regarding DynamoDB limitations, we've hitting this stone for a while already with Read Model queries. Maybe it's time to start thinking about complementing the current DynamoDB implementation with dedicated indexing services like Amazon CloudSearch or even replicating the DynamoDB tables in a different database technology that is more query-friendly like Aurora Serverless or something like that.

I completely agree to do this with ReadModels, as they are caches that should be optimized for querying (and now we are doing a whole Scan of the table and then filtering, which won't scale in the mid-term).

However, with the table of events, I am not sure how to proceed. The table of events is optimized for its most important use case: Store a huge amount of events and find them ultra quickly by entity ID. DynamoDB is perfect for this thanks to its sharding capabilities.

We could use another service but I'm not sure if it will meet the requirements. Or! We could copy all the events to another service, but I think this could be extremely costly as the table of events will be huge, and will have constant read and write operations.

> No need to do this in this PR, but just wanted to mention that while the implementation over GraphQL could make sense to use it from a client application, it could also make sense to provide lower-level REST and/or websocket interfaces that allow polling the event streams or subscribe to new events in a stream.

👍 . BTW, I have plans to add a subscription too, but in another PR

> Apart from that, and again, not necessarily as part of this PR, I think that it could make a lot of sense to provide a direct endpoint/graphQL method to inject RAW events too, as we're seeing that in many applications, we're creating commands that doesn't do any validation or processing beyond inserting an event with the same command structure.

Sure! Tomas is working on an approach that will allow to do this without any changes to the framework

alvaroloes:
## Alternative partial-solution for the auth mechanism
- Have only one query: `events`
- You _always_ need to specify the entity name
- Then you can specify filters by type, time, etc.

This way, we can put the authorization parameters on each entity and, if you query an entity you are not allowed to, you get an error.

However, this has some limitations:

1. You always need to filter by _one_ entity. So you can't get "all events in a period of time" or "events that contain a specific `requestID`" (which, again, would be one of the most useful queries)
2. Soon we will have support for events not attached to entities (I have seen the need for this several times now). These events will probably have a "topic" or a specific "shardingKey". We would need to be able to get those events too and we can't if we need to specify the name of an entity.

It is a complex topic. The current approach is based on the following assumption:
_The API to read event is meant for special roles, as, normally, applications will work with models. So you give permissions to read those events only to those roles (probably Admins or managers for back-office applications)_
Not ideal, but at least simple to understand.

javiertoledo:
I see clear use cases for the event-level API for non-admin users. One example: Allow a front-end application to react to individual events so they can build their own local read models. Also especially relevant for the events detached from entities, which won't be readable in any other way. Let's see what others think/propose about the auth schema. Id go either for my suggestion or yours, but I think we shouldn't assume this feature is only going to be used by admin-level users, that could introduce unintended security risks IMHO.

alvaroloes:
I agree that assuming only for Admins is not a good approach. Let's see if others provide more feedback. I will also think about other possible approaches. Thanks for all the amazing feedback @javiertoledo !

alvaroloes:
@javiertoledo I have come up with a new proposal that addresses your concerns and mines at once:

### Main point
Have permissions at the entity level, not at the event level. 

### Explanation
Each Entity decorator can have a parameter called `authorizeReadEvents`, whose values are exactly the same as the one in Commands or ReadModels `authorize` parameters (in the future we could have another to "authorizeWriteEvents"). Something like:
```typescript
@Entity({
  authorizeReadEvents: ["Admin", "iPhoneApp", "AndroidApp"] // List of roles or "all"
```

### How this solves all cases

Now, let's see how we can solve each possible case:
1. When a "queryBy**Entity**" is done, we can easily check if the current user has any of the roles listed in that parameter.
2. When a "queryBy**Type**" is done, as _an event type can only belong to a specific entity_ , we can get the associated entity and check the permission as we did in case number 1.
    - If in the future we allow specifying an array of types, we can do this check for each type in the array
3. When we introduce **events associated with no entities**, they would need to be associated to a "topic", which is the same as an Entity but with no reducers nor structure for snapshots. In this case, _the permission configuration would be specified in that topic_
4. If we introduce a query to **read all the events** in the future, we will check that the current user has _permissions to read events for all the entities and topics_ in the app before proceeding with the query.

### Conclusion
So this now solves the two main concerns expressed previously:
- You can control which events are exposed to which roles per entity, not all or nothing.
- It is impossible that you can get events you are not allowed to just by using another query.

javiertoledo:
I think that this is a reasonable initial implementation, but the word `authorizeReadEvents` and then filling it with a list of roles could be a bit confusing. What if we keep using the same `authorize` word for now? We'll document that providing an optional `authorize` parameter to the `Entity` annotation will allow you to query the events of this entity both by Entity or by Type, using one or the other methods provided. Using the same word is more consistent and helps to strengthen the developer's muscle memory!

One caveat of this design is that, it's kind of forcing us to design the topics in a way that requires the final user to declare empty classes to define their topics like:

```typescript
@Topic({authorize: 'all'})
export class Messages {}
```

I don't really know if this is good or bad, because forcing users to declare classes also helps typechecking their projects, but it's a thing to consider 🤷🏼

alvaroloes:
Regarding the `authorizeReadEvents`: I chose that because in the future we will allow to send events directly through the API, and there should be a way to authorize reads and writes separately. In Commands and ReadModels we don't have this problem because one of them is only for writes and another one for reads.

Regarding the topic: Completely agree. I guess it is a trade off (like with roles). When we design the topics, we could try to avoid that in a clever way...

juanjoman (commented):
Just leaving a couple of minor things I found ⬇️ . Nice work!!

packages/framework-core/src/decorators/entity.ts
```diff
@@ -1,22 +1,60 @@
...
  */
-export function Entity<TEntity extends EntityInterface>(entityClass: Class<TEntity>): void {
-  Booster.configureCurrentEnv((config): void => {
-    if (config.entities[entityClass.name]) {
-      throw new Error(`An entity called ${entityClass.name} is already registered
+export function Entity<TEntity extends EntityInterface, TParam extends EntityDecoratorParam>(
+  classOrAttributes: TParam
+): EntityDecoratorResult<TEntity, TParam> {
+  // The craziness with the param and return types is to achieve that either:
+  // - The @Entity decorator doesn't parenthesis: THEN we need to accept a class as parameter and return void
```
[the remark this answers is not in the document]
alvaroloes: Thanks!

packages/framework-core/src/booster-event-dispatcher.ts
```diff
@@ -36,8 +43,8 @@ export class BoosterEventDispatcher {
       switch (eventEnvelope.kind) {
         case 'event':
           logger.debug('[BoosterEventDispatcher#eventProcessor]: Started processing workflow for event:', eventEnvelope)
-          // TODO: Separate into two independent processes the snapshotting/read-model generation process from the event handling process
-          await Promise.allSettled([
+          // TODO: Separate into two independent processes the snapshotting/read-model generation process from the event handling process`
+          await Promises.allSettledAndFulfilled([
```
alvaroloes, line 47: Here we were waiting for a `throw` when something was wrong, but `allSettled` never throws. Check the documentation of `allSettledAndFulfilled` to know more (in this PR)

packages/framework-core/src/booster-events-reader.ts
```diff
@@ -0,0 +1,78 @@
+import {
```
alvaroloes, line 1: This is the brand new class that allow us to read events. It will be used by GraphQL and `Booster.events` to be able to read events

packages/framework-core/src/booster.ts
```diff
@@ -73,6 +75,9 @@ export class Booster {
     return new Searcher(readModelClass, searchFunction)
   }
 
+  public static async events(filters: EventFilter): Promise<Array<EventSearchResponse>> {
+    return this.config.provider.events.search(this.config, this.logger, filters)
+  }
```
alvaroloes, line 80: Allow users to read events from code (from handlers)

packages/framework-core/src/decorators/entity.ts
```diff
@@ -1,22 +1,60 @@
...
-export function Entity<TEntity extends EntityInterface>(entityClass: Class<TEntity>): void {
-  Booster.configureCurrentEnv((config): void => {
-    if (config.entities[entityClass.name]) {
-      throw new Error(`An entity called ${entityClass.name} is already registered
+export function Entity<TEntity extends EntityInterface, TParam extends EntityDecoratorParam>(
+  classOrAttributes: TParam
+): EntityDecoratorResult<TEntity, TParam> {
+  // The craziness with the param and return types is to achieve that either:
+  // - The @Entity decorator doesn't have parenthesis: THEN we need to accept a class as a parameter and return void
+  // - The @Entity decorator have parenthesis: THEN we need to accept an object with attributes and return a function accepting a class as a parameter
```
alvaroloes, line 31: Check this comment. This way, if you don't want to expose an entity, you can just write `@Entity` and not `@Entity()`

packages/framework-core/src/helpers/promise.ts
```diff
@@ -0,0 +1,40 @@
...
+   * rejected. More importantly, **it does not wait for al the promises to finish**, which could lead to undesired behaviors
+   * - `Promise.allSettled`: This method waits for all the promises to finish and then returns an array of results. Some
+   * of them will be fulfilled and some rejected. More importantly, **it never throws an error**, which could lead to
+   * unexpected consequences. For example if you do "await Promise.allSettle(...)" expecting it to throw if some of them
+   * failed, you won't get that.
+   *
+   * In brief, `Promises.allSettledAndFulfilled` behaves exactly the same way as `Promise.allSettle` but it throws with
+   * an array of the failed promises, only if there are any.
+   */
+  public static async allSettledAndFulfilled<TValue>(
```
alvaroloes, line 20: Check the above comment ⬆️

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -29,32 +38,27 @@ export class GraphQLQueryGenerator {
...
     const filterQueries = this.generateFilterQueries()
-    const fields = {...byIDQueries, ...filterQueries}
-    if (Object.keys(fields).length === 0) {
-      return new GraphQLObjectType({
-        name: 'Query',
-        fields: {
-          NoQueriesDefined: { type: GraphQLString }
-        }
-      })
-    }
```
alvaroloes, line 49: Now ther will always be queries defined: the ones to read events

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -83,6 +87,61 @@ export class GraphQLQueryGenerator {
...
   }
 
+  private generateEventQueries(): GraphQLFieldConfigMap<unknown, GraphQLResolverContext> {
+    const eventQueryResponse = this.buildEventQueryResponse()
+    return {
+      eventsByEntity: {
+        type: eventQueryResponse,
+        args: {
+          entity: {
+            type: new GraphQLNonNull(this.buildGraphqlSimpleEnumFor('EntityType', Object.keys(this.config.entities))),
```
alvaroloes, line 97: Thanks to this GraphQL enum inferred from the entities in the config, it will be impossible for the user to specify a non-existing entity name in the query. They will get a GraphQL validation error. One of the beauty of GraphQL

packages/framework-integration-tests/integration/end-to-end/cart.integration.ts
```diff
@@ -1,438 +0,0 @@
-import { ApolloClient } from 'apollo-client'
```
alvaroloes, line 1: I divided this file into three files that test different components:
- Commands
- Entities
- ReadModels
- Events

The only one where I added tests was in the events one.

packages/framework-integration-tests/integration/end-to-end/commands.integration.ts
```diff
@@ -0,0 +1,32 @@
+import { ApolloClient } from 'apollo-client'
```
alvaroloes, line 1: No change in logic. Just copied and pasted it from `cart.integration.ts`

packages/framework-integration-tests/integration/end-to-end/entities.integration.ts
```diff
@@ -0,0 +1,166 @@
+import { ApolloClient } from 'apollo-client'
```
alvaroloes, line 1: No change in logic. Just copied and pasted it from `cart.integration.ts`

packages/framework-integration-tests/integration/end-to-end/read-models.integration.ts
```diff
@@ -0,0 +1,253 @@
+import { ApolloClient } from 'apollo-client'
```
alvaroloes, line 1: No change in logic. Just copied and pasted it from `cart.integration.ts`

packages/framework-integration-tests/integration/end-to-end/events.integration.ts
```diff
@@ -0,0 +1,405 @@
...
+import { internet, random } from 'faker'
+import gql from 'graphql-tag'
+import { expect } from 'chai'
+import * as chai from 'chai'
+import { createPassword } from '../helper/auth-helper'
+import { sleep, waitForIt } from '../helper/sleep'
+import { EventSearchResponse, EventTimeFilter } from '@boostercloud/framework-types'
+chai.use(require('chai-as-promised'))
+
+describe('Events end-to-end tests', () => {
```
alvaroloes, line 13: This contains the full suite of integration tests for the new API added

packages/framework-integration-tests/integration/end-to-end/events.integration.ts
```diff
@@ -0,0 +1,405 @@
...
+              if (event.entityID === mockCartId) {
+                provisionedEventsFound++
+              }
+            }
+            expect(provisionedEventsFound).to.be.gte(numberOfProvisionedEvents)
+          })
+        })
+      })
+    })
+    describe('the result of the queries involving many events', () => {
```
alvaroloes, line 282: In this test, we create 150 events concurrently. It was the one that surfaced the bug that was causing a loss of events. 

This motivated the change in the storage of events so that they now contain a UUID suffix in their sorting key

packages/framework-integration-tests/integration/helper/sleep.ts
```diff
@@ -1,3 +1,31 @@
 export async function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms))
 }
+
+export async function waitForIt<TResult>(
```
alvaroloes, line 5: No logic change, just moved it out of the provider-specific folder so that I can use it in end-to-end tests

packages/framework-integration-tests/integration/providers/aws/utils.ts
```diff
@@ -525,6 +519,11 @@ export async function countTableItems(tableName: string): Promise<number> {
 }
 
 export async function countEventItems(): Promise<number> {
+  // TODO: This way of counting is wrong when there are a lot of events in the event store, as COUNT will only contain
+  // the number of elements in a a result set that is lower than 1MB in size:
+  // "If the size of the Query result set is larger than 1 MB, then ScannedCount and Count
+  // will represent only a partial count of the total items.
+  // You will need to perform multiple Query operations in order to retrieve all of the results\"."
```
alvaroloes, line 526: DYNAMODB....

image

We need to start creating tests with huge loads of events. We will start finding hidden bugs.

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/events-stack.ts
```diff
@@ -45,6 +45,33 @@ export class EventsStack {
...
+      partitionKey: {
+        name: eventsStoreAttributes.indexByEntity.partitionKey,
+        type: dynamodb.AttributeType.STRING,
+      },
+      sortKey: {
+        name: eventsStoreAttributes.sortKey,
+        type: dynamodb.AttributeType.STRING,
+      },
+      projectionType: dynamodb.ProjectionType.KEYS_ONLY,
+    })
```
alvaroloes, line 60: This index allows us to *query by entity*

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/events-stack.ts
```diff
@@ -45,6 +45,33 @@ export class EventsStack {
...
+      partitionKey: {
+        name: eventsStoreAttributes.indexByType.partitionKey,
+        type: dynamodb.AttributeType.STRING,
+      },
+      sortKey: {
+        name: eventsStoreAttributes.sortKey,
+        type: dynamodb.AttributeType.STRING,
+      },
+      projectionType: dynamodb.ProjectionType.KEYS_ONLY,
+    })
```
alvaroloes, line 73: This filter allows to query *by entity type*

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -75,7 +82,9 @@ export async function readEntityLatestSnapshot(
       `[EventsAdapter#readEntityLatestSnapshot] Snapshot found for entity ${entityTypeName} with ID ${entityID}:`,
       snapshot
     )
-    return snapshot as EventEnvelope
+    const foundSnapshot = snapshot as EventEnvelope
+    modifyEventsDecodingSortingKeys(foundSnapshot)
```
alvaroloes, line 86: When we read events (or snapshots), we need to "decode" de sortingKey (createdAt) to remove the suffix we added when storing it

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,250 @@
+import {
+  BoosterConfig,
+  EventEnvelope,
+  EventFilter,
+  EventInterface,
+  EventSearchResponse,
+  Logger,
+  UUID,
+} from '@boostercloud/framework-types'
+import { DynamoDB } from 'aws-sdk'
```
alvaroloes, line 10: This is the adapter that supports all the searching logic for the new events API. 

**This is the core logic in this PR**

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,250 @@
...
+  } else if (to) {
+    timeQueryData = {
+      expression: ` AND ${eventsStoreAttributes.sortKey} <= :toTime`,
+      attributeValues: { ':toTime': to },
+    }
+  }
+  return timeQueryData
+}
+
+async function executeSearch(
```
alvaroloes, line 61: Depending on the type of query (by entity, by entity with an ID, by type), we need to query different indexes

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,250 @@
...
+
+async function searchEventsByEntityAndID(
+  dynamoDB: DynamoDB.DocumentClient,
+  config: BoosterConfig,
+  logger: Logger,
+  entity: string,
+  entityID: UUID,
+  timeQuery: TimeQueryData
+): Promise<Array<EventEnvelope>> {
+  // TODO: Manage pagination
```
alvaroloes, line 96: TODO: we need to add pagination to all these queries. @NickSeagull is working on a way to do it

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,250 @@
...
+    ExpressionAttributeValues: {
+      ...timeQuery.attributeValues,
+      ':partitionKey': type,
+    },
+  }
+
+  logger.debug('Searching events by type. Index query params: ', params)
+  const partialResult = await dynamoDB.query(params).promise()
+  const indexRecords = (partialResult.Items as Array<EventStoreKeys>) ?? []
+  return findEventsDataWithKeys(dynamoDB, config, logger, indexRecords)
```
alvaroloes, line 167: Once we have queried the index, we got the primary keys of the table. Now we need to query the main table with those keys

NickSeagull:
@alvaroloes can you create an issue for the pagination of events and add it to #636 please?

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,250 @@
...
+  logger: Logger,
+  keys: Array<EventStoreKeys>
+): Promise<Array<EventEnvelope>> {
+  // TODO: Manage pagination
+  const result: Array<EventEnvelope> = []
+
+  const keysBatches = inChunksOf(dynamoDbBatchGetLimit, keys)
+  logger.debug(`Performing batch get for ${keysBatches.length} batches`)
+  for (const keysBatch of keysBatches) {
+    const batchResult = await performBatchGet(dynamoDB, config, logger, keysBatch)
```
alvaroloes, line 182: The `batchGet` operation in Dynamo only allows you to get 100 events at once, that's why we need to do it in chunks

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,250 @@
...
+}
+
+function convertToSearchResult(eventEnvelopes: Array<EventEnvelope>): Array<EventSearchResponse> {
+  // "eventEnvelopes" represent a result page. That page is absolutely ordered among all the items in the
+  // database. BUT! the local order within the page is been messed up, so we need to sort the items.
+  // Why the order is messed up? Because we are doing two queries to DynamoDB:
+  // - One to the corresponding table index to know the keys of the items we need to retrieve from the master table.
+  // The result of this query is paginated and the absolute order of items is respected.
+  // - Another one to the master table to get the items data. This query is made with "batchQueryItems", which
+  // does not preserve the order in which we specify the keys. This is why we need to sort the final result.
```
alvaroloes, line 221: Check this comment to understand why we need to sort the result we have got.

packages/framework-provider-aws/src/library/keys-helper.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+export function decodeEventStoreSortingKey(encodedSortingKey: string): string {
+  return encodedSortingKey.split(sortingKeyEncodingSeparator)[0]
+}
+
+export function modifyEventsDecodingSortingKeys(...events: Array<EventEnvelope>): void {
+  for (const event of events) {
+    event.createdAt = decodeEventStoreSortingKey(event.createdAt)
+  }
+}
```
alvaroloes, line 24: Functions to deal with the uniqueness of the sorting keys

packages/framework-provider-aws/test/library/events-search-adapter.test.ts
```diff
@@ -0,0 +1,289 @@
...
+} from '@boostercloud/framework-types'
+import { random, date } from 'faker'
+import { DynamoDB } from 'aws-sdk'
+import { searchEvents } from '../../src/library/events-searcher-adapter'
+import { eventsStoreAttributes } from '../../src'
+import { partitionKeyForEvent, partitionKeyForIndexByEntity } from '../../src/library/keys-helper'
+import { DocumentClient } from 'aws-sdk/clients/dynamodb'
+import rewire = require('rewire')
+
+describe('Events searcher adapter', () => {
```
alvaroloes, line 21: This is the full suite of unit tests for the new functionality

juanjoman (changes requested):
I left a couple of questions below, and also a meme regarding all renaming that has been done 😂 

By the way, amazing tests!!

renaming

packages/framework-integration-tests/integration/providers/aws/functionality/events.integration.ts
```diff
@@ -64,7 +65,7 @@ describe('events', async () => {
     )
 
     const mockCartId = random.uuid()
-    const eventsToCreate = 26 // Using 26 because batches are of 25, making sure the batching gets triggered
+    const eventsToCreate = 30 // Using 30 because batches are of 25, making sure the batching gets triggered
```
NickSeagull, line 68: Why 30 and not 26?

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,250 @@
...
+  dynamoDB: DynamoDB.DocumentClient,
+  config: BoosterConfig,
+  logger: Logger,
+  filters: EventFilter
+): Promise<Array<EventSearchResponse>> {
+  logger.debug('Initiating an events search. Filters: ', filters)
+  const timeFilterQuery = buildSearchEventsTimeQuery(filters.from, filters.to)
+  const eventEnvelopes = await executeSearch(dynamoDB, config, logger, filters, timeFilterQuery)
+
+  logger.debug('Events search result: ', eventEnvelopes)
```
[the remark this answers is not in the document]
NickSeagull: It is `debug` so it shouldnt worry us. If the user enables `debug` level logging, its their fault 🤷

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
NickSeagull: Check my suggestion

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,240 @@
...
+    KeyConditionExpression: `${eventsStoreAttributes.partitionKey} = :partitionKey${timeQuery.expression}`,
+    ExpressionAttributeValues: {
+      ...timeQuery.attributeValues,
+      ':partitionKey': partitionKeyForEvent(entity, entityID),
+    },
+  }
+
+  logger.debug('Searching events by entity and entity ID. Query params: ', params)
+  const result = await dynamoDB.query(params).promise()
+  return (result.Items as Array<EventEnvelope>) ?? []
```
NickSeagull, line 111: To add pagination easily:
```suggestion
  const resultItems: Array<EventEnvelope> = []
  let result: PromiseResult<somethingfromaws, AWSError>  // Check this type
  // For pagination, just wrap in a do..while
  do {
    result = await dynamoDB.query(params).promise()
    resultItems.concat(result.Items as Array<EventEnvelope> ?? [])
  } while(result.LastEvaluatedKey)
  return resultItems
```

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,240 @@
...
+
+async function searchEventsByEntity(
+  dynamoDB: DynamoDB.DocumentClient,
+  config: BoosterConfig,
+  logger: Logger,
+  entity: string,
+  timeQuery: TimeQueryData
+): Promise<Array<EventEnvelope>> {
+  // TODO: manage pagination
+  // Fist query the index
```
NickSeagull, line 127:
```suggestion
  // First query the index
```

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,240 @@
...
+}
+
+async function searchEventsByEntity(
+  dynamoDB: DynamoDB.DocumentClient,
+  config: BoosterConfig,
+  logger: Logger,
+  entity: string,
+  timeQuery: TimeQueryData
+): Promise<Array<EventEnvelope>> {
+  // TODO: manage pagination
```
NickSeagull, line 126: Same as above

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
```diff
@@ -0,0 +1,240 @@
...
+    ExpressionAttributeValues: {
+      ...timeQuery.attributeValues,
+      ':partitionKey': type,
+    },
+  }
+
+  logger.debug('Searching events by type. Index query params: ', params)
+  const partialResult = await dynamoDB.query(params).promise()
+  const indexRecords = (partialResult.Items as Array<EventStoreKeys>) ?? []
+  return findEventsDataWithKeys(dynamoDB, config, logger, indexRecords)
```
NickSeagull, line 168: I believe these 3 functions should be abstracted into just one properly parameterized, as we are duplicating much code

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
NickSeagull: You can actually do it with a `do..while` and not need to count the batches

NickSeagull (changes requested):
Left some comments

alvaroloes:
> @alvaroloes can you create an issue for the pagination of events and add it to #636 please?

@NickSeagull Here it is: https://github.com/boostercloud/booster/issues/668 (I updated the main issue too)

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -71,20 +81,30 @@ export class GraphQLGenerator {
...
 
+  public eventResolver(
+    parent: unknown,
+    args: EventFilter,
+    context: GraphQLResolverContext,
+    info: GraphQLResolveInfo
+  ): Promise<Array<EventSearchResponse>> {
+    const eventsRequestEnvelope = toEventSearchRequest(args, context)
+    return this.eventsReader.fetch(eventsRequestEnvelope)
+  }
```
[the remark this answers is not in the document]
alvaroloes: They are not needed at all, but the signature of this function must much the signature of a GraphQL resolver, so I can't delete them

packages/framework-core/test/services/graphql/graphql-generator.test.ts
```diff
@@ -390,5 +391,40 @@ describe('GraphQL generator', () => {
...
+      let fetchEventsStub: SinonStub
+      const fetchEventsResult: Array<EventSearchResponse> = []
+      const filters: EventFilterByEntity = {
+        entity: 'TestEntity',
+      }
+
+      beforeEach(() => {
+        fetchEventsStub = stub().resolves(fetchEventsResult)
+        replace(BoosterEventsReader.prototype, 'fetch', fetchEventsStub)
+        sut = GraphQLGenerator.build(mockConfig, mockLogger)
```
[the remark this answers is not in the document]
alvaroloes: "System Under Test". It is a way of naming the "thing" you are teting that @tainguyenbui initiated here.

All the tests inside this context will be focused on testing the current "SUT" which in this case is the GraphQLGenerator

packages/framework-core/src/booster-events-reader.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    if (!BoosterAuth.isUserAuthorized(entityMetadata.authorizeReadEvents, eventRequest.currentUser)) {
+      throw new NotAuthorizedError('Access denied for reading events')
+    }
+  }
+
+  private entityMetadataFromRequest(eventRequest: EventSearchRequest): EntityMetadata {
+    const { filters } = eventRequest
+    if (isByEntitySearch(filters) === isByEventTypeSearch(filters)) {
+      // Means no search type was passed or both were passed
```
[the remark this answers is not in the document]
alvaroloes: Do you mean to provide a better error message? Well, I did this to avoid cluttering too much the function code and used the message `"It should contain either "type" or "entity" field"`, that works well for when you didn't include any type or both types.

Would you prefer to throw two different errors in this case (including an extra if)?

packages/framework-core/src/booster-events-reader.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private async processFetch(eventRequest: EventSearchRequest): Promise<Array<EventSearchResponse>> {
+    return Booster.events(eventRequest.filters)
+  }
+}
+
+function isByEntitySearch(filters: EventFilter): filters is EventFilterByEntity {
+  return 'entity' in filters
+}
```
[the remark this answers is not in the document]
alvaroloes: Yep!

packages/framework-core/src/decorators/entity.ts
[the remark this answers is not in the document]
alvaroloes: I'm not sure... It is more like an internal comment. The developer using `@Entity` doesn't need to know about this...

packages/framework-integration-tests/integration/providers/aws/functionality/events.integration.ts
alvaroloes: I changed it because I'm a bit paranoid about the "limit" conditions. I wanted to be sure two batches were needed. If we use 26 and mess a bit in the loop where we register events (Imagine we start the loop in 1 instead of 0) then we would get here a false positive.

In brief, I wanted to have more than just 1 unit above the limit condition

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
alvaroloes: Two things:
- How is this code dealing with pagination? I don't get it
- The kind of pagination that this method needs is another one: the ability to specify a pointer (`lastEvaluatedKey`) and page size to get a specific page. It is related to GraphQL.

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
alvaroloes: Check my response

packages/framework-core/src/booster-events-reader.ts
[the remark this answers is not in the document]
NickSeagull: I'd say yeah, it'd be clearer for the user.

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
alvaroloes: 😳 How is so? Could you please elaborate a bit?

packages/framework-integration-tests/integration/providers/aws/functionality/events.integration.ts
NickSeagull: I see, agree 100%.

PS: That's why loops are evil :trollface:

packages/framework-core/test/services/graphql/graphql-query-generator.test.ts
```diff
@@ -4,22 +4,33 @@ import { expect } from '../../expect'
...
   let mockGraphQLType: any
 
   let mockTypeInformer: SinonStubbedInstance<GraphQLTypeInformer>
   let mockByIdResolverBuilder: SinonStub
   let mockFilterResolverBuilder: SinonStub
+  let mockEventsResolver: SinonStub
 
   let getGraphQLTypeForStub: SinonStub
 
-  let sut: GraphQLQueryGenerator
```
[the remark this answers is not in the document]
juanjoman: not at all, I was just curious 😛 . Nevermind then @alvaroloes !

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/permissions.ts
```diff
@@ -27,10 +27,13 @@ export const setupPermissions = (
...
 
-  const { graphQLLambda, subscriptionsStore, subscriptionDispatcherLambda, connectionsStore } = graphQLStack
+  const { graphQLLambda, subscriptionsStore, subscriptionNotifier, connectionsStore } = graphQLStack
   const { eventsLambda, eventsStore } = eventsStack
   graphQLLambda.addToRolePolicy(
-    createPolicyStatement([eventsStore.tableArn], ['dynamodb:Query*', 'dynamodb:Put*', 'dynamodb:BatchWriteItem'])
+    createPolicyStatement(
+      [eventsStore.tableArn + '*'], // The '*' at the end is to also grant permissions on table indexes
+      ['dynamodb:Query*', 'dynamodb:Put*', 'dynamodb:BatchGetItem', 'dynamodb:BatchWriteItem']
+    )
```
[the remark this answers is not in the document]
juanjoman: The resource name should be `${eventStore.tableArn}/index/<my_index_name>` @alvaroloes

packages/framework-provider-aws/src/library/events-searcher-adapter.ts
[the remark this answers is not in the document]
juanjoman: Makes sense, thanks both! :v:

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/permissions.ts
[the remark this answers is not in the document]
alvaroloes: Thanks! I will work on that direction and let you know 😉

packages/framework-core/src/booster-events-reader.ts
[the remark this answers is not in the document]
alvaroloes: Ok, I'll change it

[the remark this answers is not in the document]
alvaroloes: Done here: 081e964be027caedadfbfab396bf018645022701

packages/framework-core/src/booster-events-reader.ts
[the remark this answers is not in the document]
alvaroloes: Done here: 081e964be027caedadfbfab396bf018645022701

packages/framework-core/src/decorators/entity.ts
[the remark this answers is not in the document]
alvaroloes: Done here: dd87361d8f6bc69713ab8731e00c7492384239d5

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/permissions.ts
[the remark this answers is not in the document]
alvaroloes: @juanjoman I didn't find a way to get the index name or arns from the CDK, but I achieve the same with a clever rearrangement of permission. 

Done here: aac22453b1565ca7a4ed9f0fbdedc36ca128ab86
If you want to go straight to the point, inside that commit, go to lines 58-71, for example

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/permissions.ts
```diff
@@ -43,19 +46,23 @@ export const setupPermissions = (
...
-  subscriptionDispatcherLambda.addToRolePolicy(websocketManageConnectionsPolicy)
+  subscriptionNotifier.addToRolePolicy(createPolicyStatement([subscriptionsStore.tableArn], ['dynamodb:Query*']))
+  subscriptionNotifier.addToRolePolicy(websocketManageConnectionsPolicy)
 
   eventsLambda.addToRolePolicy(
-    createPolicyStatement([eventsStore.tableArn], ['dynamodb:BatchWriteItem', 'dynamodb:Query*', 'dynamodb:Put*'])
+    createPolicyStatement(
+      [eventsStore.tableArn + '*'], // The '*' at the end is to also grant permissions on table indexes
+      ['dynamodb:Query*', 'dynamodb:Put*', 'dynamodb:BatchGetItem', 'dynamodb:BatchWriteItem']
+    )
```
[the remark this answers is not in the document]
alvaroloes: Done here: aac2245

packages/framework-core/src/helpers/promises.ts
```diff
@@ -0,0 +1,46 @@
...
+   * rejected. More importantly, **it does not wait for al the promises to finish**, which could lead to undesired behaviors
+   * - `Promise.allSettled`: This method waits for all the promises to finish and then returns an array of results. Some
+   * of them will be fulfilled and some rejected. More importantly, **it never throws an error**, which could lead to
+   * unexpected consequences. For example if you do "await Promise.allSettle(...)" expecting it to throw if some of them
+   * failed, you won't get that.
+   *
+   * In brief, `Promises.allSettledAndFulfilled` behaves exactly the same way as `Promise.allSettle` but it throws with
+   * an array of the failed promises, only if there are any.
+   */
+  public static async allSettledAndFulfilled<TValue>(
```
adriangmweb, line 20: Love it!

packages/framework-core/src/services/graphql/graphql-query-generator.ts
adriangmweb: There can be conflicts with the new GraphQL schema branch, just to have it in mind

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -29,32 +38,27 @@ export class GraphQLQueryGenerator {
...
+    const fields = { ...byIDQueries, ...filterQueries, ...eventQueries }
     return new GraphQLObjectType({
       name: 'Query',
       fields: fields,
     })
   }
 
-  private generateByIDQueries(): GraphQLFieldConfigMap<any, any> {
-    const queries: GraphQLFieldConfigMap<any, any> = {}
+  private generateByIDQueries(): GraphQLFieldConfigMap<unknown, GraphQLResolverContext> {
```
adriangmweb, line 60: Great getting rid of the anys 🙌🏻

adriangmweb (approved):
This is outstanding work!

NickSeagull (approved):

alvaroloes:
Merging: It has 2 approvals and all Juanjo's comments in his "request for changes" are already addressed!

PR #591, opened 2021-01-29
Switch to `main` as the base branch of the project

javiertoledo:
Github, Git and many other organizations and open-source communities are proposing to rename `master` branch as `main` [for the reasons explained in this document](https://sfconservancy.org/news/2020/jun/23/gitbranchname/) to abandon old terminology that doesn't make sense anymore.

In this PR, I've renamed all references to the `master` branch and replaced them by `main`. Once it's merged, Github has prepared a guide and automations to easily migrate all PRs, so this should be a seamless change: https://github.com/github/renaming

Bonus: I also removed the unused references to `master-mirror`

PR #607, opened 2021-02-09
Fix: Reference error in publish GitHub action script

javiertoledo:
I went too far with the search and replace when we switched to using the branch `main` as the production one in #591

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -64,7 +64,7 @@ jobs:
       - name: Discard all the changes after integration tests
         run: git stash
 
-      - uses: theam/actions/lerna-semantic-publish@main
+      - uses: theam/actions/lerna-semantic-publish@master
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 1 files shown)


rdoria1:
@javiertoledo I'm missing something here.
If we removed the master branch aren't we suppose to execute all the actions on the main branch?

I would have thought that we needed to change this line https://github.com/boostercloud/booster/blob/03db5d1473b6d4085b926d5bcf653d8fe3dc51c7/.github/workflows/publish.yml#L5 to `main`

javiertoledo:
@rdoria1 notice that the reference you pasted in your comment goes to 03db5d1473b6d4085b926d5bcf653d8fe3dc51c7, not to `main`. In the `main` branch, all the changes are already applied, exactly as you suggested:

https://github.com/boostercloud/booster/blob/main/.github/workflows/publish.yml#L5

The change in this branch fixes a broken reference to a different repository that still uses `master` as their primary branch: https://github.com/theam/actions

rdoria1 (approved):
LGTM

AdrianLorenzoDev (approved):
LGTM! :)

juanjoman (approved):

Issue #420, opened 2020-09-21
Throw error when the Booster CLI is used in a project with a different Booster version


alvaroloes:
When the Booster CLI is used in a project, and the version of the CLI is not the same as the Booster version specified in the project, we should throw an error telling the user to upgrade/downgrade the CLI.

Things to consider:
- If the versions are different only in the "fix" section (A.B.*), then we might only throw a warning and keep going.
- If the CLI version is lower than the project Booster version, then FAIL and tell the user to update
- If the CLI version is higher, then:
   - If the difference is only in the "feat" section (A.*.C), then offer the user the option to update all dependencies automatically.
   - If the difference is in the "breaking" section (*.B.C), then FAIL and tell the user to:
      - Use a lower CLI version
      - Or upgrade the project by following the Github release notes (with a link to the specific version)

PR #609, opened 2021-02-10
Feature/420 throw error different cli version

davidverdu:
## Description
`boost` cli command now checks the project Booster version and compares with cli version. When you run the cli command, the following algorithm is executed beforehand to check versions:

- If versions match, run normally.
- If the versions are different only in the "fix" section (A.B.*), then we might only throw a warning and keep going.
- If the CLI version is lower than the project Booster version, then FAIL and tell the user to update
- If the CLI version is higher, then tell the user to Use a lower CLI version or upgrade the project

Closes: #420 

## Changes
- I created the project-updater service that updates the project package.json when applies
- I updated all commands except `new:project` to make version checks after running their main part
- I created new unit tests
- I made the code work with our integration tests

## Checks
- [x] Project Builds


packages/cli/src/services/project-updater.ts
```diff
@@ -0,0 +1,20 @@
...
+import * as path from 'path'
+
+export async function updatePackageJsonDependencyVersions(version: string, projectPath: string): Promise<void> {
+    const projectAbsolutePath = path.resolve(projectPath)
+    const packageJsonPath = path.join(projectAbsolutePath, 'package.json')
+    try {
+        const packageJsonContents = require(packageJsonPath)
+        packageJsonContents.dependencies['@boostercloud/framework-core'] = `^${version}`
+        if (packageJsonContents.dependencies['@boostercloud/framework-types'] !== undefined) {
+            packageJsonContents.dependencies['@boostercloud/framework-types'] = `^${version}`
```
juanjoman, line 11: isn't it better to remove the `^` symbol? We're checking versions anyway, right? 🤔 

Also, will rockets get updated with this?

packages/cli/src/commands/nuke.ts
```diff
@@ -54,6 +55,9 @@ export default class Nuke extends Command {
 
   public async run(): Promise<void> {
     const { flags } = this.parse(Nuke)
+
+    await checkCurrentDirBoosterVersion(this.config.userAgent)
```
juanjoman, line 59: I think we're calling this on every CLI command right? Is there a way that it could be called before all of them?

packages/cli/test/services/project-checker.test.ts
```diff
@@ -54,4 +61,171 @@ describe('project checker', (): void => {
             expect(exceptionThrown).to.be.equal(true)
         })
     })
+
+    describe('checkCurrentDirBoosterVersion', (): void => {
+        
+        beforeEach(() => {
+            replace(projectUpdater,'updatePackageJsonDependencyVersions', fake.resolves({}))
+            replace(logger,'info', fake.resolves({}))
```
juanjoman, line 69: can you check that the linter isn't failing for this? 😮 
```suggestion
            replace(logger, 'info', fake.resolves({}))
```

packages/cli/test/services/project-checker.test.ts
```diff
@@ -54,4 +61,171 @@ describe('project checker', (): void => {
...
+        })
+
+        describe('inside a Booster project', () => {
+            //project version in mocked package.json is 1.11.2
+            beforeEach(() => {
+                replace(process,'cwd', fake.returns(path.join(process.cwd(),'test', 'fixtures', 'mock_project')))
```
juanjoman, line 75:
```suggestion
                replace(process, 'cwd', fake.returns(path.join(process.cwd(), 'test', 'fixtures', 'mock_project')))
```

By the way, don't worry too much about these, I'm just worried about the linter failing :v:

juanjoman (changes requested):
Nice PR!! Just left a couple of minor comments below. Also, is there a way that we can provide integration tests simulating the dependency update process? 🙏

packages/cli/src/commands/nuke.ts
NickSeagull: Yep I was thinking in the same thing. Perhaps with the [Custom Base Class feature from `oclif`](https://oclif.io/docs/base_class)?

packages/cli/src/services/project-checker.ts
```diff
@@ -69,3 +71,66 @@ export async function checkResourceExists(name: string, placementDir: string, ex
...
+    throw new Error(
+      `There was an error when recognizing the application. Make sure you are in the root path of a Booster project:\n${e.message}`
+    )
+  }
+}
+
+async function compareVersionsAndDisplayMessages(cliVersion: string, projectVersion: string, projectPath: string): Promise<void> {
+  if (cliVersion === projectVersion)  { return }
+  const cliVersionParts = cliVersion.split('.').map((v) => parseInt(v,10))
+  const projectVersionParts = projectVersion.split('.').map((v) => parseInt(v,10))
```
NickSeagull, line 102: I'd abstract this into a function

```ts
function parseSemver(semverString: string): [number, number, number] {
  return semverString.split('.').map((v) => parseInt(v,10))
}


  const cliVersionParts = parseSemver(cliVersion)
  const projectVersionParts = parseSemver(projectVersion)
```

packages/cli/src/services/project-checker.ts
```diff
@@ -69,3 +71,66 @@ export async function checkResourceExists(name: string, placementDir: string, ex
...
+  if (cliVersionParts.length !== projectVersionParts.length) {
+    throw new Error(`Versions must have the same length. CLI version: ${cliVersion}. Project Booster version: ${projectVersion}`)
+  }
+  if (projectVersionParts.length !== 3) {
+    throw new Error(`Versions must follow semantic convention X.Y.Z | CLI version: ${cliVersion}. Project Booster version: ${projectVersion}`)
+  }
```
NickSeagull, line 108: This would go into the `parseSemver` function

packages/cli/src/services/project-checker.ts
```diff
@@ -69,3 +71,66 @@ export async function checkResourceExists(name: string, placementDir: string, ex
...
+  if (cliVersionParts[0] === projectVersionParts[0]) {
+    if (cliVersionParts[1] === projectVersionParts[1]) {
+      //differences in the 'fix' part
+      if (cliVersionParts[2] !== projectVersionParts[2]) {
+        logger.info(`WARNING: Project Booster version differs in the 'fix' section. CLI version: ${cliVersion}. Project Booster version: ${projectVersion}`)
+      }
```
NickSeagull, line 114: Instead of nesting `if`s, I'd create a check function for it, and appropriate action functions as well:

```suggestion
type Semver = [number, number, number]

function semverDifferentFixVersion(a: Semver, b: Semver): bool {
  return a[0] === b[0] && a[1] === b[1] && a[2] !== b[2]
}

function warnDifferentFixVersion(cliVersion: string, projectVersion: string){
        logger.info(`WARNING: Project Booster version differs in the 'fix' section. CLI version: ${cliVersion}. Project Booster version: ${projectVersion}`)
}

// now in the code...
if(semverDifferentFixVersion(cliVersionParts, projectVersionParts) {
  warnDifferentFixVersion(cliVersion, projectVersion)
}
```

packages/cli/src/services/project-checker.ts
```diff
@@ -69,3 +71,66 @@ export async function checkResourceExists(name: string, placementDir: string, ex
...
+    } else if (cliVersionParts[1] > projectVersionParts[1]) {
+      //cli higher than project in 'feat' section
+      const promptMsg = `@boostercloud/cli version ${cliVersion} is higher than project version (${projectVersion}) in the 'feature' section. Do you want to upgrade your project dependencies?`
+      const value: string = await new Prompter().defaultOrChoose(undefined,promptMsg,['Yes','No'])
+      if (value == 'Yes') {
+        await updatePackageJsonDependencyVersions(cliVersion, projectPath) //update package.json
+        logger.info(`package.json Booster dependencies have been updated to version ^${cliVersion}.`)
+      } else {
+        throw new Error(`CLI version ${cliVersion} is higher than your project Booster version ${projectVersion}. Please upgrade your project dependencies.`)
+      }
```
NickSeagull, line 124: I wouldn't do the upgrading ourselves, as there are tools to do that. And also, the user should do themselves, we shouldn't be responsible of it I believe.

Regarding refactoring, same as above:

```ts
if (semverGreaterFeatureVersion(cliVersionParts, projectVersionParts) {
  throw greaterVersionError(cliVersion, projectVersion)
}
```

packages/cli/src/services/project-checker.ts
```diff
@@ -69,3 +71,66 @@ export async function checkResourceExists(name: string, placementDir: string, ex
...
+    } else {
+      //cli lower than project in 'feat' section
+      throw new Error(`CLI version ${cliVersion} is lower than your project Booster version ${projectVersion}. Please upgrade your @boostercloud/cli to the same version with npm`)
```
NickSeagull, line 127: I'd do the check explicit here as well:

```ts
if (semverLowerVersion(cliVersion, projectVersion) {
  throw lowerVersionError(...)
}
```

packages/cli/src/services/project-checker.ts
```diff
@@ -69,3 +71,66 @@ export async function checkResourceExists(name: string, placementDir: string, ex
...
+    }
+  } else if (cliVersionParts[0] > projectVersionParts[0]) {
+    //cli higher than project in 'breaking' section
+    throw new Error(`CLI version ${cliVersion} is higher than your project Booster version (${projectVersion}). Please upgrade your project dependencies or install the same CLI version with \'npm install @boostercloud/cli@${projectVersion}\'.`)
+  } else {
+    //cli lower than project in 'breaking' section
+    throw new Error(`CLI version ${cliVersion} is lower than your project Booster version ${projectVersion}. Please upgrade your @boostercloud/cli to the same version with npm`)
+  }
```
NickSeagull, line 135: Also these two checks

packages/cli/src/services/project-updater.ts
NickSeagull, line 11: I'd extract these into `const`s instead of repeating them again and again

packages/cli/test/services/project-checker.test.ts
NickSeagull: Remember to run `./scripts/check-all-the-things.sh` before submitting PRs

NickSeagull (changes requested):
Left some comments

packages/cli/src/services/project-checker.ts
davidverdu: I refactored this by putting all logic in the Semver class. Now comparisons in code looks more semantic

packages/cli/src/services/project-checker.ts
davidverdu: I add this logic to Semver class

packages/cli/src/services/project-checker.ts
davidverdu: A better approach is to encapsulate this logic in Semver class

packages/cli/src/services/project-updater.ts
davidverdu: I used ^ symbol because it appears in all project's package.json files. I think I can replace all instances of @boostercloud with a loop instead

packages/cli/src/services/project-checker.ts
davidverdu: what tool can you use to update package.json externally without editing the file manually @NickSeagull ? maybe npm has a way to do that? I decided to update the package.json file because it was specified in the issue.

packages/cli/src/commands/nuke.ts
davidverdu: It's a good idea to inherit from a BaseCommand class. Actually, this refactoring can help us to address other issues from the list

packages/cli/src/services/project-checker.ts
NickSeagull: The one most used to update the dependencies is [`npm-check-updates`](https://www.npmjs.com/package/npm-check-updates) but still, that'd update every dependency. To be honest, I think that we can remove the dependencies upgrading and leave it for a future PR, don't you think @alvaroloes (mentioning you because you reported the issue)?

packages/cli/src/services/semver.ts
```diff
@@ -0,0 +1,34 @@
+export default class Semver {
```
NickSeagull, line 1: I love this!

NickSeagull (commented):
I'd remove the dependency upgrading, but everything else LGTM!

packages/cli/src/services/project-updater.ts
davidverdu: Done! Nonetheless, I think we are about to avoid upgrading package.json in the end @NickSeagull

packages/cli/src/services/project-checker.ts
davidverdu: I have created custom Error subclasses for these cases when versions differ

packages/cli/test/services/project-checker.test.ts
davidverdu: I had to run `lerna run lint:check` to fix the linting errors. `./scripts/check-all-the-things.sh` wasn't enough

packages/cli/src/services/project-updater.ts
davidverdu: @juanjoman I implemented the method in a better way. Nonetheless, we are not using it, since it's better to make the user update package.json by themself

packages/cli/src/services/project-checker.ts
```diff
@@ -69,3 +70,60 @@ export async function checkResourceExists(name: string, placementDir: string, ex
...
+    )
+  }
+}
+
+class HigherCliVersionError extends Error {
+  constructor(public cliVersion: string, public projectVersion: string, public section: string) {
+    super(`CLI version ${cliVersion} is higher than your project Booster version ${projectVersion} in the '${section}' section. Please upgrade your project dependencies.`)
+  }
+}
+
```
davidverdu, line 103: @NickSeagull I created subclasses of `Error` to display version difference messages

packages/cli/src/services/project-updater.ts
```diff
@@ -0,0 +1,28 @@
...
+          }
+        }
+        fs.outputFile(packageJsonPath, JSON.stringify(packageJsonContents,null,2))
+    } catch (e) {
+      throw new Error(
+        `There was an error when recognizing the application package.json file. Make sure you are in the root path of a Booster project:\n${e.message}`
+      )
+    }
+}
+
```
davidverdu, line 28: @NickSeagull @juanjoman this function is not used in the end, but I left the code in case we use it in the future

packages/cli/src/services/project-checker.ts
davidverdu: @NickSeagull @javiertoledo @alvaroloes in the end I decided to not update `package.json` from the cli. Users should do changes on that file manually or with `npm`. We can leave project updating for the future

packages/cli/src/services/project-updater.ts
NickSeagull: It's better to remove it then, dead code should be deleted code! 😊

NickSeagull (changes requested):
Much better! Just one small change

packages/cli/src/services/project-updater.ts
davidverdu: OK. I'll delete it!

NickSeagull (approved):

packages/cli/src/services/project-checker.ts
davidverdu: @juanjoman we won't upgrade `package.json`, so no integration tests to check that are needed

juanjoman (approved):
LGTM!! 🚀

packages/cli/src/services/project-checker.ts
```diff
@@ -69,3 +70,60 @@ export async function checkResourceExists(name: string, placementDir: string, ex
...
+  } catch (e) {
+    throw new Error(
+      `There was an error when recognizing the application. Make sure you are in the root path of a Booster project:\n${e.message}`
+    )
+  }
+}
+
+class HigherCliVersionError extends Error {
+  constructor(public cliVersion: string, public projectVersion: string, public section: string) {
+    super(`CLI version ${cliVersion} is higher than your project Booster version ${projectVersion} in the '${section}' section. Please upgrade your project dependencies.`)
```
alvaroloes, line 100:
```suggestion
    super(`CLI version ${cliVersion} is higher than your project Booster version ${projectVersion} in the '${section}' section. Please upgrade your project Booster dependencies.`)
```

packages/cli/src/services/project-checker.ts
```diff
@@ -69,3 +70,60 @@ export async function checkResourceExists(name: string, placementDir: string, ex
...
+  }
+}
+
+class LowerCliVersionError extends Error {
+  constructor(public cliVersion: string, public projectVersion: string, public section: string) {
+    super(`CLI version ${cliVersion} is lower than your project Booster version ${projectVersion}. Please upgrade your @boostercloud/cli to the same version with npm`)
```
alvaroloes, line 106: Leveraging that we are writing a very nice error message (amazing, BTW), I would go further and add the command to upgrade the CLI version:
```suggestion
    super(`CLI version ${cliVersion} is lower than your project Booster version ${projectVersion}. Please upgrade your @boostercloud/cli to the same version with "npm install -g @boostercloud/cli@${projectVersion}"`)
```

alvaroloes (approved):
Outstanding work @davidverdu. Great separation of concerns, clean code, 💯  

Please check my suggestion and question if possible 😉

packages/cli/src/services/project-checker.ts
```diff
@@ -69,3 +70,60 @@ export async function checkResourceExists(name: string, placementDir: string, ex
...
+  const cliVersion = userAgent.split(' ')[0].split('/')[2]
+  await compareVersionsAndDisplayMessages(cliVersion, projectVersion)
+}
+
+async function getBoosterVersion(projectPath: string): Promise<string> {
+  const projectAbsolutePath = path.resolve(projectPath)
+  try {
+    const packageJsonContents = require(path.join(projectAbsolutePath, 'package.json'))
+    const version = packageJsonContents.dependencies['@boostercloud/framework-core']
+    const versionParts = version.replace('^', '').replace('.tgz', '').split('-')
```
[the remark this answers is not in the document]
davidverdu: Exactly @alvaroloes . I had to do the `replacement hack` to make it work with our integration tests, where a `.tgz` is used for `@boostercloud/framework-core` dependency in place of a direct version number.

packages/cli/src/services/project-checker.ts
davidverdu: OK

packages/cli/src/services/project-checker.ts
davidverdu: OK

Issue #610, opened 2021-02-11
Transform `BoosterConfig` into an interface, move the functionality to a `BoosterApp` class


javiertoledo:
In the current implementation, both `BoosterConfig` and `BoosterApp` are located in the `ftramework-types` package. The goal of this package was sharing common interfaces and types used by the rest of the Booster packages, but while `BoosterApp` was remained an interface, `BoosterConfig` has become a class and we've been adding functionality to it to the point it has become quite complex.

In the context of #517, we're considering building an static version of the `BoosterConfig` object to pass the project metadata both to the CLI and infrastructure modules without dynamically requiring the user project from the CLI. These dynamic requires have made it really hard to reason about how the code is loaded and executed, especially regarding dependencies.

Looking at the current implementation, it could make sense to move all the functionality and object references to a new `BoosterApp` class in the core package, leaving the `BoosterConfig` as a simpler interface defining the metadata structure exclusively.

Separating the active part (functions and object references) from the static part (artifact names and project metadata) will foster separation of concerns, making it easier to serialize and export the project metadata to create functionality around Booster applications without dependencies at the code level, enabling us to build external tools in other programming languages.

Issue #566, opened 2021-01-20
Refactor framework-core package to use Singleton in BoosterConfig, Logger and other classes


davidverdu:
## Description

After refactoring GraphQLGenerator we have observed that most of the classes in framework-core package can be refactored by using Singleton pattern in config and logger classes. It can also be applied to BoosterCommandDispatcher and BoosterReadModelDispatcher. For the latter, for example, the code is creating two instances of that class to invoke a method in config, which has a unique instance.

## Possible Solution

Apply the Singleton pattern to Config and Logger. Avoid passing those instances in the constructor of the other client classes. Apply Singleton to the classes, avoiding passing config and logger in the getInstance() Singleton method.

## Additional information
Related PR: https://github.com/boostercloud/booster/pull/555

alvaroloes:
@davidverdu I think this is already solved in release https://github.com/boostercloud/booster/releases/tag/v0.11.2

Could you please confirm?

davidverdu:
@alvaroloes This issue is a refactoring that has not be applied yet. It improves code without adding new features.

PR #612, opened 2021-02-11
Reworked `Booster` class and `BoosterApp` interface

javiertoledo:
## Description

This PR reorganizes the main interfaces and classes in the `framework-types` and `framework-core` packages in preparation to address the changes suggested in #610, which are also needed for future refactors related to the framework extensibility (see #517).

## Changes

* Removed the `BoosterApp` interface from the `framework-types` package. The main goal of this interface was separating the public from private methods of the `Booster` class in the core package to avoid exporting the static lambda methods.
* Renamed the `Booster` class in the core package as `BoosterApp`
* Removed the supporting static methods re-exported as root-level from the `Booster` class in the core package and reimplemented them in the index file. The result is simpler and better organized.
* Removed an unused reference to `BoosterApp` from the `UserApp` interface, used to cast the right types when the user project was dynamically loaded via require.
* Improved tests for `index.ts` and `booster-app.ts` in the core module.
* Made the `logger` in the `BoosterApp` class public to allow reducing the need of passing a logger instance explicitly all over the place (To be tackled in future PRs). This is related to #566.

There are no functional changes.

## Checks
- [x] Project Builds


juanjoman (approved):
LGTM! :v:

packages/cli/src/services/config-service.ts
```diff
@@ -35,23 +35,29 @@ async function compileProject(projectPath: string): Promise<void> {
...
+   * a static version of the config that can be loaded without
+   * dynamically requiring the user project from the CLI.
+   */
   const projectIndexJSPath = path.resolve(path.join(userProjectPath, 'dist', 'index.js'))
   return require(projectIndexJSPath)
 }
 
-function checkEnvironmentWasConfigured(app: BoosterApp): void {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function checkEnvironmentWasConfigured(app: any): void {
```
NickSeagull, line 60: Why `any`?

NickSeagull (approved):
Just a small comment, overall LGTM!

davidverdu (approved):
LGTM

packages/cli/src/services/config-service.ts
```diff
@@ -35,23 +35,29 @@ async function compileProject(projectPath: string): Promise<void> {
...
     app.configureCurrentEnv((config: BoosterConfig): void => {
       checkEnvironmentWasConfigured(app)
       resolve(config)
     })
   })
 }
 
-function loadUserProject(userProjectPath: string): { Booster: BoosterApp } {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function loadUserProject(userProjectPath: string): { Booster: any } {
```
alvaroloes, line 49: Why changing this to `any`? On the calling site, we now have lost compile-time guarantees. 

For example, 9 lines above, on line 41 of this file, we are calling `app.configureCurrentEnv`, but that call is not checked at compile time now. If we change the name of the function or its parameters, it will silently compile but blow up in runtime.

packages/cli/src/services/config-service.ts
alvaroloes: I have the same concern and the reasons are the same as the previous comments. Now, all the usage of `app` inside this function is not compile-time checked

packages/framework-core/src/booster-app.ts
```diff
@@ -9,27 +9,26 @@ import {
...
-export class Booster {
+export class BoosterApp {
+  /**
+   * Set of environment names set from the application using `Booster.configure('environment_name', ...)`
+   */
   public static readonly configuredEnvironments: Set<string> = new Set<string>()
-  private static logger: Logger
+
+  /**
+   * Default logger, as configured in the current
```
alvaroloes, line 25: I think there is something missing after `... in the current`

packages/framework-core/src/booster-read-model-dispatcher.ts
```diff
@@ -56,7 +56,7 @@ export class BoosterReadModelDispatcher {
 
   private async processFetch(readModelRequest: ReadModelRequestEnvelope): Promise<Array<ReadModelInterface>> {
     const readModelMetadata = this.config.readModels[readModelRequest.typeName]
-    const searcher = Booster.readModel(readModelMetadata.class)
+    const searcher = BoosterApp.readModel(readModelMetadata.class)
```
alvaroloes, line 59: 🤔 To be honest, I liked more the name `Booster`. As it acts as the library the user uses in their application (it is not an app, I mean) and I think "BooserApp" adds a bit of confusion. Was there a specific reason to change it?

packages/framework-core/src/index.ts
```diff
@@ -1,12 +1,55 @@
...
-export {
-  boosterEventDispatcher,
-  boosterPreSignUpChecker,
-  boosterServeGraphQL,
-  boosterNotifySubscribers,
-  boosterTriggerScheduledCommand,
-} from './booster'
-
-export const Booster: BoosterApp = boosterModule.Booster
+export { BoosterApp } from './booster-app'
```
alvaroloes, line 55: Why are we exporting both `Booster` and `BoosterApp` names if they are the same?

packages/framework-core/test/index.test.ts
```diff
@@ -1,26 +1,88 @@
...
 
-  it('exports the `boosterNotifySubscribers` function', () => {
-    expect(BoosterCore.boosterNotifySubscribers).not.to.be.null
-    expect(BoosterCore.boosterNotifySubscribers).to.equal(Booster.boosterNotifySubscribers)
+  it('exports the BoosterApp class and the Booster alias', () => {
+    expect(BoosterCore.Booster).not.to.be.null
+    expect(BoosterCore.BoosterApp).not.to.be.null
+    expect(BoosterCore.Booster).to.be.equal(BoosterCore.BoosterApp)
   })
 })
```
alvaroloes, line 88: Thanks for the tests!!

packages/framework-types/src/user-app.ts
```diff
@@ -1,14 +1,13 @@
...
-import { BoosterApp } from '.'
 
 /**
- * `UserApp` represents the interface of the value when
- * we `require` the `dist/index.js` file of the user.
+ * `UserApp` represents the lambdas interface of a Booster app
+ * when it's dynamically imported from a different module.
  */
 export interface UserApp {
-  Booster: BoosterApp
```
alvaroloes, line 9: Still don't quite understand why this was removed. When we load the project, the field `Booster` is exported and it is needed to be used in the core and in the local provider.

alvaroloes (changes requested):
Leave this a request change especially for two comments regarding compilation time guarantees. I also left some questions because I don't fully understand the reasons

javiertoledo:
@alvaroloes @NickSeagull I see your concerns about removing the type checks from the dynamically required user project. This change is just the first change of a much bigger refactorization. I just wanted to keep the change as small as possible to be able to review and merge it in small steps rather than changing the whole architecture in a single PR.

I'd recommend to read carefully the issues #517 and #610, as suggested in the PR description, before reviewing this code.

Notice that I removed the `BoosterApp` interface from the `framework-types` and transformed it into a class in `framework-core`. The goal of this change is starting to move everything that is dynamic to the `BoosterApp` class and keep everything that is static in `BoosterConfig`. After removing the `BoosterApp` interface from `framework-types`, it's no longer accessible from the CLI package without adding a dependency to `framework-core`, which is one of the dependencies we want to break. Take into account that once all the changes described in #610 are done, the CLI will no longer load the user project dynamically, but it will load a serialized version of the `BoosterConfig` from a config file, so these `any` references will disappear too.

packages/cli/src/services/config-service.ts
javiertoledo: This is the first change towards the refactorization described in #517, so these `any` references will eventually disappear. After this change, `BoosterApp` is no longer part of the types package, so I have no easy way to require it.

packages/cli/src/services/config-service.ts
javiertoledo: The goal is to get rid of dynamic project requires, read more about it here: https://github.com/boostercloud/booster/pull/612#issuecomment-789051941

packages/framework-core/src/booster-app.ts
javiertoledo: I have no idea, so...
```suggestion
   * Default logger
```

packages/framework-core/src/booster-read-model-dispatcher.ts
javiertoledo: `BoosterApp` is the representation of the current app in the core package. The user will still use the name `Booster`. I would keep working on the refactor proposed in #517 before deciding the final name of the constants in the core package, there's a lot to untangle in there yet!

packages/framework-core/src/index.ts
javiertoledo: The ultimate goal will be exporting `Booster` exclusively, but I don't remember why I needed to export `BoosterApp` too. I'll review.

packages/framework-types/src/user-app.ts
javiertoledo: This field wasn't used at all, all usages of the `UserApp` interface only accessed the lambdas.

NickSeagull:
Thanks for the explanation @javiertoledo , much clearer now, LGTM

packages/framework-core/src/index.ts
javiertoledo: It looks like it was for no reason, I removed the export of the `BoosterApp` class

javiertoledo:
@alvaroloes in order to make the config serializable we have to separate metadata from the "active" state. The way I think about this is: The structure of the metadata belongs to the type package, but the active state belongs to the core package. This change is the first step to transform `BoosterConfig` into a serializable type, and `BoosterApp` into a class in the core package.

javiertoledo:
Closing, as this PR has become too old and is now full of conflicts

PR #614, opened 2021-02-16
Unit tests for Batch File Processing Rocket

charlietfe:
## Description
Added unit test to Batch File Processing Rocket

## Checks
- [X] Project Builds
- [N/A] Updated documentation accordingly

### DIFF
--- a/packages/rocket-batch-file-process-aws-infrastructure/src/lambdas/package-lock.json
+++ b/packages/rocket-batch-file-process-aws-infrastructure/src/lambdas/package-lock.json
@@ -1,0 +1,283 @@
+{
+  "name": "lambdas",
+  "version": "1.0.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "version": "1.0.0",
+      "bundleDependencies": [
+        "uuid"
+      ],
+      "license": "ISC",
+      "dependencies": {
+        "@types/aws-lambda": "8.10.48",
+        "aws-sdk": "2.764.0",
+        "tslib": "2.1.0"
+      },
+      "devDependencies": {
+        "uuid": "^8.3.2"
+      }
+    },
+    "node_modules/@types/aws-lambda": {
+      "version": "8.10.48",
+      "resolved": "https://registry.npmjs.org/@types/aws-lambda/-/aws-lambda-8.10.48.tgz",
+      "integrity": "sha512-+qFDcssXvrdXIxBbKCJp0atg94TJVJSt5sx3Cu6LOQX/EV2mbInjgxGeKuLmFFBjoxD7G6fSytZoeC6A9fzTuw=="
+    },
+    "node_modules/aws-sdk": {
+      "version": "2.764.0",
+      "resolved": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.764.0.tgz",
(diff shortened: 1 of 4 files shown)


adriangmweb (approved):
LGTM! 🚀

davidverdu (approved):
LGTM. The only doubt I have is if we should put the `package-lock.json` files in `.gitignore` in place of adding it in the PR. When I compile the project and launch the tests in my machine, some `package-lock.json` files, including the one added by this PR, get modified and appear to be modified if I run `git status`.

javiertoledo:
There are different opinions on what to do with lock files. In my opinion, having the `package-lock.json` checked in the repository makes sense because it helps us to make sure that everyone is using the same package versions, and if someone upgrades a package, then everyone else can see it in the PR.

packages/rocket-batch-file-process-aws-infrastructure/src/lambdas/package-lock.json
```diff
@@ -0,0 +1,283 @@
...
+        "xmlbuilder": "~9.0.1"
+      }
+    },
+    "xmlbuilder": {
+      "version": "9.0.7",
+      "resolved": "https://registry.npmjs.org/xmlbuilder/-/xmlbuilder-9.0.7.tgz",
+      "integrity": "sha1-Ey7mPS7FVlxVfiD0wi35rKaGsQ0="
+    }
+  }
+}
```
javiertoledo, line 283: We should find a better way of doing this, having a package inside of a package is... weird 🤔

javiertoledo (approved):

packages/rocket-batch-file-process-aws-infrastructure/test/unit/batch-file-processing.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    params = {
+      bucketName,
+      chunkSize,
+    }
+  })
+
+  context('on mount', () => {
+    const verifyFunctionPolicies = (currentPolicies: any[], expectedPolicies: any[]) => {
+      expectedPolicies.forEach((item) => {
+        const plcy = currentPolicies.find((policy) => {
```
NickSeagull, line 71: I think it'd be better inverting the names:
```suggestion
        const policy = currentPolicies.find((plcy) => {
```

packages/rocket-batch-file-process-aws-infrastructure/test/unit/batch-file-processing.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+      const notifications = getConstructObjValueForKey(sourceBucketObj, 'notifications')['lambdaNotifications']
+      expect(notifications).containSubset([{ Events: [EventType.OBJECT_CREATED] }])
+
+      // Stagin Bucket
+      const staginBucket = appStack.node.tryFindChild('stagingUploadBucket') as Bucket
+      expect(staginBucket).not.to.be.null
+
+      const staginBucketObj = constructToObj(staginBucket)
+      expect(staginBucketObj).containSubset([{ key: 'physicalName', value: `${bucketName}-staging` }])
```
NickSeagull, line 97: These should be `staging` not `stagin`

charlietfe:
The test will be included in its own rocket repo.

PR #619, opened 2021-02-19
Extend Batch file Rocket: row event from event handler + add support for jsonl files

rdoria1:
## Description
This PR extends the current Batch file rocket to the next level of abstraction.
As of now we used to receive a `FileAdded` event in the Event Handler representing the chunked file but we still needed to use the AWS sdk to open the file, split by "," if CSV file and etc.. logic that we can abstract to the rocket.
This PR does that. We receive an event representing a line of the file directly.

Also, add support for jsonl files
## Changes
New files.

## Checks
- [X] Project Builds

### DIFF
--- a/packages/rocket-batch-file-process-aws-infrastructure/README.md
+++ b/packages/rocket-batch-file-process-aws-infrastructure/README.md
@@ -7,11 +7,17 @@ Every time a new File is uploaded to the Source S3 Bucket, a new Lambda function
 This Lambda function will
 
 - Split the source file in smaller chunks. The chunk size is defined by the user as input parameter of the rocket.
-- Persist the new formed chunk in the Staging Bucket.
+- Persist the new formed chunk file in the Staging Bucket.
 - Persist a new event (containing fileUri and fileSize) for each chunk in the Event Store.
 
-**Disclaimer:** As of now the rocket can only process CSV files.
+Then, for every chunk file dropped in the Staging rocket a new Lambda function will be triggered that will:
 
+- Read each chunk file line by line and persist a new event in the Events Store for each row
+
+This event is defined in the Booster application and can be consumed from the Event Handler for any kind of processing.
+You drop your file, and you implement your logic based on an event representing a line of the file.
+
+*Disclaimer: Currently the rocket supports CSV and jsonl files.* 
 ## Usage
 
 Install this package as a dependency in your Booster project.
@@ -32,9 +38,16 @@ Booster.configure('development', (config: BoosterConfig): void => {
   config.provider = Provider([{
     packageName: '@boostercloud/rocket-batch-file-process-aws-infrastructure', 
     parameters: {
-      bucketName: 'test-bucket-name', 
-      chunkSize: '2',
-    }
+      config: {
+        bucketName: 'test-bucket',
+        chunkSize: '2',
+      },
+      rowEvent: {
+        entityId: 'id',
+        eventTypeName: 'RowAdded',
+        entityTypeName: 'RowEntity',
+      },
+    },
   }])
 })
 ```
--- a/packages/rocket-batch-file-process-aws-infrastructure/README.md
+++ b/packages/rocket-batch-file-process-aws-infrastructure/README.md
@@ -2,19 +2,19 @@
 
 This package is a configurable Booster rocket for parallel batch file processing.
 First, it creates a source Object Storage (S3) and a Staging Object Storage(S3).
-Every time a new File is uploaded to the Source S3 Bucket, a new Lambda function will be triggered.
 
-This Lambda function will
+Every time a new File is uploaded to the Source S3 Bucket, a new Lambda function will be triggered that will:
 
-- Split the source file in smaller chunks. The chunk size is defined by the user as input parameter of the rocket.
+- Split the source file in smaller chunks. The `chunkSize` is defined by the user as input parameter of the rocket (See example below).
 - Persist the new formed chunk file in the Staging Bucket.
-- Persist a new event (containing fileUri and fileSize) for each chunk in the Event Store.
+- Persist a new event (only containing `fileUri` and `fileSize`) in the Events Store for each chunk created.
 
 Then, for every chunk file dropped in the Staging rocket a new Lambda function will be triggered that will:
 
-- Read each chunk file line by line and persist a new event in the Events Store for each row
+- Read each chunk file line by line and persist a new event in the Events Store for each row.
 
-This event is defined in the Booster application and can be consumed from the Event Handler for any kind of processing.
+This event is registered in the Booster application's store as a regular event. Then, you can create event handlers to perform any kind of processing.
+
 You drop your file, and you implement your logic based on an event representing a line of the file.
 
 *Disclaimer: Currently the rocket supports CSV and jsonl files.* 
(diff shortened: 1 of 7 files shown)


packages/rocket-batch-file-process-aws-infrastructure/README.md
```diff
@@ -7,11 +7,17 @@ Every time a new File is uploaded to the Source S3 Bucket, a new Lambda function
...
 
-**Disclaimer:** As of now the rocket can only process CSV files.
+Then, for every chunk file dropped in the Staging rocket a new Lambda function will be triggered that will:
 
+- Read each chunk file line by line and persist a new event in the Events Store for each row
+
+This event is defined in the Booster application and can be consumed from the Event Handler for any kind of processing.
+You drop your file, and you implement your logic based on an event representing a line of the file.
+
+*Disclaimer: Currently the rocket supports CSV and jsonl files.*
```
juanjoman, line 20: Can we put in the docs an example of the event "the booster way"? For example, when I used this rocket on 0.11.3 version, I defined my event like this:
```
@Event
export class S3Changed {
  public constructor(readonly id: UUID, readonly s3uri: string, readonly fileSize: string) {}

  public entityID(): UUID {
    return this.id
  }
}
```
Will it have a different format now? Do I need 2 events now that we have 2 lambdas saving events? Which one per which case?

packages/rocket-batch-file-process-aws-infrastructure/README.md
```diff
@@ -32,9 +38,16 @@ Booster.configure('development', (config: BoosterConfig): void => {
...
     parameters: {
-      bucketName: 'test-bucket-name', 
-      chunkSize: '2',
-    }
+      config: {
+        bucketName: 'test-bucket',
+        chunkSize: '2',
+      },
+      rowEvent: {
+        entityId: 'id',
```
juanjoman, line 46: Aren't we restricting the user by adding `entityId`? I mean, we'd be supposing that the user wants to update the exact same entity all the time, no matter the S3 structure.

packages/rocket-batch-file-process-aws-infrastructure/src/batch-file-processing.ts
```diff
@@ -29,47 +36,71 @@ export class BatchFileProcessingStack {
...
     })
-    fileTriggerFunction.addEventSource(uploadEvent)
+    fileSplitterFunction.addEventSource(sourceUploadEvent)
 
     const eventsHandlerLambda = stack.node.tryFindChild('events-main') as Function
     eventsHandlerLambda.addToRolePolicy(
       createPolicyStatement([stagingBucket.bucketArn, stagingBucket.bucketArn + '/*'], ['s3:*'])
     )
+
+    const fileToLineEventFunction = new Function(stack, 'fileToLineEventFunction', {
```
juanjoman, line 76: I have a question regarding these two functions: now we send 2 events right? (one per lambda)
1. One with `fileUri` and `fileSize`, that will trigger the `fileToLineEventFunction` lambda.
2. Another one with the jsonl or csv contents (`fileToLineEventFunction`).

If this is the case, shouldn't be this lambda (2) optional? Maybe I don't need this second one because I'm updating PDFs/images/txts to my S3 and I don't want a second lambda to be deployed nor triggered.

juanjoman (changes requested):
An amazing addition to the rocket 😛. I'm just a bit worried about docs + the second lambda being mandatory, let me know your thoughts below!

packages/rocket-batch-file-process-aws-infrastructure/src/file-splitter-lambda/index.ts
```diff
@@ -16,34 +18,55 @@ export const handler = async (event: S3Event): Promise<void> => {
...
         chunkContent = []
       }
-      if (lines[itr]) {
-        chunkContent.push(lines[itr])
+      if (rows[currentRowNumber]) {
+        chunkContent.push(rows[currentRowNumber])
       }
 
-      itr = itr + 1
+      currentRowNumber = currentRowNumber + 1
```
NickSeagull, line 64: I believe this `while` loop can be separated in two parts, the processing and the side effects:

```ts
// Processing part (probably should be a function)

const isWellFormedChunk = (chunkContent: Array<unknown>) => 
  chunkContent.length > 0 && chunkContent.length % chunkSize
const chunkSize = Number(process.env.CHUNK_SIZE ?? throw new Error('CHUNK_SIZE env var not set'))
const firstRow = rows[0] ?? throw new Error('Rows to be processed arrived empty')
const contentParts = 
  rows
    .filter(isWellFormedChunk)
    .map(row => [`part-${fileNumber}-${key}`, [firstRow, ...chunkContent, ...row])

// Side effects
for (const [partKey, chunkContent] of contentParts) {
  await saveFile(partKey, chunkContent)
  await saveEvent(partKey)
}
```

NickSeagull (approved):
Looks good! I left a comment for refactoring the while loop in a more functional way, feel free to implement it, or not, depends on the urgency of this 😄 

Thank you a lot! 🙏

packages/rocket-batch-file-process-aws-infrastructure/README.md
```diff
@@ -7,11 +7,17 @@ Every time a new File is uploaded to the Source S3 Bucket, a new Lambda function
...
-- Persist the new formed chunk in the Staging Bucket.
+- Persist the new formed chunk file in the Staging Bucket.
 - Persist a new event (containing fileUri and fileSize) for each chunk in the Event Store.
 
-**Disclaimer:** As of now the rocket can only process CSV files.
+Then, for every chunk file dropped in the Staging rocket a new Lambda function will be triggered that will:
 
+- Read each chunk file line by line and persist a new event in the Events Store for each row
+
+This event is defined in the Booster application and can be consumed from the Event Handler for any kind of processing.
```
javiertoledo, line 17: I'd rephrase this a little bit for consistency:

```suggestion
This event is registered in the Booster application's event store as a regular event named `RowAdded`. Then, you can create event handlers to perform any kind of processing.
```

I'd also put an example of a full event handler processing a `RowAdded` event here.

packages/rocket-batch-file-process-aws-infrastructure/README.md
```diff
@@ -32,9 +38,16 @@ Booster.configure('development', (config: BoosterConfig): void => {
...
-      bucketName: 'test-bucket-name', 
-      chunkSize: '2',
-    }
+      config: {
+        bucketName: 'test-bucket',
+        chunkSize: '2',
+      },
+      rowEvent: {
+        entityId: 'id',
+        eventTypeName: 'RowAdded',
```
javiertoledo, line 47: When developing rockets, we have to be aware of too standard names, because it's easy that they collide with existing or future application event names. Ideally we should find a way to "scope" our rocket events, like using `BatchProcessor.RowAdded` instead of just `RowAdded`. Or an alternative If that's not possible, you get the point.

packages/rocket-batch-file-process-aws-infrastructure/src/file-to-line-event-lambda/index.ts
```diff
@@ -0,0 +1,90 @@
...
+import { DynamoDB, S3 } from 'aws-sdk'
+
+const CSV_FILE = 'csv'
+const JSONL_FILE = 'jsonl'
+
+const s3 = new S3()
+const ddb = new DynamoDB.DocumentClient()
+const { v4: uuidv4 } = require('uuid')
+
+const ENTITY_ID_NAME = process.env.ENTITY_ID!
```
javiertoledo, line 11: If we want to use environment variables in rockets, we should scope them somehow too to avoid inter-rocket collisions. Ideally, all parameters should be provided via config.

packages/rocket-batch-file-process-aws-infrastructure/src/file-to-line-event-lambda/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  for (let rowNumber = 0; rowNumber < rows.length; rowNumber++) {
+    const boosterEvent = JSON.parse(rows[rowNumber])
+    const entityIdValue = boosterEvent[ENTITY_ID_NAME]
+    await saveEvent(entityIdValue, boosterEvent)
+  }
+}
+
+async function saveEvent(entityIdValue: string, boosterEvent: any): Promise<void> {
+  const params = {
+    TableName: process.env.EVENT_STORE_NAME!,
```
javiertoledo, line 71: I wonder if we could have a better way of passing parameters than environment variables. It looks a bit fragile...

packages/rocket-batch-file-process-aws-infrastructure/src/file-splitter-lambda/index.ts
javiertoledo: One fun thing to consider is: what happens if there's an exception in one row in the middle? (i.e: the lambda could crash if the file is too long and the execution timeout is exceeded). One possible solution is having an extra DynamoDB (Or redis) table that records when a specific row is processed, so, if the process needs to be restarted, it knows where to continue.

I also have some doubts about using Lambda for this kind of workflow, because you'll have a lot of problems like this. I think that it could make a lot of sense to build a proper "job/worker" rocket where job orders, job statuses and data are persisted, the worker processes can run indefinitely (For instance, in a Fargate instance), and they can be safely stopped, restarted or scaled anytime (They'll just look at the jobs queues for things to do when they're running)

javiertoledo (commented):
Nice changes, just left a few comments/questions

packages/rocket-batch-file-process-aws-infrastructure/README.md
rdoria1: @juanjoman Yes I will update the doc to reflect his in a better way. 

In the previous version we were reacting to the Event Handler to the following event (defined in the rocket btw)

```
import { Event } from '@boostercloud/framework-core'
import { UUID } from '@boostercloud/framework-types'

@Event
export class FileAdded {
  public constructor(readonly fileURI: string, readonly filesize: number) {}

  public entityID(): UUID {
    return this.fileURI
  }
}
```

where `FileAdded` represents the chunk file.

In the current implementation you only need the event defined in your booster apps that represents one line of your file. See the example https://github.com/theam/process-jsonl-file

I need to explain this better overall :)

packages/rocket-batch-file-process-aws-infrastructure/src/batch-file-processing.ts
rdoria1: The second lambda could be optional if you want to handle with Files from your booster app. The goal is this further abstraction is exactly to deal with that.
Previous version: Dealing with events representing the chunk files. You still need the SDK to open the files, split the data etc..
Curren version: you receive an event representing one line of the file, you even forget you are dealing with files.

packages/rocket-batch-file-process-aws-infrastructure/README.md
rdoria1: I didn't get this comment @juanjoman We are passing the entityId of the event we defined in our booster app. The rocket needs to know about which attribute is the entity ID because when saving a booster event we need this piece of info https://github.com/boostercloud/booster/blob/606d13c4e0bee83b777c631f74a7c1e571724780/packages/rocket-batch-file-process-aws-infrastructure/src/file-to-line-event-lambda/index.ts#L74

packages/rocket-batch-file-process-aws-infrastructure/README.md
rdoria1: Yes totally! I will re-write the whole doc with a full end to end example. It is not clear now.

packages/rocket-batch-file-process-aws-infrastructure/README.md
rdoria1: @javiertoledo Actually the `RowAdded` comes from the Booster app and is defined by the user. I think my README is definitely not clear. I will improve it :) 

This is the example of the Booster app.

juanjoman (approved):
Ok we synched online and I thought this was [the same rocket as this one](https://github.com/boostercloud/booster/tree/main/packages/rocket-fileuploader-aws-infrastructure) so I got confused. LGTM!!

packages/rocket-batch-file-process-aws-infrastructure/src/file-to-line-event-lambda/index.ts
rdoria1: This env variable is used only in one lambda of the rocket and its scope is only that lambda function. 
Not sure what you mean by `Ideally, all parameters should be provided via config.`

packages/rocket-batch-file-process-aws-infrastructure/src/file-to-line-event-lambda/index.ts
rdoria1: The problem is we don't control the input params of the lambda as they are S3Events triggered. So the only I can think of passing them as env variables of those lambdas. 

We could setup a DB and save those as well but not sure if we are gaining much more.

packages/rocket-batch-file-process-aws-infrastructure/src/file-splitter-lambda/index.ts
rdoria1: @javiertoledo Deciding if Lambda is the right fit for this use case is of the tasks we have in mind actually.

The reasons you mentioned are not the ones I was thinking of but they totally makes sense (job order, job status) we can implement this in lambda but let's see.

As of now there are 2 things that concerns me in regards of Lambda

- How much will it cost? Is it going to be more expensive than their current solution with pods? We have a script they use with dummy data for their tests (the schema is accurate) so we'll try a few things with that

- Is the timeout going to be an issue? One way to mitigate this issue is the first lambda we have to split the file. Of course this totally dependes on the value of `chunkSize` which as of now is in total control of the user. We need to find out how long our lambdas are taking with the max number of records they handle in a single file (1 Million) We can do a test for 10K and then multiply and see. I agree this is not a stable solution as it totally depends on the schema and the `chunkSize` so depending on the case we can always hit the 15 min timeout but getting an idea of their most extreme case would give us and idea of where we are standing. 

My current reasoning is: If it turns out to be more expensive for similar workloads then we need to consider already a different alternative because they are price sensitive as our hackers already told us. If it happens to be about same or cheaper let's then check the timeout issue and see from there.

Fargate seems definitely a good alternative.

What do you think about this? @javiertoledo

Issue #627, opened 2021-02-24
Add .mocharc.yml file when "boost new:project app" + test env variables


juanjoman:
While working on a Booster application, I noticed that `npm run test` was failing because:

1. No booster env was found
2. Region is not specified

I could fix these by adding the missing env variables to the `test` script:
#### Before
```
"scripts": {
    (...)
    "test": "nyc --extension .ts mocha --forbid-only \"test/**/*.test.ts\""
  }
```
#### After
```
"scripts": {
    (...)
    "test": "AWS_SDK_LOAD_CONFIG=true BOOSTER_ENV=test nyc --extension .ts mocha --forbid-only \"test/**/*.test.ts\""
  }
```

Right after that, I received another error after running `npm run test` again: `Syntax Error: Cannot use import statement outside a module`

[After researching for a while](https://stackoverflow.com/questions/60993812/mocha-typescript-cannot-use-import-statement-outside-a-module), I found out that the generated project doesn't have the `.mocharc.yml` file, which should look like this:
```
diff: true
require: 'ts-node/register'
extension:
  - ts
package: './package.json'
recursive: true
reporter: 'spec'
timeout: 5000
full-trace: true
bail: true
```

## Expected behaviour
When running `boost new:project my_project`, the `.mocharc.yml` file should be generated. Also, `AWS_SDK_LOAD_CONFIG` and `BOOSTER_ENV` should be added to the `test` script in `package.json`

juanjoman:
Also noticed that some testing dependencies are missing:
- chai-as-promise (and the expect.ts file under `test` folder)
```typescript
// expect.ts file
import * as chai from 'chai'

chai.use(require('sinon-chai'))
chai.use(require('chai-as-promised'))

export const expect = chai.expect
```
- sinon and @types/sinon
- sinon-chai and @types/sinon-chai
- chai and @types/chai

And probably I'm missing one or two more, so it would be great to write a very simple test like this one to check:
```typescript
const mySpy = spy(<object>, <method>)
// calledOnceWith and those "advanced" methods are only accessible via the exported "export" from the `expect.ts` file
expect(mySpy).to.have.been.calledOnceWith
```

PR #642, opened 2021-03-01
[Docs] - Testing Booster applications

juanjoman:
## Description
These are some guidelines that might be useful for people testing a Booster application. It helps also fixing some of the errors from https://github.com/boostercloud/booster/issues/627

## Changes
Add docs about how to test a Booster application

## Checks
- [X] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -95,6 +95,7 @@
     - [Performing Auth requests](#performing-auth-requests)
     - [Performing GraphQL requests](#performing-graphql-requests)
     - [Inspect Database information](#inspect-database-information)
+  - [Testing Booster applications](#testing-booster-applications)
   - [Booster examples](#booster-examples)
 - [Frequently Asked Questions](#frequently-asked-questions)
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -2957,6 +2958,76 @@ Variables
...
+    register.events(new MyEvent(command.ID, readModelName))
+  }
+}
+```
+
+#### Recommended files
+These are some files that might help you speed up your testing with Booster.
+
+```typescript
+// <root_dir>/test/expect.ts
```
charlietfe, line 3007: Are we providing this? Where the expect comes from?

juanjoman: No we are not, this is a way of allowing the user to use more specific methods from `sinon-chai` and `chai-as-promised`. It's not mandatory but very useful for testing :v: @charlietfe

davidverdu:
This is only the doc, isn't it? We will probably need to create another PR to add the pending mocha files. It would be awesome that our cli generate the test file as well as the src file when you run, for example `boost new:command CreateComment`. I'll create a ticket for the latter.

juanjoman:
Yes @davidverdu this is only docs regarding this issue #627, until we fix it properly with the CLI

davidverdu (approved):

docs/README.md
```diff
@@ -2957,6 +2958,91 @@ Variables
...
+"scripts": {
+  "test": "nyc --extension .ts mocha --forbid-only \"test/**/*.test.ts\""
+}
+```
+
+The only thing that you should add to this line are the `AWS_SDK_LOAD_CONFIG=true` and `BOOSTER_ENV=test` environment variables, so the script will look like this:
+
+```typescript
+"scripts": {
+  "test": "AWS_SDK_LOAD_CONFIG=true BOOSTER_ENV=test nyc --extension .ts mocha --forbid-only \"test/**/*.test.ts\""
```
adriangmweb, line 2975: Why these env vars have to be added manually?

juanjoman: Each one solves a different "error":
- BOOSTER_ENV=test => `No environment found for your Booster application`
- AWS_SDK_LOAD_CONFIG=true => `region not found`

I'm not 100% sure why this error happens, but it's just a quick fix @adriangmweb

adriangmweb (approved):

Issue #628, opened 2021-02-24
CLI new:entity command wrong flag 


Fecony:
CLI `new:entity` command flag name was changed few times, from `--reactsTo` to `--projects`  in [commit](https://github.com/boostercloud/booster/commit/4d5a28208618561473bb2c3026b1dd79054f4c71) and from `--projects` to `--reduces` in another [commit](https://github.com/boostercloud/booster/commit/ffe7b945af7878bd2e6a385ede86796dc13f6d79), but shorthand `char` wasn't changed the last time with flag. And now it is a little bit frustrating, since it feels like `--reduces` short form should be `-r` and not `-p`
 https://github.com/boostercloud/booster/blob/96c391e039b2851bf4c04a4b8b64b57fa19e922e/packages/cli/src/commands/new/entity.ts#L29-L31

## Possible Solution

Swap `char: 'p'` to `char: 'r'`

NickSeagull:
Agree! Thanks a lot for reporting :)

davidverdu:
Closed by https://github.com/boostercloud/booster/pull/643

PR #643, opened 2021-03-01
fix(cli): update new:entity command flag

sachinh19:
## Description
This is a simple bug fix described in https://github.com/boostercloud/booster/issues/628 where the new:entity cli command flag needed an update.

## Changes
Changed the flag from p to r.

## Checks
- [x] Project Builds
- [x] Updated documentation accordingly.
 
## Additional information
I could not find anything in the documentation that needed an update as a result of this change.


NickSeagull (approved):
Thanks a lot for your contribution, this is super useful, as it led to confusion to users!

adriangmweb (approved):
Thanks for your contribution! 🙏🏻

Issue #94, opened 2020-05-25
[CLI] Remove stack trace from CLI's stdout messages


glammers1:
**Environment**

`@boostercloud/cli/0.3.3 darwin-x64 node-v12.16.3`

**Steps to reproduce it:**

You just need to generate an error executing a CLI command, for instance, having on purpose a wrong AWS credentials configuration.

**Expected result**

Stack trace could be removed from those messages because it doesn't help much the user. In fact, in previous versions like 0.2.1, there was no stack trace:

```
$ boost deploy -e production
ℹ boost deploy [production] 🚀
ℹ Bootstrapping the following environment: {"name":"Default environment","account":"155431809731","region":"us-east-1"}
✖ The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details. (You can see the full error logs in ./errors.log)
```

**Result**

```
$ boost deploy -e production
ℹ boost deploy [production] 🚀
ℹ Bootstraping the following environment: {"name":"Default environment","account":"155431809731","region":"us-east-1"}
✖ SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.
    at Request.extractError (/Users/glammers/development/theam/my-first-booster-project/node_modules/@boostercloud/framework-provider-aws-infrastructure/node_modules/aws-sdk/lib/protocol/query.js:50:29)
    at Request.callListeners (/Users/glammers/development/theam/my-first-booster-project/node_modules/@boostercloud/framework-provider-aws-infrastructure/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/Users/glammers/development/theam/my-first-booster-project/node_modules/@boostercloud/framework-provider-aws-infrastructure/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/Users/glammers/development/theam/my-first-booster-project/node_modules/@boostercloud/framework-provider-aws-infrastructure/node_modules/aws-sdk/lib/request.js:683:14)
    at Request.transition (/Users/glammers/development/theam/my-first-booster-project/node_modules/@boostercloud/framework-provider-aws-infrastructure/node_modules/aws-sdk/lib/request.js:22:10)
[...]

PR #651, opened 2021-03-02
Feature/94 remove stack trace from stdout

davidverdu:
## Description

Cli Error displaying has been improved just to print a short message in the terminal and the error stack trace in the `./errors.log` file.

Fixes: #94 

## Changes
- Cli error handled improved
- Refactoring of CLI version check to be simpler 

## Checks
- [x] Project Builds


juanjoman (approved):
LGTM! Can we write some docs about this new error file?

packages/cli/src/common/base-command.ts
```diff
@@ -1,14 +1,17 @@
...
 
-  async catch(err: any) {
-    // add any custom logic to handle errors from the command
-    // or simply return the parent class error handling
-    return super.catch(err)
+  async catch(fullError: any) {
+    const errorMessage = fullError.message.split("\n")[0].replace('Error:','')
+    const logRefMessage = "\n(You can see the full error logs in ./errors.log)"
+    const errorForFile = `\nboost ${this.id} ${this.argv.join(' ')}\n${fullError.message}`
+    appendOnErrorsFile(errorForFile)
```
NickSeagull, line 14: Should this be optional? I believe it would could get uncomfortable if the file is generated by default

NickSeagull (approved):
LGTM, just consider a small change

packages/cli/src/common/base-command.ts
davidverdu: I was thinking of adding the flag `--verbose`, making the errors be displayed on the terminal in place of the file. What do you think @NickSeagull ?

NickSeagull: I think that is much better instead of a file, if the user wants a file they can always pipe the output

NickSeagull:
Any reason this didn't get merged?

davidverdu:
> Any reason this didn't get merged?

@NickSeagull this PR dumps the stack trace into a file but here in the comments our colleagues suggested to avoid files and provide the --verbose flag to display the detailed output of the cli command. This last suggestion hasn't been applied yet to this PR. This is why it's not merged

packages/cli/src/common/base-command.ts
alvaroloes: Agree here. the `verbose` flag seems a good option

alvaroloes (approved):
Except from the `--verbose` flag, everything else LGTM

jesusAlvSoto:
@javiertoledo proposed in a [comment of the linked issue #94 ](https://github.com/boostercloud/booster/issues/94#issuecomment-679974219)  to not force the user to use the `--verbose` flag if they want to see the full stack trace, as it might not be possible to run the command again.

Can we merge this PR?

NickSeagull (approved):

javiertoledo (approved):
LGTM. It could make sense to make the file creation optional or configure the path and file name, but that can be another issue. Let's close this PR if all integration tests pass

PR #658, opened 2021-03-04
Add automatic ESLint fixing

NickSeagull:
## Description

We are manually telling all contributors to run `./scripts/check-all-the-things.sh` to run unit tests and to autoformat the code.

We are already running unit tests, why not also adding automating formatting?

## Changes

Made the Lint action now run `lint:fix` and create a new PR with changes. Not sure if this will work, because you know how the CI/CD stuff works... Trial and error....

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,10 +1,5 @@
-name: Lint
+name: Lint and Format
 on:
-  push:
-    branches:
-      - main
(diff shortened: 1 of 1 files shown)


adriangmweb (approved):
That's a great idea! 💯

javiertoledo (approved):
Interesting, let's see how this works!

PR #660, opened 2021-03-05
Fix/620 fix resource name generation

### LINKED ISSUE #620: PascalCased resource class name
## Feature Request
Adjust resource generation

## Description
When trying Booster cli, I have noticed that it doesn't change provided name for resource. 
For example, if I run `boost new:command testCommand` it will generate `test-command.ts` file with `testCommand` class name. Then, eslint will point out, that `Class 'testCommand' must be PascalCased.eslint@typescript-eslint/class-name-casing`

image

Fecony:
## Description
Fix resource name generation

## Changes
- Add method `resourceNameToClassName()` to generate class name from passed resource name.
- Add few tests for `resourceNameToClassName()` and `classNameToFileName()` functions in filenames.ts
- Adjust generator test.
- Formatting in generator test was fixed.

## Checks
- [x] Project Builds
 
## Additional information
`resourceNameToClassName()` will transform passed name to escape `' '`,`-`, `_` characters.

`test command` -> `TestCommand`
`testCommand` -> `TestCommand`
`test-command` -> `TestCommand`

Using inflected methods might be not a good idea, so if you think this should be changed to use regex or other logic, let me know.

Closes #620 

> After running `update-tocs` I noticed changes in docs readme file, that have beed done after working with docs. I assume they were made in IDE that auto formats markdown file on save and `npm run update-tocs` script was not launched. 

> Might be good idea to add commit hook to automatically run `update-tocs` script, since it is missed sometimes?


docs/README.md
```diff
@@ -3,100 +3,103 @@
...
+    + [Performing Auth requests](#performing-auth-requests)
+    + [Performing GraphQL requests](#performing-graphql-requests)
+    + [Inspect Database information](#inspect-database-information)
+  * [Testing Booster applications](#testing-booster-applications)
+    + [Testing with sinon-chai](#testing-with-sinon-chai)
+    + [Recommended files](#recommended-files)
+  * [Booster examples](#booster-examples)
 - [Frequently Asked Questions](#frequently-asked-questions)
 
 <!-- tocstop -->
```
NickSeagull, line 105: I believe this shouldn't be committed? 😅

packages/cli/src/common/filenames.ts
```diff
@@ -3,3 +3,7 @@ import * as inflected from 'inflected'
 export function classNameToFileName(name: string): string {
   return inflected.dasherize(inflected.underscore(name))
 }
+
+export function resourceNameToClassName(name: string): string {
+  return inflected.classify(inflected.underscore(inflected.parameterize(inflected.tableize(name))))
```
NickSeagull, line 8: Actual question, why not `classify` directly?

NickSeagull (commented):
Thanks!! LGTM, just some small comments!

packages/cli/src/common/filenames.ts
Fecony: `classify` itself will transform name like `test-command` to  `Test-Command`, 
`parameterize` is used to make sure that spaces are changed to `-`
And since classify is not removing `-`,  `underscore` is used to replace `-` to `_`

I think `tableize` can be removed since it looks like it is not needed

docs/README.md
Fecony: But those are docs as they should be, after running `update-tocs` command. Documentation wasn't changed, but it is not formatted as it should be... 😥

packages/cli/src/common/filenames.ts
Fecony: @NickSeagull Just checked once more, `tableize` is needed too. It is adding underscore between words `Inflector.tableize('RawScaledScorer')  // => 'raw_scaled_scorers'`
Without it, `testCommand` will transform to `testcommand` by `parameterize` method

docs/README.md
NickSeagull: Gah! I have to create another GH Action for this 😂 

Thanks for this!

packages/cli/src/common/filenames.ts
NickSeagull: I see, thanks!

NickSeagull (approved):

Fecony:
@NickSeagull I have applied last main branch changes ✨

AdrianLorenzoDev (approved):
LGTM! :)

juanjoman (approved):
LGTM!! Amazing tests :v:

Issue #661, opened 2021-03-08
Current metadata limitations


adriangmweb:
## Description

The work in #616 introduces TypeScript transformers with the new feature of reflecting the metadata of the classes by default. This adds a lot of possibilities, but it still has limitations that, once fixed, will improve the smooth usage of the Framework.

## Limitations

- Custom types like `SKU` are still lost at runtime. It's necessary to use a class instead.
- `Interfaces` are lost at runtime, so it's also necessary to convert them to classes.
- The `GraphQL` schema needs to ideally have the `types` and `interfaces` converted to classes in order to be able to use the proper type, otherwise, it will just use a generic `GraphQLJSONType`, which decreases the description and security of the filter.
- `Enum` are converted to plain objects at runtime. It would be great if there were an extra property to be able to know where the object is an `enum`.
- TypeScript custom classes like `URL`, used in `Picture`, are not being reflected by default.

## Additional information

The framework can still work without implementing those things, but the GraphQL schema will be way more secure and descriptive if could retrieve that information. Also, users won't need to use just classes when defining new elements, but they can use types and interfaces as they have been doing in the past.

adriangmweb:
Hi, @alvaroloes! I'm assigning this to you as you are currently working on this, just so you are aware of the limitations.

Issue #652, opened 2021-03-03
DynamoDB nested attributes 


adriangmweb:
# Bug Report

## Description

DynamoDB has a limitation when querying a nested attribute of an array, so let's say that we have the following schema.
```typescript
class CartItem {
  public constructor(public productId: UUID, public quantity: number) {}
}

class Cart {
  public constructor(
    readonly id: UUID,
    readonly cartItems: Array<CartItem>
  ) {}
```
And we want to use the new `GraphQL` filers to get all the carts that have the product with `id = "0a9c7881-1f1e-44cd-9e7d-cf1df886fcbd"`, knowing that there is at least one cart that meets that criteria.

So we will use the following query: 
```graphql
CartReadModels
  (cartItems: {
    includes: {
      productId: { eq: "0a9c7881-1f1e-44cd-9e7d-cf1df886fcbd" }
      quantity: { eq: 7 }
    }
  })
  ```

## Expected behavior

As we know that there is a Cart that has that product, we will get at least the given Cart.
```json
  "data": {
    "CartReadModels": [{
        "id": "957603c5-9dad-42be-a33a-438d85658c68",
        "cartItems": [
          {
            "productId": "0a9c7881-1f1e-44cd-9e7d-cf1df886fcbd",
            "quantity": 62976
          }
        ]
      }]
  }
}
  ```

## Current Behavior

The result of the query is empty, but there is at least a Cart that has a product with that id.
```json
  "data": {
    "CartReadModels": []
  }
}
  ```
## Possible Solution

There is not a solid solution for this as it's a known limitation of `DynamoDB` that doesn't allow filtering a `Map` within a `List`. **Yeah, I know**🤦🏻‍♂️ 

 [This is the AWS thread about it.](https://forums.aws.amazon.com/thread.jspa?messageID=585008#585008)

There is a _ workaround_ using the `contain` function, but then the value inside the filter `includes` **must be the exact same object** instead of just a property, and we won't be able to use the filters like `eq, ne, gt, etc`. 

Of course, there is also the option of just using a list of IDs and the filter by those IDs.

**Any ideas about it?**

### Environment
- Booster version: Version in development

adriangmweb:
As a partial solution, I changed the `GraphQL` schema for arrays from: 
```graphql
input cartItemsPropertyFilter {
  includes: CartItemPropertyFilter
}

input CartItemPropertyFilter {
  productId: StringPropertyFilter
  quantity: NumberPropertyFilter
[...]

Issue #603, opened 2021-02-05
Document new filter DSL


adriangmweb:
## Description
Create detailed documentation about the new ReadModels filtering DSL with all the available keywords and their meaning

Issue #602, opened 2021-02-05
Migrate aws subscriptions to the new GraphQL filters


adriangmweb:
## Description
When you subscribe to a ReadModel, you use the same filter as when querying them. We need to implement the behavior of those filters (as you did in AWS) but, this time, by filtering the array of reading models in memory, in pure javascript (instead of with AWS SQL).

The GraphQL engine will call this code with an array of ReadModels of a specific type that is eligible to send to the user who subscribed, but we need to filter them using the filters the user specified in the Subscription operation.

PR #662, opened 2021-03-08
New graphql schema

adriangmweb:
## Description

_Note: This branch has been extended from #616 as the metadata reflection introduced by the TypeScript transformers is necessary. Unfortunately, this has its limitations as stated in #661._

This branch introduces the GraphQL new schema in the core, allowing filtering by more than one property at the same time and by complex properties as well in queries and subscriptions.

## Changes

The GraphQL schema now is way more secure and complete, allows complex and multiple filters instead of just one basic one. Also, the filter API is way more user-friendly.

## Checks
- [x] Project Builds
 
## Additional information

This PR solves: #600 #602 and also #603 

Due to a `DynamoDB` limitation, the schema has to make use of generic `GraphQLJSONType` types, which doesn't give much info to the user and doesn't give us full security when updating types. This also doesn't give us the ability yet to query by properties of a complex property that is inside a `List`. You can see more information about this limitation here #652 

## Next steps
Refactor the rest of the providers to use the new GraphQL schema, see #601


packages/framework-integration-tests/src/common/address.ts
```diff
@@ -1,8 +1,10 @@
-export interface Address {
-  firstName: string
-  lastName: string
-  country: string
-  state: string
-  postalCode: string
-  address: string
+export class Address {
```
juanjoman, line 1: Why are these classes now?

juanjoman (approved):
LGTM!! Just left some minor questions :v:

PS: The docs are amazing <3

packages/framework-integration-tests/src/common/address.ts
adriangmweb: Because sadly right now the `interfaces`, `types`, and `TypeScript` custom classes are still lost at runtime, so only the user classes are being reflected. You can see the issue with the limitations that would be great to support in the future here #661.

packages/framework-core/src/services/graphql/common.ts
```diff
@@ -1,7 +1,8 @@
 import { GraphQLList, GraphQLScalarType, GraphQLObjectType, GraphQLType } from 'graphql/type/definition'
-import { AnyClass, PropertyMetadata, UserEnvelope, UUID, GraphQLOperation } from '@boostercloud/framework-types'
+import { AnyClass, UserEnvelope, UUID, GraphQLOperation } from '@boostercloud/framework-types'
 import { GraphQLFieldResolver } from 'graphql'
 import { ReadModelPubSub } from '../pub-sub/read-model-pub-sub'
+import { PropertyMetadata } from 'metadata-booster'
```
AdrianLorenzoDev, line 5: Shouldn't `metadata-booster` be inside the `@boostercloud` group? Why did you choose to add it outside that package group?

AdrianLorenzoDev (approved):
LGTM!

docs/README.md
```diff
@@ -1582,6 +1582,142 @@ For more information about queries and how to use them, please check the [GraphQ
...
+| le         |  String  |        Lower or equal than |
+| in         | [String] |      Exists in given array |
+| beginsWith |  String  | Starts with a given substr |
+| contains   |  String  |    Contains a given substr |
+
+Example:
+
+```graphql
+query {
+  ProductReadModels(filter: { sku: { begingsWith: "jelwery" } }) {
```
NickSeagull, line 1671:
```suggestion
  ProductReadModels(filter: { sku: { begingsWith: "jewelry" } }) {
```

docs/README.md
```diff
@@ -1582,6 +1582,142 @@ For more information about queries and how to use them, please check the [GraphQ
 
 Booster GraphQL API also provides support for real-time updates using subscriptions and websocket, to get more information about it go to the [GraphQL API](#subscribing-to-read-models) section.
 
+#### Filtering a read model
+
+Booster GraphQL API provides support for filtering ReadModels on queries, subscriptions and at code level.
```
NickSeagull, line 1587:
```suggestion
The Booster GraphQL API provides support for filtering Read Models on queries, subscriptions and at code level.
```

docs/README.md
```diff
@@ -1582,6 +1582,142 @@ For more information about queries and how to use them, please check the [GraphQ
...
+
+Using the GraphQL API endpoint you can retrieve the schema of your application so you can see what are the filters for every ReadModel and its properties. You can filter like this:
```
NickSeagull, line 1589:
```suggestion
Using the GraphQL API endpoint you can retrieve the schema of your application so you can see what are the filters for every Read Model and its properties. You can filter like this:
```

docs/README.md
```diff
@@ -1582,6 +1582,142 @@ For more information about queries and how to use them, please check the [GraphQ
...
+
+Searching for a specific ReadModel by id
```
NickSeagull, line 1591:
```suggestion
Searching for a specific Read Model by `id`
```

docs/README.md
```diff
@@ -1582,6 +1582,142 @@ For more information about queries and how to use them, please check the [GraphQ
...
+    sku
+    availability
+    price
+  }
+}
+```
+
+#### Supported filters
+
+The currently supported filters are in the following ones. |
```
NickSeagull, line 1606:
```suggestion
The currently supported filters are the following ones:
```

docs/README.md
```diff
@@ -1582,6 +1582,142 @@ For more information about queries and how to use them, please check the [GraphQ
...
+
+##### Boolean filters
+
+| Filter |   Value    |  Description |
+| :----- | :--------: | -----------: |
+| eq     | true/false |     Equal to |
+| ne     | true/false | Not equal to |
```
NickSeagull, line 1613:
```suggestion
| eq     | boolean |     Equal to |
| ne     | boolean | Not equal to |
```

docs/README.md
```diff
@@ -1582,6 +1582,142 @@ For more information about queries and how to use them, please check the [GraphQ
...
+_Note: Right now there is a limitation with complex properties in Arrays, where you just can filter them if you know the exact value of an element but is not possible to filter from a property of the element. As a workaround, you can use an array of ids of the complex property and filter for that property as in the example above._
+
+##### Filter combinators
+
+All the filters can be combined to create a more complex search on the same properties of the ReadModel.
+| Filter | Value | Description |
+| :----- | :--------: | -----------: |
+| and | [Filters] | AND - all filters on the list have a match |
+| or | [Filters] | OR - At least one filter of the list has a match |
+| not | Filter/and/or | The element does not match the filter
```
NickSeagull, line 1707: This should be formatted probably

packages/cli/src/templates/project/package-json.ts
```diff
@@ -37,7 +39,7 @@ export const template = `{
   "scripts": {
     "lint:check": "eslint --ext '.js,.ts' **/*.ts",
     "lint:fix": "eslint --quiet --fix --ext '.js,.ts' **/*.ts",
-    "compile": "npx tsc -b tsconfig.json",
+    "compile": "npx ttypescript -b tsconfig.json",
```
NickSeagull, line 42: `[rant]`

Whyyyyyy 😭😭😭😭 Would've been better to use `ts-patch`...

Anyways, go on... 🥲

`[/rant]`

packages/framework-core/src/decorators/metadata.ts
```diff
@@ -1,22 +1,14 @@
...
 import 'reflect-metadata'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function getPropertiesMetadata(classType: Class<any>): Array<PropertyMetadata> {
-  const propertyNames = Object.getOwnPropertyNames(new classType())
-  const propertyTypes = Reflect.getMetadata('design:paramtypes', classType)
-  if (propertyNames.length != propertyTypes.length) {
+  const meta: ClassMetadata = Reflect.getMetadata('booster:typeinfo', classType)
+  if (!meta) {
     // eslint-disable-next-line prettier/prettier
```
NickSeagull, line 9: I believe this line is not needed anymore
```suggestion
```

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -141,75 +131,120 @@ export class GraphQLQueryGenerator {
...
-    const filterName = `${type.name}PropertyFilter`
     if (!this.generatedFiltersByTypeName[filterName]) {
-      const graphQLValueType = this.typeInformer.getGraphQLTypeFor(type)
+      const propFilters: GraphQLInputFieldConfigMap = {}
+      property.typeInfo.parameters.forEach(() => {
+        propFilters.includes = {
+          ...propFilters.includes,
+          ...{
+            type: GraphQLJSONObject,
+          },
```
NickSeagull, line 170: Why doing this and not instead:
```suggestion
          type: GraphQLJSONObject,
```

packages/framework-core/src/services/graphql/graphql-type-informer.ts
```diff
@@ -107,4 +141,13 @@ export class GraphQLTypeInformer {
     }
     return inputFields
   }
+
+  private generateOperationEnumValuesFor(operationsEnum: AnyClass): GraphQLEnumValueConfigMap {
+    const enumValuesConfig: GraphQLEnumValueConfigMap = {}
+    for (const opSymbol in operationsEnum) {
+      const opName = (operationsEnum as any)[opSymbol]
```
NickSeagull, line 148: `any` police 👮🏻‍♂️

What's the contents of `operationsEnum`? Maybe we can type it in a better way? Can you explain more please?

NickSeagull (changes requested):
Thanks a lot! Left some suggestions

packages/cli/src/templates/project/package-json.ts
adriangmweb: I just extended form the existing branch, we'll have plenty of time to change it in the future 🙌🏻

packages/framework-core/src/services/graphql/graphql-type-informer.ts
adriangmweb: I didn't create that function, it was added after a rebase, so there isn't much I can tell you here 😅

NickSeagull: Ok, keep circulating 👮

NickSeagull (approved):

docs/README.md
```diff
@@ -1566,6 +1566,143 @@ For more information about queries and how to use them, please check the [GraphQ
...
+Example:
+
+```graphql
+query {
+  CartReadModels(filter: { or: [{ id: { contains: "a" } }, { id: { contains: "b" } }] }) {
+    id
+    price
+    itemsIds
+  }
+}
```
davidverdu, line 1703: If it is supported, I would also add an example of combining `and` and `or` operators. Good documentation by the way!

davidverdu (approved):
LGTM. Good change!! Good tests!!! Good documentation!!
- I tested it with Asker and PostComment sample applications and it works
- I left a minor comment regarding documentation
- I agree with the other comments

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -141,75 +131,118 @@ export class GraphQLQueryGenerator {
...
+        }
+
+        fields = () => ({
+          ...nestedProperties,
+          and: { type: new GraphQLList(this.generatedFiltersByTypeName[filterName]) },
+          or: { type: new GraphQLList(this.generatedFiltersByTypeName[filterName]) },
+          not: { type: this.generatedFiltersByTypeName[filterName] },
+        })
+      } else {
+        fields = this.generateFilterInputTypes(type)
```
alvaroloes, line 205: If `this.typeInformer.canFilter` says the propType can be filtered, why are we generating filters here? Maybe I didn't understood something

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -141,75 +131,118 @@ export class GraphQLQueryGenerator {
...
+
+      if (!this.typeInformer.canFilter(graphQLPropType)) {
+        const properties = getPropertiesMetadata(type)
+        this.typeInformer.generateGraphQLTypeFromMetadata({ class: type, properties })
+
+        let nestedProperties: GraphQLInputFieldConfigMap = {}
+        for (const prop of properties) {
+          const propPrimitiveType = this.typeInformer.getPrimitiveExtendedType(prop.typeInfo.type)
+          const property = { [prop.name]: { type: this.generateFilterFor(propPrimitiveType) } }
+          nestedProperties = { ...nestedProperties, ...property }
```
alvaroloes, line 195: I'm a bit confused about the use of ellipsis here. Would this code be similar to:
```suggestion
          const propPrimitiveType = this.typeInformer.getPrimitiveExtendedType(prop.typeInfo.type)
          nestedProperties[prop.name] = { type: this.generateFilterFor(propPrimitiveType) }
```

packages/framework-core/src/services/graphql/graphql-query-generator.ts
alvaroloes: Oh! `canFilter` tells whether the type is complex or primitive?

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -141,75 +131,118 @@ export class GraphQLQueryGenerator {
...
+          eq: { type: GraphQLFloat },
+          ne: { type: GraphQLFloat },
+          lte: { type: GraphQLFloat },
+          lt: { type: GraphQLFloat },
+          gte: { type: GraphQLFloat },
+          gt: { type: GraphQLFloat },
+          in: { type: GraphQLList(GraphQLFloat) },
+        }
+      case Date:
+      case String:
```
alvaroloes, line 231: Does this work with UUID? (probably yes, but asking for confirmation)

packages/framework-core/src/services/graphql/graphql-type-informer.ts
```diff
@@ -41,25 +44,56 @@ export class GraphQLTypeInformer {
...
+            const properties = getPropertiesMetadata(param.type)
+            this.generateGraphQLTypeFromMetadata({ class: param.type, properties })
+          }
+
+          fields[prop.name] = {
+            type: GraphQLList(graphQLPropType),
+          }
+        })
+      } else {
+        if (!prop.typeInfo.type.prototype) {
```
alvaroloes, line 61: I loved you included support for enums! But my question is Why this condition means the type is an enum?

packages/framework-core/src/services/graphql/graphql-type-informer.ts
```diff
@@ -41,25 +44,56 @@ export class GraphQLTypeInformer {
...
+        } else {
+          fields[prop.name] = { type: this.getGraphQLTypeFor(prop.typeInfo.type) }
+        }
+      }
     }
     return fields
   }
 
+  public canFilter(graphQLType: GraphQLNonInputType): boolean {
+    return graphQLType instanceof GraphQLScalarType && graphQLType != GraphQLJSONObject
```
alvaroloes, line 77: Can't we also filter complex objects?

packages/framework-core/src/services/graphql/graphql-type-informer.ts
```diff
@@ -107,4 +141,13 @@ export class GraphQLTypeInformer {
...
   }
+
+  private generateOperationEnumValuesFor(operationsEnum: AnyClass): GraphQLEnumValueConfigMap {
+    const enumValuesConfig: GraphQLEnumValueConfigMap = {}
+    for (const opSymbol in operationsEnum) {
+      const opName = (operationsEnum as any)[opSymbol]
+      enumValuesConfig[opName] = { value: opSymbol }
+    }
+    return enumValuesConfig
+  }
```
alvaroloes, line 152: Is this function still needed? It was used for the old filteres

packages/framework-core/test/services/graphql/graphql-type-informer.test.ts
```diff
@@ -98,10 +98,10 @@ describe('GraphQLTypeInformer', () => {
         expect(result).to.be.equal(GraphQLBoolean)
       })
 
-      it('should return GraphQLList(GraphQLJSONObject)', () => {
+      it('should return GraphQLJSONObject', () => {
         const result = sut.getGraphQLTypeFor(Array)
 
-        expect(result).to.be.deep.equal(GraphQLList(GraphQLJSONObject))
+        expect(result).to.be.deep.equal(GraphQLJSONObject)
```
alvaroloes, line 104: Why not a GraphQLList of JSON objects?

alvaroloes (commented):
Left some comments. Mostly questions to understand it better.

I have one concern: There are no integration tests for:
- Query with filters
- Subscriptions with filters

Is there any chance you can add them? I would be a bit worried if those are not included in `main`

packages/framework-core/src/services/graphql/graphql-query-generator.ts
adriangmweb: exactly, I'll rename it to avoid confusion

packages/framework-core/src/services/graphql/graphql-query-generator.ts
adriangmweb: yes, thanks to `this.typeInformer.getPrimitiveExtendedType` it knows that it extends from `String`

packages/framework-core/src/services/graphql/graphql-type-informer.ts
adriangmweb: Because it's the only case the transformer returns an Enum. There is no other way of knowing if it's an enum, but this will be refactored later once we improve the transformer.

packages/framework-core/src/services/graphql/graphql-type-informer.ts
adriangmweb: yes, this just checks if it's a primitive type. I will rename the function.

packages/framework-core/src/services/graphql/graphql-type-informer.ts
adriangmweb: It's still used when there is an Enum

PR #663, opened 2021-03-08
Fix/631 deleteAllSubscriptions pagination

### LINKED ISSUE #631: Page deleteAllSubscription queries and batchWrite them
We are doing a Query + a BatchWrite operation with its result. So we should:

* Get all items we can on a single page with the Query.
* Split the result into batches of max 25 items and send the BatchWrite operation for each of them.
* Get the next Query page of items and start again.

NickSeagull:
## Description

Made deleting of subscriptions behave accordingly to DynamoDB limits. Also removed some `sleeps`

## Checks
- [x] Project Builds
- [ ] ~Updated documentation accordingly~ (No need to)
 
## Additional information

Closes #631

### DIFF
--- a/packages/framework-provider-aws/src/library/subscription-adapter.ts
+++ b/packages/framework-provider-aws/src/library/subscription-adapter.ts
@@ -2,8 +2,9 @@ import { BoosterConfig, Logger, SubscriptionEnvelope } from '@boostercloud/frame
 import { DynamoDB } from 'aws-sdk'
 import { PromiseResult } from 'aws-sdk/lib/request'
 import { AWSError } from 'aws-sdk/lib/error'
-import { subscriptionsStoreAttributes } from '../constants'
+import { dynamoDbBatchWriteLimit, subscriptionsStoreAttributes } from '../constants'
 import { sortKeyForSubscription } from './partition-keys'
+import { inChunksOf } from '../pagination-helpers'
 
 export interface SubscriptionIndexRecord {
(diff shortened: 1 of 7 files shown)


packages/framework-provider-aws/src/library/subscription-adapter.ts
```diff
@@ -71,40 +74,46 @@ export async function deleteSubscription(
...
+    await db
+      .delete({
+        TableName: config.resourceNames.subscriptionsStore,
+        Key: {
+          [subscriptionsStoreAttributes.partitionKey]: subscriptionToDelete[subscriptionsStoreAttributes.partitionKey],
+          [subscriptionsStoreAttributes.sortKey]: subscriptionToDelete[subscriptionsStoreAttributes.sortKey],
+        },
+      })
+      .promise()
+  } while (result.LastEvaluatedKey)
```
alvaroloes, line 113: Why do we need this loop here? I think we are dealing with only one subscription and we are querying with the full primary key, so there won't be any need for pagination here (as far as I can tell)

NickSeagull: Because DynamoDB will run the query for the first 100 items, and will return something if its in that page, we have to keep querying until we run out of items to query. If not, it could be that we run the query, its not on the first 100 items, but maybe on the next, but we think its not subscribed.

Not sure if I explained this properly.

adriangmweb (approved):

NickSeagull:
Closing as stale

PR #670, opened 2021-03-13
fix(azure): Azure cosmosDB error at deployment

adriangmweb:
## Description

This fixes an error that doesn't allow deployments to Azure.

## Changes

It sets the `cosmosClient` to `{}` when the `cosmosDbConnectionString` environment variable doesn't exist because it's only supposed to exist at runtime.

## Checks
- [x] Project Builds
 
## Additional information

Azure deployments are working now! 🚀

### DIFF
--- a/packages/framework-provider-azure/src/index.ts
+++ b/packages/framework-provider-azure/src/index.ts
@@ -68,10 +68,8 @@ export const Provider = (): ProviderLibrary => ({
     rawToEnvelope: undefined as any,
   },
   // ProviderInfrastructureGetter
-  infrastructure: () => {
-    console.log('puf')
-    return require(require('../package.json').name + '-infrastructure').Infrastructure as ProviderInfrastructure
-  },
+  infrastructure: () =>
+    require(require('../package.json').name + '-infrastructure').Infrastructure as ProviderInfrastructure,
 })
 
 function notImplemented(): void {}
--- a/packages/framework-provider-azure/src/index.ts
+++ b/packages/framework-provider-azure/src/index.ts
@@ -13,10 +13,12 @@ import { environmentVarNames } from './constants'
 import { fetchReadModel, storeReadModel } from './library/read-model-adapter'
 import { searchReadModel } from './library/searcher-adapter'
 
+let cosmosClient: CosmosClient
 if (typeof process.env[environmentVarNames.cosmosDbConnectionString] === 'undefined') {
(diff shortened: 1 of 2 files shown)


packages/framework-provider-azure/src/index.ts
```diff
@@ -66,8 +68,10 @@ export const Provider = (): ProviderLibrary => ({
     rawToEnvelope: undefined as any,
   },
   // ProviderInfrastructureGetter
-  infrastructure: () =>
-    require(require('../package.json').name + '-infrastructure').Infrastructure as ProviderInfrastructure,
+  infrastructure: () => {
+    console.log('puf')
```
Fecony, line 72: @adriangmweb I think you forgot this 😄

adriangmweb: Oh my! 😅  thanks for spotting it mate!

rdoria1 (approved):
LGTM! Thanks for figuring this out. I was not aware of the change in the deploy process.

rdiaz82 (approved):
LGTM

packages/framework-provider-azure/src/index.ts
```diff
@@ -13,10 +13,12 @@ import { environmentVarNames } from './constants'
 import { fetchReadModel, storeReadModel } from './library/read-model-adapter'
 import { searchReadModel } from './library/searcher-adapter'
 
+let cosmosClient: CosmosClient
 if (typeof process.env[environmentVarNames.cosmosDbConnectionString] === 'undefined') {
```
NickSeagull, line 17: Why is this a string `'undefined'`?

NickSeagull (approved):
Approving, just left a small question

alvaroloes (approved):
Curiosity question: Do you know why this started to happen? 

As I see it, this error should always have been happening on deployment. Why was it not happening before?

packages/framework-provider-azure/src/index.ts
adriangmweb: Because it was done fast 😂. But it should just check if the env var exists.

adriangmweb:
> Curiosity question: Do you know why this started to happen?
> 
> As I see it, this error should always have been happening on deployment. Why was it not happening before?

Honestly, I agree with you. I'm not sure why it worked like that, but I suppose that some months ago the deployment process was using the infrastructure package directly.

PR #671, opened 2021-03-14
fix to allows service to get the ip from hostname

rdiaz82:
## Description
In order to make the Kubernetes provider compatible with AWS EKS, we discovered that the services IP is provided using an AWS url rather than an IP. This IP is provided in a field called `hostname` under the ingress configuration. Previously we were using the property `ip` because Google Cloud clusters provide the service IP in this property. 

## Changes
In this PR we are simply adding the support to get the service URL from the hostname in case that the IP field is empty

## Checks
- [X] Project Builds
- [ ] Updated documentation accordingly (Not necessary)


adriangmweb (approved):

javiertoledo (approved):
🚢 🇮🇹 ❗

PR #682, opened 2021-03-16
Fix/627 Add mocha file + test env variables

### LINKED ISSUE #627: Add .mocharc.yml file when "boost new:project app" + test env variables

Moneyba:
## Description
Add `.mocharc.yml` file when `boost new:project app` and add `BOOSTER_ENV` to test script in `package.json`.

## Checks
- [x] Project Builds
 
## Additional information

Closes #627

### DIFF
--- a/packages/cli/src/services/project-initializer.ts
+++ b/packages/cli/src/services/project-initializer.ts
@@ -11,6 +11,7 @@ import * as packageJson from '../templates/project/package-json'
 import * as configTs from '../templates/project/config-ts'
 import * as indexTs from '../templates/project/index-ts'
 import * as prettierRc from '../templates/project/prettierrc-yaml'
+import * as mochaRc from '../templates/project/mocharc-yml'
 import { wrapExecError } from '../common/errors'
 import { installAllDependencies } from './dependencies'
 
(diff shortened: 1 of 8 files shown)


javiertoledo (approved):
LGTM!

davidverdu (approved):
LGTM. Remember to lock conversation to launch integration tests in GitHub. If all tests pass, you can merge

PR #698, opened 2021-03-17
Fix azure provider infra

adriangmweb:
## Description

It fixes some issues when trying to deploy a serverless Booster app into Azure.

## Changes

- The deployment zip file just contains the `.deploy/` directory content instead of all the root folder, so it's more than 10 times lighter. This avoids an issue where sometimes the upload was failing.

- Removed catch from the `deployApp` and `nukeApp` functions because it was not throwing the errors, so it was not possible to know when there was an issue.

- Disable CosmosDB free tier flag to be able to deploy several apps.

## Checks
- [x] Project Builds
 
## Additional information

Azure Serverless deployments are working and allow performing GraphQL queries and mutations. 🚀

### DIFF
--- a/packages/framework-provider-azure-infrastructure/src/infrastructure/arm-templates/cosmos-db-account.ts
+++ b/packages/framework-provider-azure-infrastructure/src/infrastructure/arm-templates/cosmos-db-account.ts
@@ -46,7 +46,7 @@ export const template = {
         ipRangeFilter: '',
         dependsOn: [],
         capabilities: [],
-        enableFreeTier: true,
+        enableFreeTier: false,
       },
       resources: [
         {
--- a/packages/framework-provider-azure-infrastructure/src/infrastructure/index.ts
+++ b/packages/framework-provider-azure-infrastructure/src/infrastructure/index.ts
@@ -8,11 +8,9 @@ import {
   createWebSiteManagementClient,
 } from './setup'
 
-export const deploy = (configuration: BoosterConfig, logger: Logger): Promise<void> =>
-  deployApp(logger, configuration).catch(logger.error)
+export const deploy = (configuration: BoosterConfig, logger: Logger): Promise<void> => deployApp(logger, configuration)
 
-export const nuke = (configuration: BoosterConfig, logger: Logger): Promise<void> =>
-  nukeApp(logger, configuration).catch(logger.error)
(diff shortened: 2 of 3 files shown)


packages/framework-provider-azure-infrastructure/src/infrastructure/arm-templates/cosmos-db-account.ts
```diff
@@ -46,7 +46,7 @@ export const template = {
         ipRangeFilter: '',
         dependsOn: [],
         capabilities: [],
-        enableFreeTier: true,
```
rdoria1, line 49: As context we needed to do this because we can only have one Cosmos DB account with free tier but since we have one account per booster app we can't deploy more than one booster app so we needed to change this in order to test in parallel

packages/framework-provider-azure-infrastructure/src/infrastructure/index.ts
```diff
@@ -8,11 +8,9 @@ import {
   createWebSiteManagementClient,
 } from './setup'
 
-export const deploy = (configuration: BoosterConfig, logger: Logger): Promise<void> =>
-  deployApp(logger, configuration).catch(logger.error)
```
rdoria1, line 12: this catch mas masking all the errors, we were debugging kind of blindly.

packages/framework-provider-azure-infrastructure/src/infrastructure/utils.ts
```diff
@@ -45,13 +45,13 @@ export async function buildResource(
...
 
 export async function packageAzureFunction(functionDefinitions: Array<FunctionDefinition>): Promise<any> {
-  const output = fs.createWriteStream(path.join(os.tmpdir(), 'example.zip'))
+  const output = fs.createWriteStream(path.join(os.tmpdir(), 'functionApp.zip'))
   const archive = archiver('zip', {
     zlib: { level: 9 }, // Sets the compression level.
   })
 
   archive.pipe(output)
-  archive.glob('**/*')
```
rdoria1, line 54: This was the hardest issue to solve. 
The deployment were working sometimes. The times it failed the errors were different. IT was really weird. It turns out that the zip function deployment to azure storage was not ending because not all the packages in the nodes modules were copied (38 MB) We don't need to copy everything only what is in the .deploy folder. The zip got smaller and we got a bunch of successful deployments consistently now.

rdoria1 (approved):
LGTM

PR #699, opened 2021-03-17
fix graphql body bug. Return application/json content type

davidverdu:
## Description
return the correct body in k8s implementation

## Changes
- update Express server handler
- update GraphQL handler

## Checks
- [x ] Project Builds

### DIFF
--- a/packages/framework-provider-kubernetes/src/library/graphql-adapter.ts
+++ b/packages/framework-provider-kubernetes/src/library/graphql-adapter.ts
@@ -31,5 +31,5 @@ export const handleResult = async (
     ...headers,
   },
   statusCode: 200,
-  body: result ? result : undefined,
+  body: result
 })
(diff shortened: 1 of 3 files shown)


packages/framework-provider-kubernetes/src/library/graphql-adapter.ts
```diff
@@ -31,5 +31,5 @@ export const handleResult = async (
     ...headers,
   },
   statusCode: 200,
-  body: result ? JSON.stringify(result) : '',
+  body: result ? result : undefined,
```
NickSeagull, line 34: If you're returning undefined you can avoid this:
```suggestion
  body: result,
```

NickSeagull (approved):

rdiaz82 (approved):
LGTM

packages/framework-provider-kubernetes/src/library/graphql-adapter.ts
davidverdu: I though it was null if you don't use undefined.

Issue #601, opened 2021-02-05
Add new GraphQL filters to pending providers


adriangmweb:
## Feature Request
In order to have the same ReadModel filtering system in all the providers and deprecating the `FilterOld` type, all pending providers have to be migrated to the new structure as it's being done with AWS in the [PR 573](https://github.com/boostercloud/booster/pull/573)

## Description

To make use of the new `Filter` from the `searcher` class of the `framework-types` package, all providers `searcher-adapters` have to be refactored.

## Pending providers

- [x] [AWS](https://github.com/boostercloud/booster/pull/573)
- [x] Local provider
- [x] Azure
- [x] K8s
- [x] Subscriptions

## Additional tasks
The deprecated `filterOld` type has to be removed once all the providers are migrated to the new filters system.

adriangmweb:
The azure provider was partially refactored in #702 with the basic filters. It still needs to add the rest of them.

adriangmweb:
Ended up adding the complex filters on #702, so the Azure provider is finished! 🎉

adriangmweb:
Kubernetes provider has been included. But subscriptions need to be refactored as well, as it's stated on #751

PR #702, opened 2021-03-17
refactor searcher-adapter to use the new filters

adriangmweb:
## Description

Refactor to make Azure search filters compatible with the new version.

## Changes
Refactor to Azure provider `searcher-adapter.ts` file.

## Checks
- [x] Project Builds
~- [ ] Updated documentation accordingly~ not necessary
 
## Additional information

~This just makes it compatible with the simple filters, the complex filters, Arrays filters, and the filter combinators like `and`, `or` and `not` are still pending. So it doesn't finish Azure at #601 yet.~ **I ended up adding them so the PR has support for all the filters.**


javiertoledo (approved):
LGTM

adriangmweb:
I finally added complex properties and filter combinators to the filters. I'll update the description!

rdoria1 (approved):
LGTM

PR #703, opened 2021-03-18
generate a redis key for events and snapshots

davidverdu:
## Description
generate a more appropriate Redis key to store events and snapshots

## Changes
- how we generate the key

## Checks
- [x ] Project Builds

### DIFF
--- a/packages/framework-provider-kubernetes/src/services/event-registry.ts
+++ b/packages/framework-provider-kubernetes/src/services/event-registry.ts
@@ -8,7 +8,7 @@ export class EventRegistry {
   public async store(event: EventEnvelope, logger: Logger): Promise<void> {
     const stateUrl = `${this.url}/v1.0/state/statestore`
     logger.debug('About to post', event)
-    const data = [{ key: event.entityID, value: event }]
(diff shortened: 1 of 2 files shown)


rdiaz82 (approved):
LGTM

packages/framework-provider-kubernetes/src/services/event-registry.ts
```diff
@@ -32,4 +32,19 @@ export class EventRegistry {
     // TODO: Implement accordingly to Dapr
     throw new Error('EventRegistry#queryLatest: Not implemented yet')
   }
+
+  private eventKey(event: EventEnvelope): string {
+    //ee_Post_95ddb544-4a60-439f-a0e4-c57e806f2f6e_event_PostCreated_2021-03-17T16:49:29.143Z_j9qn8kd8
+    const hash = Math.random().toString(36).substring(5)
```
NickSeagull, line 38: This is not 100% random, and it could be possible that in the same ms two similar values could appear. Can you replace with UUID.generate() ?

NickSeagull (approved):

PR #707, opened 2021-03-18
start to create redis adapter

davidverdu:
## Description
create a Redis class to communicate with Redis server directly

## Changes
- Create RedisAdapter class with keys method uncompleted

## Checks
- [ ] Project Builds

### DIFF
--- a/packages/framework-provider-kubernetes/src/services/event-registry.ts
+++ b/packages/framework-provider-kubernetes/src/services/event-registry.ts
@@ -2,11 +2,14 @@
 import { EventEnvelope, Logger } from 'framework-types/dist'
 import { v4 as uuidv4 } from 'uuid';
 import fetch from 'node-fetch'
+//import { RedisAdapter } from './redis-adapter'
 
 export class EventRegistry {
   constructor(readonly url: string) {}
 
   public async store(event: EventEnvelope, logger: Logger): Promise<void> {
+    //await RedisAdapter.build().keys('*', logger)
+
     const stateUrl = `${this.url}/v1.0/state/statestore`
     logger.debug('About to post', event)
     const data = [{ key: this.eventKey(event), value: event }]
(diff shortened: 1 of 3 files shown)


packages/framework-provider-kubernetes/src/services/event-registry.ts
```diff
@@ -2,11 +2,14 @@
...
 import { v4 as uuidv4 } from 'uuid';
 import fetch from 'node-fetch'
+//import { RedisAdapter } from './redis-adapter'
 
 export class EventRegistry {
   constructor(readonly url: string) {}
 
   public async store(event: EventEnvelope, logger: Logger): Promise<void> {
+    //await RedisAdapter.build().keys('*', logger)
+
```
NickSeagull, line 12: Left some commented code

packages/framework-provider-kubernetes/src/services/redis-adapter.ts
```diff
@@ -0,0 +1,51 @@
...
+import fetch from 'node-fetch'
+const redis = require("redis");
+
+export class RedisAdapter {
+
+    constructor(readonly daprUrl: string, readonly redisUrl: string) {}
+
+    public static build() : RedisAdapter {
+        const redisPwd: string = process.env['DB_PASSWORD'] || 'password'
+        const redisUrl: string = `redis://redis-master:6379?password=${redisPwd}`
```
NickSeagull, line 11: Should this be hardcoded? Maybe we have to pass the URL with env variable?

NickSeagull (approved):

packages/framework-provider-kubernetes/src/services/event-registry.ts
davidverdu: I left it intended to know how to use it. I'll remove it on next PR

packages/framework-provider-kubernetes/src/services/redis-adapter.ts
davidverdu: Yes, the host could be passed as an env variable

rdiaz82: upss!! my fault! totally agree it can be also passed as an ENV var. @davidverdu let's suppose that we have a `DB_HOST` passing the host + port and I'll add it in my next PR, so the line would be: `redis://${dbHost}?....`

rdiaz82 (approved):
LGTM

PR #709, opened 2021-03-19
working on mutations for k8s provider

rdiaz82:
- Added the dbHost into the K8s Booster template
- Started to work on mutations

### DIFF
--- a/packages/framework-provider-kubernetes/src/services/event-registry.ts
+++ b/packages/framework-provider-kubernetes/src/services/event-registry.ts
@@ -51,7 +51,7 @@ export class EventRegistry {
       event.kind, // 'event' | 'snapshot'
       event.typeName, // 'PostCreated' event name
       event.createdAt, // timespan
-      new UUID(), // hash to make key unique
+      UUID.generate(), // hash to make key unique
     ]
     return keyParts.join('_')
   }
--- a/packages/framework-provider-kubernetes/src/services/event-registry.ts
+++ b/packages/framework-provider-kubernetes/src/services/event-registry.ts
@@ -1,15 +1,13 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import { EventEnvelope, Logger } from 'framework-types/dist'
-import { v4 as uuidv4 } from 'uuid';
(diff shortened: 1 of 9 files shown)


packages/framework-provider-kubernetes/src/services/event-registry.ts
```diff
@@ -1,15 +1,13 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import { EventEnvelope, Logger } from 'framework-types/dist'
-import { v4 as uuidv4 } from 'uuid';
+import { EventEnvelope, Logger, UUID } from '@boostercloud/framework-types'
```
davidverdu, line 2: Since we are no longer using uuid dependency, we should remove it from package.json

davidverdu (approved):

packages/framework-provider-kubernetes/src/services/event-registry.ts
```diff
@@ -27,26 +25,33 @@ export class EventRegistry {
...
-      event.entityID,       // entityId
-      event.kind,           // 'event' | 'snapshot'
-      event.typeName,       // 'PostCreated' event name
-      event.createdAt,      // timespan
-      uuidv4()              // hash to make key unique
+      event.entityID, // entityId
+      event.kind, // 'event' | 'snapshot'
+      event.typeName, // 'PostCreated' event name
+      event.createdAt, // timespan
+      new UUID(), // hash to make key unique
```
davidverdu, line 54: we should use UUID.generate(), since this instruction is not returning an uuid string

packages/framework-provider-kubernetes/src/services/redis-adapter.ts
```diff
@@ -1,51 +1,50 @@
 import { Logger } from 'framework-types/dist'
 import fetch from 'node-fetch'
-const redis = require("redis");
+const redis = require('redis')
```
NickSeagull, line 3:
```suggestion
// TODO: Convert redis require to import
const redis = require('redis')
```

NickSeagull (approved):

PR #713, opened 2021-03-19
Set the log level *after* the project configuration has been loaded. …

alvaroloes:
Right now the log level was just being ignored


alvaroloes:
wrong PR

PR #721, opened 2021-03-22
fixed query multiple items in readmodels

rdiaz82:
fixed read model queries for multiple elements

### DIFF
--- a/packages/framework-provider-kubernetes/src/library/events-adapter.ts
+++ b/packages/framework-provider-kubernetes/src/library/events-adapter.ts
@@ -29,7 +29,7 @@ export const store = async (
 const isNewerThan = (isoString: string) => (key: string) => {
   const keyIsoString = key.split('_')[5]
   if (!keyIsoString) throw new Error(`No ISO string in key ${key}`)
-  return new Date(keyIsoString) >= new Date(isoString)
+  return new Date(keyIsoString) > new Date(isoString)
 }
 
 export const forEntitySince = async (
--- a/packages/framework-provider-kubernetes/src/services/read-model-registry.ts
+++ b/packages/framework-provider-kubernetes/src/services/read-model-registry.ts
@@ -23,27 +23,44 @@ export class ReadModelRegistry {
     filters: Filters
   ): Promise<Array<ReadModelInterface>> {
     const l = (msg: string): void => logger.debug('readModelRegistry#search: ' + msg)
-    const readModelId = filters.id.eq
-    l(`Got id ${readModelId ?? 'UNDEFINED'}`)
-    if (!readModelId) throw new Error('Only searching by ID is supported')
-    const url = `${this.url}/v1.0/state/statestore/${this.readModelKey(readModelName, readModelId)}`
-    l(`Performing a fetch to ${url}`)
-    const response = await fetch(url)
-    if (!response.ok) {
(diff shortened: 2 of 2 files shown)


packages/framework-provider-kubernetes/src/services/read-model-registry.ts
```diff
@@ -23,27 +23,44 @@ export class ReadModelRegistry {
...
+        return []
       }
-      return [body.value]
-    } catch (err) {
-      l(`Error ${JSON.stringify(err)}`)
-      return []
+    } else {
+      const keys = await this.redis.keys(`rm_${readModelName}_*`, logger)
+      l(`Obtainer following keys for query: ${keys}`)
+      const results: ReadModelInterface[] = []
```
rdiaz82, line 53: I'm not really happy with this implementation. 

I was having issues while dealing with types and the map function because the `value` inside the map can be undefined and that forces the return type to be `ReadModelEvenloper | undefined []` and this type is different from the return type 🤷🏼‍♂️ . 

Using the `results` array, I avoid the type being `undefined` but I think that there is a more elegant way to solve this.

packages/framework-provider-kubernetes/src/services/read-model-registry.ts
```diff
@@ -23,27 +23,44 @@ export class ReadModelRegistry {
...
-      l(`Error on performing GET from ${url}`)
-    }
-    l(`Got result ${JSON.stringify(response)}`)
-    try {
-      const body = await response.json()
-      l(`Got JSON ${JSON.stringify(body)}`)
-      // TODO: Remove unnecessary id setting from k8s read model registry
-      if (body) {
-        body.id = readModelId
+    if (filters.id) {
```
rdiaz82, line 26: We need to discuss this because I think that right now we could remove the ID filter part from this method because if I'm not wrong the filter by ID (the singular query method in graphQL) is managed by the fetch method.

In this case, we are enabling the search by ID in the multiple-element search (plural queries in graphQL), and having id search in this query is quite contradictory because we have a multiple-element query that only allows filter by a unique ID, so we are going to return just one element 😛

davidverdu (approved):

packages/framework-provider-kubernetes/src/library/events-adapter.ts
```diff
@@ -29,7 +29,7 @@ export const store = async (
 const isNewerThan = (isoString: string) => (key: string) => {
   const keyIsoString = key.split('_')[5]
   if (!keyIsoString) throw new Error(`No ISO string in key ${key}`)
-  return new Date(keyIsoString) >= new Date(isoString)
+  return new Date(keyIsoString) > new Date(isoString)
```
davidverdu, line 32: Here the operator is isNewerThan so equal dates should not be included. Actually there was a bug when you reduced 5 events one of them where counted twice, so having a balance of 1000 and adding 5 events of increments of 10 were resulting in 1060 in place of 1050

packages/framework-provider-kubernetes/src/services/read-model-registry.ts
NickSeagull: Agree, but we can improve implementations once we have something working

NickSeagull (approved):

PR #723, opened 2021-03-22
Documentation: Authentication and Authorization improvements

AdrianLorenzoDev:
## Description
This PR integrates **improvements on the `Authentication and Authorization` section,** clarifying some of the aspects of the feature, reorganising the text to be easier to understand and fixing some mistakes.

## Changes
- Restructured `Authentication and authorization` section.
- Improved `Custom authentication` section, explaining a little bit better some of the aspects of the feature and fixing some misleading explanations or orthographic mistakes.

## Checks
- [X] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,2807 @@
+# Booster Documentation
+
+<!-- toc -->
+
+- [Introduction](#introduction)
+  - [What is Booster?](#what-is-booster)
+  - [Booster Principles](#booster-principles)
+  - [Why use Booster](#why-use-booster)
+- [Getting started](#getting-started)
+  - [Installing Booster](#installing-booster)
+    - [Prerequisites](#prerequisites)
+    - [Installing the Booster CLI](#installing-the-booster-cli)
+  - [Your first Booster app in 10 minutes](#your-first-booster-app-in-10-minutes)
+    - [1. Create the project](#1-create-the-project)
+    - [2. First command](#2-first-command)
+    - [3. First event](#3-first-event)
+    - [4. First entity](#4-first-entity)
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1622,7 +1623,7 @@ export class GetProductsCount {
 
 ### Authentication and Authorization
 
-You can use the [Authentication Rocket](https://github.com/boostercloud/rocket-auth-aws-infrastructure) for adding the authentication and authorization to you application. But first, you need to know that authorization in Booster is done through roles. Every Command and ReadModel has an `authorize` policy that tells Booster who can execute or access it. It consists of one of the following two values:
+First of all, you need to know that the authorization in Booster is done through roles. Every Command and ReadModel has an `authorize` policy that tells Booster who can execute or access it. It consists of one of the following two values:
```
adriangmweb, line 1626: I would add here that to use auth, the need to install the rocket, So it's clear since the beginning. Maybe something like _" To enable authorization in Booster, you have to install the authorization rocket which is compatible with JWT thrid party auth providers. In this section, we will cover how to install and use this rocket"_

adriangmweb, line 1626: I would add here that to use auth, the need to install the rocket, So it's clear since the beginning. Maybe something like " To enable authorization in Booster, you have to install the authorization rocket which is compatible with JWT thrid party auth providers. In this section, we will cover how to install and use this rocket"

adriangmweb (approved):
LGTM! Just left a small comment :)

docs/README.md
davidverdu: I agree with this comment

davidverdu (approved):

PR #725, opened 2021-03-22
Documentation: Small improvements in "Booster architecture" section

AdrianLorenzoDev:
## Description
This PR includes **some small improvements on the  "Booster architecture" section.**

## Changes
- Clarified some subsections.
- Rearranged some file lines to improve writer readability.
- Fixed some minor issues in the document.

## Checks
- [X] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1026,7 +1026,7 @@ export class MoveStock {
 }
 ```
 
-In this case, the client who submitted the command can still complete the operation. Then an event handler will take care of that `ErrorEvent` and proceed accordingly.
+In this case, the client who submitted the command can still complete the operation. Then, an event handler will take care of that `ErrorEvent` and proceed accordingly.
 
 ##### Reading entities
 
@@ -1124,7 +1124,7 @@ export class CheckCartCount {
 }
 ```
 
-Notice that you can pass as parameters `minute`, `hour`, `day`, `month`, `weekDay` and `year` to set up a cron expression. By default, if no paramateres are passed, the scheduled command will not be triggered.
+Notice that you can pass as parameters `minute`, `hour`, `day`, `month`, `weekDay` and `year` to set up a cron expression. By default, if no paramaters are passed, the scheduled command will not be triggered.
 
 #### Creating a scheduled command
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1026,7 +1026,7 @@ export class MoveStock {
 }
 ```
 
-In this case, the client who submitted the command can still complete the operation. Then an event handler will take care of that `ErrorEvent` and proceed accordingly.
+In this case, the client who submitted the command can still complete the operation. Then, an event handler will take care of that `ErrorEvent` and proceed accordingly.
```
rdoria1, line 1029: `In this case, the command operation can still be completed. An event handler will take care of that `ErrorEvent` and proceed accordingly.`

docs/README.md
```diff
@@ -1149,7 +1149,7 @@ export class EventName {
 }
 ```
 
-Events and [entities](#4-entities-and-reducers) are intimately related. Each event belongs to one entity through the `entityID` method, and entities represent the application's state after reducing the stream of events. Indeed, an entity is just an aggregated representation of the same data present in its events, so it is possible to rebuild entities from events at any time. Booster guarantees that all the events associated with an entity will be reduced in the same order they were stored. Take a look at this event:
+Events and [entities](#4-entities-and-reducers) are intimately related. Each event belongs to one entity identified through the `entityID` method, and entities represent the application's state after reducing the stream of events. Indeed, an entity is just an aggregated representation of the same data present in its events, so it is possible to rebuild entities from events at any time. Booster guarantees that all the events associated with an entity will be reduced in the same order they were stored. Take a look at this event:
```
rdoria1, line 1152: `are intimately related` -> `are closely related` 
`Each event belongs to one entity identified through the entityID method` -> `Each event belongs to one entity uniquely identified through the entityID method` this is true right? it is unique and can be identified with tis entityID

docs/README.md
```diff
@@ -1253,13 +1253,9 @@ export class HandleAvailability {
...
 ```
 
-#### Events ordering
-
-<!-- TODO: several people have asked about how Booster ensures event ordering. I think it makes sense to explain that here  -->
-
 ### 3. Event handlers
 
-In event-driven architectures we have different parts of our application that react to events. In the case of Booster, we have the entities (in charge of reducing the events), and the _event handlers_. These are classes decorated with the `@EventHandler` decorator whose goal is to execute business logic or trigger other events when a specific event occurs.
+In event-driven architectures we have different parts of our application that react to events. In the case of Booster, we have the entities (in charge of reducing the events), and the _event handlers_. These are classes decorated with the `@EventHandler` decorator which can be defined as reactive side effect of the registration of a certain event, that can be used to execute some business logic or register other events related to the handler.
```
rdoria1, line 1258: As expected with event-driven architectures, multiple parts of our application react to events. In the case of Booster, we have entities (in charge of reducing the events) and _event handlers_. These last ones are classes decorated with the `@EventHandler` decorator. Every time a new instance of a given event is registered, the `handle` method of this class is triggered. This method can contain any business logic defined by the user or it can also register new events.

I think this explains the concept better, what do you think?

docs/README.md
AdrianLorenzoDev: That is right @rdoria1. It is uniquely identified through its entityID.

Great recommendations, I will work on it. Thank you :D

docs/README.md
AdrianLorenzoDev: I like it. It is simpler and cleaner. Thank you :)

rdoria1 (approved):
LGTM

Moneyba (approved):

PR #739, opened 2021-03-23
doc: Update rockets section

Moneyba:
## Description
This PR tries to improve the Rockets section

## Checks
- [ ] Project Builds

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2578,7 +2578,7 @@ const YourRocketInitializator = (params: YourRocketParams): InfrastructureRocket
 export default YourRocketInitializator
 ```
 
-In `mountStack` you will receive an initialized AWS CDK stack that you can use to add new resources. Check out [the Stack API in the official CDK documentation](https://docs.aws.amazon.com/cdk/latest/guide/stacks.html#stack_api). This is the same stack instance that Booster uses to deploy its resources, so your resources will automatically be deployed along with the Booster's on the same stack. You will also receive the `config` object which includes properties of the Booster project that is about to be deployed.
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -2580,11 +2578,11 @@ const YourRocketInitializator = (params: YourRocketParams): InfrastructureRocket
 export default YourRocketInitializator
 ```
 
-In `mountStack` you will receive an initialized AWS CDK stack that you can use to append new resources. Check out [the Stack API in the official CDK documentation](https://docs.aws.amazon.com/cdk/latest/guide/stacks.html#stack_api). This is the same stack instance that Booster uses to deploy its resources, so your resources will be deployed automatically along with the Booster's ones in the same stack. Also you will receive the `config` object which includes properties of the Booster project that is about to be deployed.
+In `mountStack` you will receive an initialized AWS CDK stack that you can use to add new resources. Check out [the Stack API in the official CDK documentation](https://docs.aws.amazon.com/cdk/latest/guide/stacks.html#stack_api). This is the same stack instance that Booster uses to deploy its resources, so your resources will automatically be deployed along with the Booster's on the same stack. You will also receive the `config` object which includes properties of the Booster project that is about to be deployed.
```
AdrianLorenzoDev, line 2581:
```suggestion
In `mountStack` you will receive an initialized AWS CDK stack that you can use to add new resources. Check out [the Stack API in the official CDK documentation](https://docs.aws.amazon.com/cdk/latest/guide/stacks.html#stack_api). This is the same stack instance that Booster uses to deploy its resources, so your resources will automatically be deployed along with the Booster's ones on the same stack. You will also receive the `config` object which includes properties of the Booster project that is about to be deployed.
```

I think that `ones` is needed. What do you think?

AdrianLorenzoDev (approved):
LGTM :) Just one comment.

rdoria1 (approved):

PR #741, opened 2021-03-23
add searcher-adapter for redis db

adriangmweb:
## Description

Adds a ReadModel searcher-adapter which enables filtering queries on Redis.

## Changes

Added the searcher adapter file instead of using a function.

## Checks
- [x] Project Builds
 
## Additional information

Complex filters are not yet supported.

### DIFF
--- a/packages/framework-provider-kubernetes/src/library/searcher-adapter.ts
+++ b/packages/framework-provider-kubernetes/src/library/searcher-adapter.ts
@@ -1,0 +1,79 @@
+import { RedisAdapter } from './../services/redis-adapter'
+import {
+  BoosterConfig,
+  Logger,
+  InvalidParameterError,
+  FilterFor,
+  Operation,
+  ReadModelEnvelope,
+} from '@boostercloud/framework-types'
+
+export async function searchReadModel(
+  redis: RedisAdapter,
+  config: BoosterConfig,
+  logger: Logger,
+  readModelName: string,
+  filters: FilterFor<unknown>
+): Promise<Array<any>> {
+  logger.debug('Running search with the following filters: \n', filters)
+
+  const keys = await redis.keys(`rm_${readModelName}_*`, logger)
+  logger.debug(`Obtainer following keys for query: ${keys}`)
+  const results: Array<unknown> = []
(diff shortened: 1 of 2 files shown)


davidverdu (changes requested):
After the change of Redis key separator the keys query you perform with _ separator is not going to work properly. I'll left a comment with the FIX

packages/framework-provider-kubernetes/src/library/searcher-adapter.ts
```diff
@@ -0,0 +1,79 @@
...
+export async function searchReadModel(
+  redis: RedisAdapter,
+  config: BoosterConfig,
+  logger: Logger,
+  readModelName: string,
+  filters: FilterFor<unknown>
+): Promise<Array<any>> {
+  logger.debug('Running search with the following filters: \n', filters)
+
+  const keys = await redis.keys(`rm_${readModelName}_*`, logger)
```
davidverdu, line 20: rm_${readModelName}_* is not going to retrieve any results, since the separator now is _____ . I'll encourage you to use this syntax to generate the keys search term
`redis.keys(['rm', readModelName, '*'].join(RedisAdapter.keySeparator)`

packages/framework-provider-kubernetes/src/library/searcher-adapter.ts
```diff
@@ -0,0 +1,79 @@
...
+  logger.debug(`Obtainer following keys for query: ${keys}`)
+  const results: Array<unknown> = []
+  await Promise.all(
+    keys.map(async (k: any) => {
```
NickSeagull, line 24: `any` police 👮🏻‍♂️
```suggestion
    keys.map(async (k: string) => {
```

NickSeagull (approved):
Small suggestion

packages/framework-provider-kubernetes/src/library/searcher-adapter.ts
adriangmweb: Done, and works like a charm 👌🏻

davidverdu (approved):
LGTM

PR #744, opened 2021-03-23
added configuration for deploy in multiple clusters

rdiaz82:
The configuration for deploy in multiple clusters is the following:

 
Screenshot 2021-03-23 at 23 53 04

Before moving from one cluster to another we need to delete de component folder, to avoid the bug that we have with non-existing eventstore

### DIFF
--- a/packages/framework-provider-kubernetes-infrastructure/src/infrastructure/index.ts
+++ b/packages/framework-provider-kubernetes-infrastructure/src/infrastructure/index.ts
@@ -4,6 +4,9 @@ import { HelmManager } from './helm-manager'
 import { DeployManager } from './deploy-manager'
 import { DaprManager } from './dapr-manager'
 import { Promises } from '../helpers/promises'
+export interface BoosterK8sConfiguration extends BoosterConfig {
+  context?: string
+}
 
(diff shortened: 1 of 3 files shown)


packages/framework-provider-kubernetes-infrastructure/src/infrastructure/index.ts
```diff
@@ -4,18 +4,25 @@ import { HelmManager } from './helm-manager'
 import { DeployManager } from './deploy-manager'
 import { DaprManager } from './dapr-manager'
 import { Promises } from '../helpers/promises'
+export interface BoosterK8sConfiguration extends BoosterConfig {
```
rdiaz82, line 7: Any suggestion here to add a new property to BoosterConfig without the need of extends and create a new interface. I don't really like this approach.

NickSeagull: That, in fact, is the way to adding a new property to BoosterConfig 😂

NickSeagull (approved):
Lovely, thanks!!

davidverdu (approved):

PR #745, opened 2021-03-24
Implement optimistic concurrency for read models

alvaroloes:
## Description
The way to control and apply **concurrency** in a Booster application is through **Entities**: 
- Events with a different combination of {Entity type, entity ID} are eligible to be processed concurrently. 
- Events with the same  {Entity type, entity ID} will be processed in series

When we say _process_ we mean (and in this order):
1. Calculate the new entity snapshot by applying the reductions
2. Calculate the new read models that project that entity and store its new value
3. Execute event handlers

So steps 1, 2, and 3, are all done in the same processing unit (aka the same spawned lambda) _sequentially_

It could be the case that we have **a single read model instance** that projects **multiple instances of the same entity**. As you now know, those multiple entity instances will be updated (when an event comes) concurrently, because they have different IDs.
This means that the _same instance_ of the read model _will be updated concurrently_, possibly causing data corruption if no measures are applied to deal with the concurrency.

There are two possible solutions for this:
a) Introduce some kind of locking mechanism so that only one processing unit can be written the read model at the same time
b) Use [optimistic concurrency](https://en.wikipedia.org/wiki/Optimistic_concurrency_control)

Option a) is used in common relational databases. They allow you to lock a row and unlock it when you finish writing. Any attempt from other processing units to read/write that row will be blocked until the lock is released.
For highly distributed infrastructures like Booster uses, there is no reliable way to apply this method. Those infrastructures normally provide facilities to use optimistic concurrency

This PR solves the problem using option b) optimistic concurrency.
The main logic is in the framework-core package, meaning that only a small change on each provider is needed. 
Right now, implementing that change on each provider is optional, meaning that the provider can choose whether support optimistic concurrency or not. This was made this way to be backwards compatible but in the next follow-ups this feature will be required for each provider.

In this PR it is only implemented for AWS

## Checks
- [x] Project Builds
 
## Additional information
Here is a screenshot of the logs of a load test with very high concurrency:
image
Those places where the number of retires is higher than 1 would have produced data corruptions in read models if it weren't for optimistic concurrency


packages/framework-core/src/helpers/retrier.ts
```diff
@@ -0,0 +1,32 @@
+import { Class, Logger } from '@boostercloud/framework-types'
+
+export async function retryIfError<TReturn>(
```
alvaroloes, line 3: This function is the core of the optimistic concurrency solution. It retries a logic if a specific error is found

packages/framework-core/src/services/read-model-store.ts
```diff
@@ -32,48 +35,26 @@ export class ReadModelStore {
...
-          )
-          return
-        }
-        this.logger.debug(
-          `[ReadModelStore#project] Storing new version of read model ${readModelName} with ID ${readModelID}:`,
-          newReadModel
+        return await retryIfError(
+          this.logger,
+          () => this.applyProjectionToReadModel(readModelName, readModelID, projectionMetadata, entitySnapshot),
+          OptimisticConcurrencyUnexpectedVersionError
```
alvaroloes, line 52: The above logic (read the read model, update, and store it) was moved to a method. Now we use the `retryIfError` function here. The logic will be retried if the provider throws `OptimisticConcurrencyUnexpectedVersionError`

packages/framework-core/src/services/read-model-store.ts
```diff
@@ -84,6 +65,55 @@ export class ReadModelStore {
...
   }
 
+  private async applyProjectionToReadModel(
+    readModelName: string,
+    readModelID: UUID,
+    projectionMetadata: ProjectionMetadata,
+    entity: EntityInterface
+  ): Promise<unknown> {
+    const readModel = await this.fetchReadModel(readModelName, readModelID)
+    const currentReadModelVersion: number = readModel?.boosterMetadata?.version ?? 0
```
alvaroloes, line 75: We store the `version` of the read model (which matches with the number of times it has been updated)
in the special field `boosterMetadata`

packages/framework-core/src/services/read-model-store.ts
```diff
@@ -84,6 +65,55 @@ export class ReadModelStore {
...
+      `[ReadModelStore#project] Storing new version of read model ${readModelName} with ID ${readModelID}:`,
+      newReadModel
+    )
+
+    return this.provider.readModels.store(
+      this.config,
+      this.logger,
+      readModelName,
+      newReadModel,
+      currentReadModelVersion
```
alvaroloes, line 106: Here we tell the provider that if the read model instance we want to update _is not currently_ in the version `currentReadModelVersion`, then throw the `OptimisticConcurrencyUnexpectedVersionError`

packages/framework-provider-aws/src/library/read-models-adapter.ts
```diff
@@ -38,14 +45,27 @@ export async function storeReadModel(
...
+          ':version': expectedCurrentVersion,
+        },
+      })
+      .promise()
+  } catch (e) {
+    // The error will be thrown, but in case of a conditional check, we throw the expected error type by the core
+    if (e.name == 'ConditionalCheckFailedException') {
+      throw new OptimisticConcurrencyUnexpectedVersionError(e.message)
+    }
+    throw e
```
alvaroloes, line 67: The change needed in the provider is:

Throw the error `OptimisticConcurrencyUnexpectedVersionError` if the attribute `boosterMetadata.version` is not equal to `expectedCurrentVersion`

packages/framework-core/src/helpers/retrier.ts
```diff
@@ -0,0 +1,32 @@
...
+      logger.debug(`[retrier] Logic succeeded after ${tryNumber} retries`)
+      return result
+    } catch (e) {
+      checkRetryError(e, errorClassThatRetries, logger)
+      errorAfterMaxTries = e
+    }
+  }
+  throw new Error(
+    `[retrier] Reached the maximum number of retries (${maxRetries}), but still getting the following error: ${errorAfterMaxTries}`
+  )
```
javiertoledo, line 24: In future releases, I think that it could make sense to put the failed job in a dead letter queue instead of just failing so devs can inspect and maybe fix some of the problems

packages/framework-provider-aws/src/library/read-models-adapter.ts
```diff
@@ -38,14 +45,27 @@ export async function storeReadModel(
...
-      TableName: config.resourceNames.forReadModel(readModelName),
-      Item: readModel,
-    })
-    .promise()
+  try {
+    await db
+      .put({
+        TableName: config.resourceNames.forReadModel(readModelName),
+        Item: readModel,
+        ConditionExpression: 'attribute_not_exists(boosterMetadata.version) OR boosterMetadata.version = :version',
```
javiertoledo, line 56: Interesting. Is this check transactional? I mean, does it exist any possibility that after the check happens and before the write has performed, another thread write succeeds? If it's transactional, this solution is 💯

javiertoledo (approved):
If you confirm that the check and the write happen transactionally, this change is 🔥

davidverdu (approved):
Good Job!!!
- I tested it manually with the Kubernetes-Dapr provider and Asker demo and it works (We haven't implemented the Optimistic Concurrency in Kubernetes-Dapr provider yet).
- Before merging, I would wait until somebody test this with the Azure provider
- Please also consider Javi's comments

NickSeagull (approved):
Awesome!!!!

packages/framework-core/src/helpers/retrier.ts
charlietfe: This change is important and is part of the error handling we had in mind. Nice!

charlietfe (approved):
🥇 LGTM! Good change @alvaroloes

adriangmweb (approved):
Really impressed, it's a great feature!

packages/framework-core/src/helpers/retrier.ts
alvaroloes: I agree. As part of the error handler task, we should have an entity (or anything that acts as a stream of events) to store events that represent the errors we get.

This way developers can not only inspect those problems, but they can even subscribe to them and get then in real time

packages/framework-provider-aws/src/library/read-models-adapter.ts
alvaroloes: It is guaranteed by DynamoDB to occur atomically 😃 

This is the requirement each provider must fulfil: a way to perform a conditional write atomically

packages/framework-core/src/helpers/retrier.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+function checkRetryError(e: Error, errorClassThatRetries: Class<Error>, logger: Logger): void {
+  if (!(e instanceof errorClassThatRetries)) {
+    logger.debug('[retrier] Logic failed with an error that must not be retried. Rethrowing')
+    throw e
```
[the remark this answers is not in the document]
alvaroloes: Interesting! Maybe I can throw a new error while keeping the stack trace, I will investigate

packages/framework-core/src/helpers/retrier.ts
alvaroloes: Anyway! I'm writing integration tests to prove this (I already did manual tests and checked it)

alvaroloes: Integration tests created and proven that they fail if no optimistic concurrency is used

packages/framework-core/src/helpers/retrier.ts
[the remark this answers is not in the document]
alvaroloes: After investigating, the stack trace is tracked when the error is created, not when it is thrown. So in this case, the stack trace will be preserved 😃

PR #747, opened 2021-03-25
Feature: Remove Auth Infra From Core

charlietfe:
## Description

In this PR, I've removed all the auth related code from the core and now we don't have any dependency on Cognito. 

## Changes

- Modified the Booster Config to accept a `publicKey` property with the value of the public certificate key, instead of the `jwks.json` file used in the `jwksUri`. This will avoid hitting an endpoint during tests. All the tests now are self contained. 
- Created the function `getTokenForUser`  inside `utils.ts` to generate tokens for a user with a specific role. This function will use the private key file to sign the tokens.
- Removed all the auth tests which depend on Cognito and authentication endpoints
- Removed all Cognito helpers in `utils.ts`.
- Modified test project config to use the new `publicKey` property in the tokenVerifier configuration, which uses a local public certificate included in the `assets/certs` folder. 

```
  /* We use an automatically generated app name suffix to allow
   * running integration tests for different branches concurrently.
   */
  const appNameSuffix = process.env.BOOSTER_APP_SUFFIX ?? 'default'

  // The app suffix must be copied to the test app lambdas
  config.env['BOOSTER_APP_SUFFIX'] = appNameSuffix

  config.appName = 'my-store-' + appNameSuffix
  config.provider = AWS.Provider()
  config.assets = ['assets']
  config.tokenVerifier = {
    issuer: 'booster',
  }
})
```

## Checks
- [X] Project Builds


packages/framework-core/src/booster-token-verifier.ts
```diff
@@ -46,8 +43,17 @@ export class BoosterTokenVerifier {
         })
       }
 
+      let keys: jwt.Secret | jwt.GetPublicKeyOrSecret = getKey
```
charlietfe, line 46: By default the token verifier will rely on jkws file located in a pubic url. If that url wasn't configured using `jkwsUri` we'll check the `publicKey` instead.

packages/framework-integration-tests/assets/certs/public.key
```diff
@@ -0,0 +1,9 @@
+-----BEGIN PUBLIC KEY-----
```
charlietfe, line 1: For integration testing purpose.

packages/framework-integration-tests/integration/providers/aws/utils.ts
```diff
@@ -554,3 +366,70 @@ export async function countSnapshotItems(entityTypeName: string, entityID: strin
...
+
+    await sleep(trialDelayMs)
+    return doWaitFor()
+  }
+}
+
+// This helper will create a valid token using a real private key for testing
+// the tokens will be validate against the public keyset file
+// located in: /keys/private.key file
+export const getTokenForUser = (email: string, role: string): string => {
```
charlietfe, line 416: This method is the key to perform the integration tests using valid tokens. It will create token, signed with a private key, containing the user envelop including the role.

packages/framework-integration-tests/integration/providers/aws/keys/private.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
charlietfe, line 1: For integration testing purpose.

packages/framework-core/src/booster-token-verifier.ts
```diff
@@ -10,25 +10,22 @@ export class BoosterTokenVerifier {
...
-      this.client = jwksRSA({
-        jwksUri,
-        cache: true,
-        cacheMaxAge: 900000, // 15 Minutes, at least to be equal to AWS max lambda limit runtime
-      })
+      if (this.config.tokenVerifier.jwksUri) {
+        this.client = jwksRSA({
+          jwksUri: this.config.tokenVerifier.jwksUri,
+          cache: true,
+          cacheMaxAge: 900000, // 15 Minutes, at least to be equal to AWS max lambda limit runtime
```
NickSeagull, line 17: Nitpicking here, why not:

```suggestion
          cacheMaxAge: 15 * 60 * 1000, // 15 Minutes, at least to be equal to AWS max lambda limit runtime
```

packages/framework-core/src/booster-token-verifier.ts
```diff
@@ -46,8 +43,17 @@ export class BoosterTokenVerifier {
         })
       }
 
+      let keys: jwt.Secret | jwt.GetPublicKeyOrSecret = getKey
+      if (!this.client) {
+        if (this.config.tokenVerifier?.publicKey) {
+          keys = this.config.tokenVerifier.publicKey
+        } else {
+          throw new Error('Token verifier not well configured')
+        }
```
NickSeagull, line 52: Maybe:
```suggestion
        if (!this.config.tokenVerifier?.publicKey) {
          throw new Error('Token verifier not well configured')
        }
        keys = this.config.tokenVerifier.publicKey
```

packages/framework-core/src/booster-token-verifier.ts
```diff
@@ -46,8 +43,17 @@ export class BoosterTokenVerifier {
...
+      }
+
       token = this.sanitizeToken(token)
-      jwt.verify(token, getKey, this.options, (err: Error | null, decoded: object | undefined) => {
+      jwt.verify(token, keys, this.options, (err: Error | null, decoded: object | undefined) => {
```
NickSeagull, line 56:
```suggestion
      jwt.verify(token, keys, this.options, (err: Error | null, decoded?: object) => {
```

NickSeagull (approved):
Some small suggestions, LGTM

PR #765, opened 2021-03-31
Fix snapshots strategy reduce directly from stream

alvaroloes:
## Description
Changes the way we process the events and generates snapshots to avoid failures (leave unprocessed events) on high loads.

Before, when DynamoDB Streams told us there was a change in the event store, we ignore those changes and just follow this process:
1. Get the latest snapshot
2. Read all events from that snapshots onwards (we relied on the createdAt timestamp for the ordering)
3. Reduce those events
4. Store a new snapshot if there were more than 5 new events.

This process had several problems on high loads:
a) Timestamps are good to sort if you ensure that no more than one event will be generated in the same millisecond. This is not true on highloads
b) The reading of events and storing a snapshot that included those while more events are being stored concurrently was causing some race condition that caused the generation of snapshots that didn't include some events

The applied solution consists of relying on the events that DynamoDB Streams send to us (or any provider), instead of re-read them from the DB. So the process now is the following:
1. Get the latest snapshot
2. Reduce the batch of events that we have received
3. Store a new snapshot.

As you see, we are now always storing a new snapshot after processing a batch of event. No mater if that batch had 1 or 1000 events.

## Extra improvements
- [AWS] Applied optimistic concurrency when storing events so that they are always ordered. If two events are stored in the same millisecond, one of them will fail and we will retry the storing with a newer timestamp.
- Create a new `framework-common-helpers` package to have helpers used across several packages (we were starting to duplicate too many helpers)
- Removed the generation of snapshots in the method `fetchEntitySnapshot`, as it can be called from any handler, what would cause reductions and snapshots generations concurrently for the same entityID (which will break things)

## Checks
- [x] Project Builds


packages/framework-core/src/services/event-store.ts
```diff
@@ -42,20 +40,43 @@ export class EventStore {
         newEntitySnapshot
       )
 
-      if (newEntitySnapshot && pendingEvents.length >= numberOfEventsBetweenSnapshots) {
-        await this.storeSnapshot(newEntitySnapshot)
-      }
-
```
alvaroloes, line 48: This function is now safe to be used by any handler, as it is not generating snapshots anymore

packages/framework-core/src/services/event-store.ts
```diff
@@ -90,7 +111,8 @@ export class EventStore {
         entityTypeName: eventEnvelope.entityTypeName,
         typeName: eventEnvelope.entityTypeName,
         value: newEntity,
-        createdAt: eventEnvelope.createdAt,
+        createdAt: new Date().toISOString(), // TODO: This could be overridden by the provider. We should not set it. Ensure all providers set it
+        snapshottedEventCreatedAt: eventEnvelope.createdAt,
```
alvaroloes, line 115: "Pointer" to the latest event this snapshot inlcudes

packages/framework-core/src/services/raw-events-parser.ts
```diff
@@ -1,14 +1,46 @@
...
     rawEvents: any,
-    callbackFn: (eventEnvelope: EventEnvelope, config: BoosterConfig) => Promise<void>
+    callbackFn: EventsStreamingCallback
   ): Promise<void> {
-    for (const event of config.provider.events.rawToEnvelopes(rawEvents)) {
-      await callbackFn(event, config)
+    const eventEnvelopesPerEntity = config.provider.events
+      .rawToEnvelopes(rawEvents)
+      .filter(isEventKind)
+      .reduce(groupByEntity, {})
```
alvaroloes, line 22: Now we process events in batches, not one by one (this was very inefficient before). So we need to remove snapshots and divide the events by entity ID, so they are safely reduced together

PR #768, opened 2021-04-05
Add first kubernetes integration test

juanjoman:
## Description
This PR adds the first integration test for the kubernetes provider. This integration test will deploy a booster project to [minikube ](https://minikube.sigs.k8s.io/docs/) and checks that it has been well configured.

## Changes
- Make the kubernetes provider compatible with Minikube (to set github actions and run the provider in local)
- Add all folders and configs to start the integration tests: `lerna run integration/k8s-deploy --stream`
- Add the first integration test, that checks that all services are up and running

## Checks
- [X] Project Builds


juanjoman:
AWS deployment is failing because the account reached the bucket limitation on AWS. I'll remove some as soon as I can.

.github/workflows/integration-tests.yml
```diff
@@ -97,4 +97,27 @@ jobs:
           node-version: 12.x
       - run: npx lerna bootstrap
       - run: npx lerna run compile
-      - run: npx lerna run integration/local --stream
\ No newline at end of file
+      - run: npx lerna run integration/local --stream
+
+  kubernetes-deploy-integration-tests:
+    name: Kubernetes deployment integration tests
+    runs-on: ubuntu-latest
```
adriangmweb, line 104: I wonder if it wouldn't be better to fix the ubuntu version to avoid running into breaking changes on the newest versions 🤔 I guess is a trade-off; either ensuring test pass always running in the same environment or have them always up to date fixing those breaking changes. So I'll leave it up to you

adriangmweb (approved):
LGTM! Just left a minor comment 🚀

.github/workflows/integration-tests.yml
juanjoman: Good point! For now I think it's a nice "problem" to have since Ubuntu releases a version every 2 years, we can create a very small PR changing all versions for all integration tests or just make it compatible with the next one :v:.

juanjoman:
Latest changes: 

- Last week the redis image changed all "slaves" by "replicas" so I had to update a test
- Now the redis image doesn't include the `app=redis` label anymore so I have to add an if to take the name

packages/framework-integration-tests/package.json
```diff
@@ -13,25 +13,28 @@
     "@boostercloud/framework-core": "^0.15.1",
     "@boostercloud/framework-provider-aws": "^0.15.1",
     "@boostercloud/framework-provider-local": "^0.15.1",
-    "@boostercloud/framework-types": "^0.15.1"
+    "@boostercloud/framework-provider-kubernetes": "^0.15.1",
+    "@boostercloud/framework-types": "^0.15.1",
+    "aws-sdk": "2.764.0"
```
juanjoman, line 18: @boostercloud/booster-core I had to move this `aws-sdk` package to dependencies because we are adding a second provider. This is a bug that @adriangmweb found recently.

davidverdu (approved):
LGTM. It works in Moneyba's computer

PR #772, opened 2021-04-07
Doc/config azure k8s

davidverdu:
## Description
Document Azure and K8S providers

## Changes
- Updating README.md

## Checks
- [x ] Project Builds
- [x ] Project passes tests and checks
- [x ] Updated documentation accordingly


docs/README.md
```diff
@@ -327,6 +333,59 @@ the application with the selected profile.
...
+
+You can check the different login options in this [reference document](https://docs.microsoft.com/en-us/cli/azure/authenticate-azure-cli). 
+
+Once you are logged in from the CLI, you can deploy applications with Booster.
+
+> You can see more details in [Providers configuration](#providers-configuration) section.
+
+#### Kubernetes Provider Prerequisites
+
+This step is optional, in the case you want to use the Kubernetes provider.
```
rdiaz82, line 371: Why are these steps optional? if you want to run the app on one of these providers you have to do these steps so they are not optional

docs/README.md
```diff
@@ -327,6 +333,59 @@ the application with the selected profile.
...
+This step is optional, in the case you want to use the Kubernetes provider. 
+
+The main requirement is having a Kubernetes Cluster already configured. This provider has been succesfully tested in EKS (Amazon Elastic Kubernetes Service), AKS (Azure Kubernetes Service) and GKE (Google Kubernetes Engine). Since Kubernetes is a standard, this provider can also work in other Kubernetes clusters, including on-premises configurations.
+
+> **Note**:
+>
+> Booster is free to use, but notice that the Kubernetes cluster resources used for the deployment
+> might generate some expenses.
+
+On the other hand, it is also required to install **kubectl** and [Helm](https://helm.sh).
```
rdiaz82, line 380: I would rephrase the beginning `On the other hand`  because we are not contrasting with something like `Additionally, it is also ....`

Regarding Helm we need Helm version >=3 I think that we can call it Helm3 (not really sure about this name) or just mention the minimum version

docs/README.md
```diff
@@ -2544,6 +2615,32 @@ source <path-to-your-bash-file> && boost deploy -e production
 
 Now just let the magic happen, Booster will create everything for you and give you back your app ready to use URL. 🚀
 
+##### Kubernetes provider
+
+To configure Kubernetes as a provider you need to meet certain prerequisites:
+
+- Config the Kubernetes cluster beforehand in a cloud provider or on-premises.
+- Install **kubectl**.
```
rdiaz82, line 2623: We may rephrase this with something like: `install Kubectl and connect it to your Kubernetes cluster`

docs/README.md
```diff
@@ -2544,6 +2615,32 @@ source <path-to-your-bash-file> && boost deploy -e production
...
+- Install [Helm](https://helm.sh).
+- Check `@boostercloud/framework-provider-kubernetes` is listed in your app `package.json` dependencies.
+- Check `@boostercloud/framework-provider-kubernetes-infrastructure` is listed in your app `package.json` devDependencies.
+- Check both dependencies are installed, otherwise use `npm install` in the root of your project.
+
+At this moment you should be able to access your Kuberbetes cluster from **kubectl** command.
```
rdiaz82, line 2629: If we change the line where we install kubectl we can remove this line

docs/README.md
```diff
@@ -2544,6 +2615,32 @@ source <path-to-your-bash-file> && boost deploy -e production
...
+
+Now go to your `config.ts` file, import the kubernetes provider library and set up your app environment.
+
+```typescript
+import { Booster } from '@boostercloud/framework-core'
+import { BoosterConfig } from '@boostercloud/framework-types'
```
rdiaz82, line 2635: we have a custom configuration for the K8s provider so the import is this one: `import { BoosterK8sConfiguration } from '@boostercloud/framework-provider-kubernetes-infrastructure'`

docs/README.md
```diff
@@ -2544,6 +2615,32 @@ source <path-to-your-bash-file> && boost deploy -e production
...
+import { Provider } from '@boostercloud/framework-provider-kubernetes'
+
+Booster.configure('production', (config: BoosterConfig): void => {
```
rdiaz82, line 2638: Change `BoosterConfig` by `BoosterK8sConfiguration`

rdiaz82 (changes requested):
Some minor changes

docs/README.md
davidverdu: I copied the AWS part which is also optional. These are the prerequisites to deploy an application and you don't need to set up all providers to go. Would it be better to say "This step is only required if you want to work with K8s provider" ? @rdiaz82

docs/README.md
davidverdu: Makes sense. I'll add the version

docs/README.md
davidverdu: ok

docs/README.md
davidverdu: ok

docs/README.md
davidverdu: ok

docs/README.md
davidverdu: ok

rdiaz82 (approved):
LGTM

docs/README.md
```diff
@@ -2595,7 +2690,7 @@ The only thing you need to do to deploy a whole new completely-independent copy
 
 You can extend Booster by creating rockets. A rocket is just a node package that implements the public Booster rocket interfaces. You can use them for many things:
 
-1. Extend your infrastructure (Currently, only in AWS): You can write a rocket that adds provider resources to your application stack.
+1. Extend your infrastructure: You can write a rocket that adds provider resources to your application stack.
 2. Runtime extensions (Not yet implemented): Add new annotations and interfaces, which combined with infrastructure extensions, could implement new abstractions on top of highly requested use cases.
```
Moneyba, line 2694: Can we create rockets with the other providers?
In that case, we should also change [this part](https://github.com/boostercloud/booster/tree/main/docs#create-your-own-rocket).

davidverdu: No, we cannot at this moment, but the idea is allowing it in the future. Let's keep the mentioned part as it meanwhile. Thanks for the comment

Moneyba (approved):

PR #780, opened 2021-04-12
changes in k8s provider

rdiaz82:
## Description
Some minor fixes for the K8s provider.

## Changes
- Fix get the latest snapshot to get the latest snapshot instead of the first (k8s provider)
- add a method to check that a service is already available pinging the public url to check that it is returning a 200 (k8s provider)
- defaulted the logger to info instead of debug to reduce the amount of info during the deployment (k8s provider)
- added a info message when we start to create the `.deploy` folder during the deploy (cli package)

## Checks
- [x] Project Builds
- [x] Updated documentation accordingly (these changes don't affect documentation)

### DIFF
--- a/packages/framework-provider-kubernetes-infrastructure/src/helpers/logger.ts
+++ b/packages/framework-provider-kubernetes-infrastructure/src/helpers/logger.ts
@@ -1,9 +1,21 @@
-import { Logger } from '@boostercloud/framework-types'
+import { Level, Logger } from '@boostercloud/framework-types'
 
-export const scopeLogger = (scope: string, logger: Logger): Logger => {
+export const scopeLogger = (scope: string, logger: Logger, level = Level.info): Logger => {
   const newLogger: Logger = { ...logger }
-  newLogger.debug = (...args: any[]) => logger.debug([`[${scope}]`, ...args].join(' '))
-  newLogger.info = (...args: any[]) => logger.info([`[${scope}]`, ...args].join(' '))
-  newLogger.error = (...args: any[]) => logger.error([`[${scope}]`, ...args].join(' '))
+  newLogger.debug = (...args: any[]) => {
+    if (level <= Level.debug) {
+      logger.debug([`[${scope}]`, ...args].join(' '))
+    }
+  }
+  newLogger.info = (...args: any[]) => {
+    if (level <= Level.info) {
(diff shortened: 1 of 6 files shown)


packages/framework-provider-kubernetes-infrastructure/src/helpers/logger.ts
```diff
@@ -1,9 +1,21 @@
...
+import { Level, Logger } from '@boostercloud/framework-types'
 
-export const scopeLogger = (scope: string, logger: Logger): Logger => {
+export const scopeLogger = (scope: string, logger: Logger, level = Level.info): Logger => {
   const newLogger: Logger = { ...logger }
-  newLogger.debug = (...args: any[]) => logger.debug([`[${scope}]`, ...args].join(' '))
-  newLogger.info = (...args: any[]) => logger.info([`[${scope}]`, ...args].join(' '))
-  newLogger.error = (...args: any[]) => logger.error([`[${scope}]`, ...args].join(' '))
+  newLogger.debug = (...args: any[]) => {
+    if (level <= Level.debug) {
```
charlietfe, line 6: Just for curiosity, if you are alway calling the specific logger method why you need the level?

charlietfe (approved):
Minor comment for clarification. The rest LGTM

packages/framework-provider-kubernetes-infrastructure/src/helpers/logger.ts
rdiaz82: we added a lot of debug messages during the development of the K8s provider for debugging and here we are muting all those debug messages defaulting the logger to info level. 

This will be refactored after the meeting to use the application logging level

NickSeagull: This function shouldn't be doing this, we should have another function that initializes the logger this way. Let's do that: Remove these checks from here and add them to the initialization

NickSeagull (changes requested):
Just one small changes

davidverdu (approved):
LGTM. Nonetheless, I experienced a compilation problem because the dev dependency @types/node-fetch was missing in the Kubernetes provider package. This dev dependency should be added

rdiaz82:
closing this PR because the Logging stuff is going to be moved to a new ticket/PR

PR #791, opened 2021-04-19
Doc: Add seo plugin

Moneyba:
## Changes
Add SEO plugin

## Checks
- [ ] Project Builds

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,73 +1,114 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-  <meta charset="UTF-8">
-  <title>Booster cloud</title>
-  <link rel="icon" href="img/favicon.png">
-  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
-  <meta name="description" content="The fastest way to create an application for the cloud.">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0">
+    <meta charset="UTF-8">
+    <title>Booster cloud</title>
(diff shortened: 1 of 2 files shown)


docs/index.html
```diff
@@ -1,73 +1,101 @@
...
+            document.querySelector(`meta[name="${name}"]`).content = content || ''
+          },
+          function (hook, vm) {
+            const refreshInfo = () => {
+              const { config, route, frontmatter } = vm
+              const { description, keywords } = frontmatter || {}
+              const entries = { description, keywords }
+
+              for (const key in entries) {
+                var value = entries[key]
```
davidverdu, line 82: please use const or let in place of var

davidverdu (changes requested):
The page is displaying the error `Uncaught TypeError: document.querySelector(...) is null`. I also left a minor comment

docs/index.html
```diff
@@ -1,73 +1,101 @@
...
+                    '<footer style="text-align: center;margin-top: 50px;">',
+                    '<span> Booster is a spin-off project from <a href="https://www.theagilemonkeys.com">The Agile Monkeys</a>.</span>',
+                    '</footer>'
+                ].join('')
+                hook.afterEach(function (html) {
+                    return html + footer
+                })
+            },
+          function meta(name, content) {
+            document.querySelector(`meta[name="${name}"]`).content = content || ''
```
davidverdu, line 73: I believe this function is throwing the error, since the element searched with querySelector is not found. Do you know when this function is executed? Is the DOM loaded before this function is run?

davidverdu (approved):

PR #805, opened 2021-04-29
feat: PascalCase resource name generation check

Fecony:
## Description
This PR is fixing test issue where tests fail because new feature is not formatting user input properly.
Change flow to check if name is valid, and then throw error if it is not.

## Changes
- Remove `resourceNameToClassName` method
- Add 3 new methods, 1 exported `checkResourceNameIsValid`
- Throw error if resource name is not valid PascalCase to force user to use correct naming
- Fix generator tests to use valid PascalCase name for files
- Add new tests

## Checks
- [x] Project Builds


packages/cli/src/common/filenames.ts
```diff
@@ -1,5 +1,38 @@
 import * as inflected from 'inflected'
+import { logger } from '../services/logger'
+import Brand from './brand'
 
 export function classNameToFileName(name: string): string {
   return inflected.dasherize(inflected.underscore(name))
 }
+
+export async function checkResourceNameIsValid(name: string): Promise<void> {
+  logger.info(Brand.mellancholize('Checking if resource has valid name...'))
```
davidverdu, line 10: I would't display this message in the console. It only makes sense displaying on errors

davidverdu (changes requested):
I had to rebase main because there was an error launching `lerna run integration/cli` . Now the PR has some conflicts. To accept this PR you need to solve the conflicts first. I also left a minor comment. 

Before submitting the PR, please verify that `lerna run test` and `lerna run integration/cli --stream` run without errors (these scripts doesn't need having a AWS account)

packages/cli/src/common/filenames.ts
```diff
@@ -3,3 +3,32 @@ import * as inflected from 'inflected'
 export function classNameToFileName(name: string): string {
   return inflected.dasherize(inflected.underscore(name))
 }
+
+export async function checkResourceNameIsValid(name: string): Promise<void> {
```
rdiaz82, line 7: one question, Why is this function async? I don't see any async call inside the method

rdiaz82 (approved):
In general LGTM just a question to clarify!!

charlietfe (approved):
In general LGTM, nothing to add. Thanks for contributing @Fecony

packages/cli/src/common/filenames.ts
Fecony: Hey, @rdiaz82 yeap, looks like it is not required, I only used it to do fancy `.then/.catch` in test file 😸

davidverdu (approved):
LGTM

PR #808, opened 2021-05-03
fix: add husky 

Moneyba:
## Description
You can currently submit code without the correct format.

## Changes

- Add format command in  chek-all-the-things script
- Add a git hook using [Husky](https://typicode.github.io/husky/#/) and [lint-staged](https://www.npmjs.com/package/lint-staged) to format the code before commit.

## Checks
- [x] Project Builds


packages/cli/test/services/user-prompt.test.ts
```diff
@@ -4,84 +4,83 @@ import { expect } from '../expect'
...
+    expect(value).to.equal('randomValue')
+  })
 
-    it('defaultOrPrompt with null value', async () => {
-        const promptStub = stub(inquirer,'prompt')
-        promptStub.resolves({value: 'mockedValue'})
-        const value = await new Prompter().defaultOrPrompt(null,"Enter a value")
-        expect(value).to.equal('mockedValue')
-    })
+  it('defaultOrPrompt with undefined value', async () => {
```
davidverdu, line 18: Why rename null into undefined?

scripts/check-all-the-things.sh
```diff
@@ -1,9 +1,3 @@
 #!/usr/bin/env sh
 
-lerna clean --yes \
-&& lerna bootstrap \
-&& lerna run clean --stream \
-&& lerna run compile --stream \
-&& lerna run lint:fix --stream \
-&& lerna run lint:check --stream \
-&& lerna run test --stream
+lerna run format --stream
```
davidverdu, line 3: I believe that all these steps must be run. `lerna run format` replaces all these steps?

packages/framework-integration-tests/integration/fixtures/events/post-created.ts
```diff
@@ -3,11 +3,7 @@ import { UUID } from '@boostercloud/framework-types'
 
 @Event
 export class PostCreated {
-  public constructor(
-    readonly postId: UUID,
-    readonly title: string,
-    readonly body: string,
-  ) {}
+  public constructor(readonly postId: UUID, readonly title: string, readonly body: string) {}
```
davidverdu, line 6: I'm not sure if you change the fixtures, the integration test pass. Have you run integration tests after this formatting?

packages/cli/src/services/user-prompt.ts
```diff
@@ -3,10 +3,10 @@ import inquirer = require('inquirer')
 export default class Prompter {
   public async defaultOrPrompt(value: string | undefined | null, promptMessage: string): Promise<string> {
     if (value) {
-      return Promise.resolve(value.replace(/\"/g,"\\\""))
+      return Promise.resolve(value.replace(/\"/g, '\\"'))
```
davidverdu, line 6: I hope that the unit tests related to user-prompt pass

davidverdu (changes requested):
Good change! Now all code seems to be formatted. However, I found two issues:
(1) Integration test fails because integration test fixtures were changed in the formatting and now file comparison doesn't match. Can you leave fixtures as original and create a rule so that the formatted ignore the fixtures?
(2) I believe that you shouldn't replace the `check-all-the-things.sh` script that way. Maybe `Lerna run format` replaces the lint tasks but not the others like `bootstrap` or `compile`. What do you think? Another details is that you also need to modify the Windows script accordingly.

PR #809, opened 2021-05-04
Fixed related with the K8s provider

rdiaz82:
## Description
Added a verification before trying to upload any code into the cluster because in AWS the services are DNS based and the domain takes some time to propagate (around 1 min) and the deployment was failing some times because of this. 

Also added a small fix because we were obtaining the first snapshot instead of the last snapshot when we were using the read models. 

## Changes
Changes in K8s provider

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-provider-kubernetes/src/services/event-registry.ts
+++ b/packages/framework-provider-kubernetes/src/services/event-registry.ts
@@ -58,7 +58,7 @@ export class EventRegistry {
     if (result.length <= 0) {
       return null
     }
-    return result[0]
+    return result[result.length - 1]
   }
 
   private eventKey(event: EventEnvelope): string {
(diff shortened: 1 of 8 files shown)


juanjoman (approved):
LGTM!! :v:

davidverdu (approved):
LGTM. I left a minor comment

packages/framework-provider-kubernetes/src/services/event-registry.ts
```diff
@@ -58,7 +58,7 @@ export class EventRegistry {
     if (result.length <= 0) {
       return null
     }
-    return result[0]
+    return result[result.length - 1]
```
davidverdu, line 61: Did you manage to test if this is the right order of getting items?

rdiaz82: yes!! the readmodels reductions were failing because of this. I checked using logs messages that the array was ordered from old to new so the newest snapshot is the latest in the array! 🤓

PR #814, opened 2021-05-05
K8s nuke integration tests

juanjoman:
In this PR:
* Added 'Running' status checks to the kubernetes deployment integration tests
* Added kubernetes `boost nuke` integration tests
* Fixed a little bug with helm: Warnings are stored in `stderr` so it was failing on `helm uninstall dapr -n <namespace>`
* Merged both deployment and nuke integration tests in the same job, but with different steps


.github/workflows/integration-tests.yml
```diff
@@ -120,4 +120,4 @@ jobs:
       - run: minikube start --namespace=booster-my-store-kubernetes
       - run: npx lerna bootstrap
       - run: npx lerna run compile
-      - run: npx lerna run integration/k8s-deploy --stream
```
juanjoman, line 123: I mixed both k8s-deploy and k8s-nuke in the same job

packages/framework-provider-kubernetes-infrastructure/src/infrastructure/dapr-manager.ts
```diff
@@ -97,8 +97,8 @@ export class DaprManager {
    * remove dapr services from your cluster
    */
   public async deleteDaprService(): Promise<void> {
-    const { stderr } = await this.helmManager.exec(`uninstall dapr -n ${this.namespace}`)
-    if (stderr) throw new Error(stderr)
+    const { stdout, stderr } = await this.helmManager.exec(`uninstall dapr -n ${this.namespace}`)
+    if (!stdout && stderr) throw new Error(stderr)
```
juanjoman, line 101: Warnings are returned inside `stderr`, so I need to check both `stdout` and `stderr`

packages/framework-integration-tests/integration/providers/kubernetes/deployment/deployment.integration.ts
```diff
@@ -1,6 +1,7 @@
 import { expect } from '../../../helper/expect'
 import * as Kubernetes from '@kubernetes/client-node'
 import { boosterKubernetesServices, kubernetesNamespace } from '../constants'
+import { waitForIt } from '../../aws/utils'
```
juanjoman, line 4: should we move this method to a generic `utils` file?

packages/framework-provider-kubernetes-infrastructure/src/infrastructure/helm-manager.ts
```diff
@@ -20,7 +20,7 @@ export class HelmManager {
     const l = scopeLogger('exec', this.logger)
     const cmd = `${this.BASE_COMMAND} ${args}`
     l.debug('Executing command', cmd)
-    return exec(this.BASE_COMMAND, args)
+    return execFile(this.BASE_COMMAND, args.split(' '))
```
juanjoman, line 23: I did a quick research and `execFile` is more secure than `exec`. Also removed the codeql warning

packages/framework-integration-tests/integration/providers/kubernetes/deployment/deployment.integration.ts
davidverdu: I think so, since it's used in other providers than aws

davidverdu (approved):
LGTM. Some minor comments
- The integration test failed occasionally due to wait time issues. I would increase the wait time between Deploy and Nuke or the wait time after Deploy, so that the failure is avoided
- I left a comment to refactor waitForIt as a common util

packages/framework-integration-tests/integration/providers/kubernetes/deployment/deployment.integration.ts
alvaroloes: It is already in the generic `helpers` folder. It was duplicated.

.github/workflows/integration-tests.yml
alvaroloes: 🤔  Just take into account that it would be harder this way to know which test failed from the PR, as we would only have one "check" for the whole k8s tests. 

It might be ok, though, just saying

packages/framework-integration-tests/integration/providers/kubernetes/deployment/deployment.integration.ts
alvaroloes: Also, I removed the whole `utils.ts` file in PR https://github.com/boostercloud/booster/pull/817

Another point of that PR is that, when it gets merged, you can create some little implementation of an interface and run the integration test for k8s too (although I'm not sure if all of them would pass, as it doesn't have subscriptions, for example)

.github/workflows/integration-tests.yml
juanjoman: Agree! The only problem with this is that with a separate job we don't have the required services/pods deployed on that machine, so I'd need to deploy (again) + nuke @alvaroloes

packages/framework-integration-tests/integration/providers/kubernetes/deployment/deployment.integration.ts
juanjoman: awesome! I changed it to the `helpers` one 😄

packages/framework-integration-tests/integration/providers/kubernetes/deployment/deployment.integration.ts
```diff
@@ -26,6 +29,19 @@ describe('Kubernetes provider', () => {
       return item?.metadata?.name
     })
 
+    pods.body.items.map(async (pod: Kubernetes.V1Pod) => {
```
rdiaz82, line 32: I'm wondering if this kind of things can be extracted on a manager to be reused or simply use the K8sManager from the main project. I think that there are some tasks that we want to perform inside the K8s cluster during tests. What do you think?

packages/framework-integration-tests/integration/providers/kubernetes/nuke/nuke.integration.ts
```diff
@@ -0,0 +1,41 @@
...
+describe('After nuke', () => {
+  const kubernetesConfig = new Kubernetes.KubeConfig()
+  kubernetesConfig.loadFromDefault()
+  const k8sClient = kubernetesConfig.makeApiClient(Kubernetes.CoreV1Api)
+
+  describe('the pods and services', () => {
+    it('are deleted successfully', async () => {
+      const services = await waitForIt(
+        async () => {
+          return await k8sClient.listNamespacedService(kubernetesNamespace)
```
rdiaz82, line 17: The same concern here! should this live in a separated manager?

.github/workflows/integration-tests.yml
rdiaz82: What about performing some kind of "deploy step" before testing to prepare the environment before tests? In that case, you can have tests in different jobs

rdiaz82 (approved):
LGTM in general but I left some comments to think about them 🤔

packages/framework-integration-tests/integration/providers/kubernetes/deployment/deployment.integration.ts
juanjoman: I think I'm using a manager for tests since the current one is in the provider-infrastructure package and I would have to move it to `dependencies` to use it. Thanks @rdiaz82 !

juanjoman:
Hey @rdiaz82 I can't answer the last message above but I can't create a setup job: https://github.community/t/use-same-container-in-multiple-jobs/17562. It seems that every job is run on a new machine so I can't do that. The only viable way to split them would be to copy-paste the current job and adding these two steps:

- run: 'npm install -g @boostercloud/cli'
- run: 'boost deploy -e kubernetes'

But I don't like it very much because we're not using the `boost deploy` of that project inside the job machine, but the one already released.

cc @alvaroloes

packages/framework-integration-tests/integration/providers/kubernetes/nuke/setup.ts
```diff
@@ -0,0 +1,23 @@
...
+import { createDaprComponentFile } from '../util'
+
+before(async () => {
+  const sandboxPath = sandboxPathFor(sandboxProjectName)
+  const configuredAssets = ['assets', 'assetFile.txt']
+  const sandboxedProject = createSandboxProject(sandboxPath, configuredAssets)
+
+  await overrideWithBoosterLocalDependencies(sandboxPath)
+  // Only the deploy command creates the production dependencies
+  await exec('npm install --production --no-bin-links', { cwd: sandboxPath })
```
alvaroloes, line 16: This is not needed. We were getting an error if we didn't include this with npm version < 7, but with npm 7+, this is unneeded. 

You can safely remove it (check the nuke setup test in AWS, it didn't have it anymore)

packages/framework-integration-tests/integration/providers/kubernetes/nuke/setup.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  // boost deploy removes the dapr components folder, so we need to create it again.
+  await createDaprComponentFile(sandboxPath)
```
alvaroloes, line 19: Just wondering: Could this be replaced by adding that file to the `config.assets` configuration field (and, thus, in the above `configuredAssets` variable)?

cc @rdiaz82

alvaroloes:
Answering your comment @juanjoman (https://github.com/boostercloud/booster/pull/814#issuecomment-836726309)

I understand now the problem. There is a midground approach that will help us here, avoiding the duplication and also being able to know which Kubernetes tests failed (deploy, nuke, integration, etc.), although only when we enter in the details:

The approach is to have only one job for the tests, where you configure the machine, and then one different step for each test in the order in which they should run. For example, right now we have the following:

```yml
  kubernetes-deploy-integration-tests:
    name: Kubernetes deployment integration tests
    runs-on: ubuntu-latest
    env:
      KUBECONFIG: '/home/runner/.kube/config'
    steps:
      - uses: actions/checkout@v2
      - uses: manusa/actions-setup-minikube@v2.3.1
        with:
          minikube version: 'v1.18.1'
          kubernetes version: 'v1.19.7'
          driver: 'docker'
      - uses: actions/setup-node@v1
        with:
          node-version: 12.x
      - run: sudo snap install helm --classic
      - run: kubectl create ns booster-my-store-kubernetes
      - run: kubectl config set-context --current --namespace=booster-my-store-kubernetes
      - run: minikube start --namespace=booster-my-store-kubernetes
      - run: npx lerna bootstrap
      - run: npx lerna run compile
      - run: npx lerna run integration/k8s --stream
```

So instead of calling the "all tests in one" script `integration/k8s`, call each tests separately, starting with deploy and ending with nuke. For example:
```yml
kubernetes-deploy-integration-tests:
    name: Kubernetes deployment integration tests
    runs-on: ubuntu-latest
      # ... all the same ....
      - run: minikube start --namespace=booster-my-store-kubernetes
      - run: npx lerna bootstrap
      - run: npx lerna run compile
      - run: npx lerna run integration/k8s-deploy --stream
      - run: npx lerna run integration/k8s-integration --stream
      - run: npx lerna run integration/k8s-load --stream
      - run: npx lerna run integration/k8s-nuke --stream
```

This way, when you enter in the "details" from the Github check, you will see those steps with logs grouped individually, which is much easier to deal with.

What do you think?

packages/framework-integration-tests/integration/providers/kubernetes/nuke/nuke.integration.ts
```diff
@@ -0,0 +1,38 @@
...
+    it('are deleted successfully', async () => {
+      const services = await waitForIt(
+        async () => {
+          return await k8sManager.getKubernetesServices()
+        },
+        (services) => {
+          return services.items.length === 0
+        },
+        5000,
+        150000
```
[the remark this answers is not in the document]
juanjoman: Depending on the PC you're running this from, removing the services and pods could take less/more. For example, on my PC it takes around 50s or less, but @davidverdu was having problems on his computer. cc @alvaroloes

packages/framework-integration-tests/integration/providers/kubernetes/util.ts
```diff
@@ -0,0 +1,38 @@
...
+  const renderedYaml = Mustache.render(stateStore.template, templateValues)
+
+  console.log('Rendered Yaml:\n', renderedYaml)
+
+  console.log('Writing yaml file', outFile)
+  await fs.promises.writeFile(outFile, renderedYaml).catch(() => {
+    console.log("Couldn't write file, throwing")
+    throw new Error(`Unable to create the index file for your app: Tried to write ${outFile} and failed`)
+  })
+}
```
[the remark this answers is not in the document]
juanjoman: Hmm I'll try with the config.assets, because that function is in the provider-infrastructure package

packages/framework-integration-tests/integration/providers/kubernetes/nuke/setup.ts
juanjoman: Hey @alvaroloes the problem with this is that the deployment removed the `.deploy` folder, and the nuke doesn't generate a new one. Also the nuke command is not looking for a `.deploy` folder, but for the root path of the application (not sure if it's a bug or it's expected): https://github.com/boostercloud/booster/blob/main/packages/cli/src/commands/nuke.ts#L61

packages/framework-integration-tests/integration/providers/kubernetes/nuke/setup.ts
juanjoman: oooh this worked hehe

.github/workflows/integration-tests.yml
juanjoman: We already talked about this and we're splitting the tests on the same job but different steps

packages/cli/test/common/sandbox.test.ts
```diff
@@ -1,24 +1,25 @@
...
+import * as fse from 'fs-extra'
 import { createSandboxProject } from '../../src/common/sandbox'
 import { Dirent } from 'fs'
 import * as path from 'path'
 
 describe('sandbox', () => {
   let fakeRmSync: SinonSpy
   let fakeMkdirSync: SinonSpy
-  let fakeCopyFileSync: SinonSpy
+  let fakeFileSync: SinonSpy
```
alvaroloes, line 12: Naming thing: I think you wanted to write `fakeCopySync` 😉

packages/framework-integration-tests/integration/provider-specific/deploy.ts
```diff
@@ -1,5 +1,6 @@
 import * as path from 'path'
-import { runCommand } from '../../helper/run-command'
```
alvaroloes, line 2: Would you mind calling this file `cli-helper.ts` and put it inside the `integration/helper` folder?

I think it makes more sense.

alvaroloes (approved):
Amazing! It ended up very cool. Thanks for applying the suggestions.

I left two more suggestions but just to change the name of a variable and a file.

Pre approving!

PR #816, opened 2021-05-06
Fix/multiprovider config

Moneyba:
## Description
When a Booster app is configured to use different providers on different environments:

> Booster.configure('aws', (config: BoosterConfig): void => {
>   config.appName = 'adriangm-blog'
>   config.provider = AWSProvider()
> })
> 
> Booster.configure('azure', (config: BoosterConfig): void => {
>   config.appName = 'adriangm-blog'
>   config.provider = AzureProvider()
> })

AWS works but Azure fails because it’s trying to import a `dynamodb` dependency from AWS that it shouldn’t.

Error: Cannot find module 'aws-sdk/clients/dynamodb''Stack: Error: Cannot find module 'aws-sdk/clients/dynamodb'at Function.Module._resolveFilename

## Changes
To solve this bug we have used [optionalDependencies](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#optionaldependencies). An optional dependency works the same as a dependency, with the difference that if there is a dependency that is not installed,, the project doesn't break and by not breaking you can check in the code if that dependency exists or not.
Following the instructions of the official npm documentation we have added a "try-catch" to check if aws-sdk exists. If it doesn't exist, we return an empty ProviderLibrary type object to avoid conflicts when other providers ask for the aws-sdk dependency.

## Checks
- [ ] Project Builds
 
## Additional information
https://agilemonkeys.atlassian.net/wiki/spaces/B/pages/2292678661/Booster+Providers+setup?focusedCommentId=2293268481#comment-2293268481


packages/framework-provider-aws/src/index.ts
```diff
@@ -57,63 +9,66 @@ function loadInfrastructurePackage(packageName: string): HasInfrastructure {
...
+  } catch (e) {
+    return {
+      // ProviderEventsLibrary
+      events: {
+        rawToEnvelopes: undefined as any,
+        forEntitySince: undefined as any,
+        latestEntitySnapshot: undefined as any,
+        store: undefined as any,
+        search: undefined as any,
+      },
```
adriangmweb, line 25: Sweet! I'm still wondering if it would be worth moving this `empty provider` to a generic package 🤔  

It doesn't have to be in this PR, but it would be worth having access to it from any package.

docs/custom.css
```diff
@@ -1,6 +1,6 @@
 :root {
     --link-color: var(--theme-color);
-    --code-inline-color: var(--codeTextColor);
+    --code-inline-color: #d22778;;
```
adriangmweb, line 3: why do you stop using CSS vars here?

packages/framework-provider-aws/src/index.ts
```diff
@@ -57,63 +9,66 @@ function loadInfrastructurePackage(packageName: string): HasInfrastructure {
...
-            `  - Or it has been installed globally. You can do so by running 'npm install -g ${infrastructurePackageName}'`
-        )
-      }
-
-      return infrastructure.Infrastructure(rockets)
-    },
+  try {
+    require('aws-sdk')
+    const { Provider } = require('./setup.ts')
+    return Provider(rockets)
```
adriangmweb, line 15: For anybody wondering if this is the way to go, it is done following the [npm documentation](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#optionaldependencies)

packages/framework-provider-aws/src/setup.ts
```diff
@@ -0,0 +1,122 @@
...
+  Infrastructure: (rockets?: RocketDescriptor[]) => ProviderInfrastructure
+}
+
+/* We load the infrastructure package dynamically here to avoid including it in the
+ * dependences that are deployed in the lambda functions. The infrastructure
+ * package is only used during the deploy.
+ * Notice that this is done in a separate function to ease testing
+ */
+export function loadInfrastructurePackage(packageName: string): HasInfrastructure {
+  console.log('loadInfr')
```
adriangmweb, line 51: You might want to remove this log

packages/framework-provider-aws/src/index.ts
adriangmweb: It's done to avoid issues when other providers try to load the `aws-sdk` dependency when the `AWS` provider is also set in the config file.

packages/framework-provider-aws/test/index.test.ts
```diff
@@ -47,7 +56,7 @@ describe('the `framework-provider-aws` package', () => {
         },
       ]
 
-      const providerLibrary: ProviderLibrary = providerPackage.Provider(rockets)
+      const providerLibrary: ProviderLibrary = providerPackageSetup.Provider(rockets)
```
adriangmweb, line 59: Aren't we missing some more tests?

adriangmweb (changes requested):
Looks lovely already! I added some comments to add clarifications and requested some changes before it's ready to go

docs/custom.css
Moneyba: That variable no longer exists.

packages/framework-provider-aws/src/index.ts
```diff
@@ -57,63 +9,56 @@ function loadInfrastructurePackage(packageName: string): HasInfrastructure {
...
+    require('aws-sdk')
+    const { Provider } = require('./setup')
+    return Provider(rockets)
+  } catch (e) {
+    return {
+      // ProviderEventsLibrary
+      events: {
+        rawToEnvelopes: undefined as any,
+        forEntitySince: undefined as any,
+        latestEntitySnapshot: undefined as any,
```
juanjoman, line 22: Not sure about this, but isn't it better to throw an error message?

juanjoman (changes requested):
Just have a mini question 🙏

juanjoman (approved):
LGTM then! :v:

adriangmweb (approved):
LGTM! Thanks for your hard work, now multi-provider deployments are out of bugs! 🚀

packages/framework-provider-aws/src/index.ts
adriangmweb: Hey @alvaroloes! This was not the ideal technical solution, but it was the one that made sense from a business perspective. I'll send you some information about the issue so you can understand it better.

PR #817, opened 2021-05-06
Create package to do integration tests and make them provider-unaware

alvaroloes:
## Description
In the following video, I explain the motivations and other things that will help a lot to understand what's done in this PR:

https://www.loom.com/share/2970c7187c8042eabc32f74f5b6cc1b0

Here is the summary:

There are two main features introduced in this PR:

### 1.- New "application-tester" package
The contents of this package are a refactor of the code we had in the infamous `utils.ts` file inside the integration-tests package. It has been refactored in a way that it is completely provider-unaware. All the things that depend on the provider have been abstracted to the interface `ProviderTestHelper`. 

This package allows any Booster application to create its own integration tests much easier (no need to create all the authorization tokens and GraphQL Apollo clients code) and in a **provider-unaware way**. 
This means that no matter if the user chose AWS provider in one environment and Kubernetes or Azure in another, _the integration tests they wrote do not need to change_.

This opens the door to even autogenerate fully working integration tests!

The only thing needed in the provider side to be able to be used with this package is to implement the `ProviderTestHelper` interface.

### 2.- Booster Framework integration tests are not fully provider-unaware
As a direct result of point 1 above, I did the needed refactors in the framework integration tests to use the new package, meaning that they are all now independent from the provider we deploy to.

This is ideal to ensure the same quality in all the providers, as they will all need to pass the same tests, event the load tests (and without the need of rewriting them, of course)

Right now, I only implemented the `ProviderTestsHelper` interface in the AWS provider, so only that provider is being "integration-tested". We now need to do these follow-ups to tests all the providers with the same integration tests:

- Implement the `ProviderTestsHelper` for Azure
- Implement the `ProviderTestsHelper` for Kubernetes
- Implement the `ProviderTestsHelper` for the Local provider
- Add the corresponding script names in the `package.json` file
- Call those scripts from the `integration-tests.yml` and `publish.yml` so that the tests are run in PRs and releases.

## Checks
- [x] Project Builds
- [ ] Updated documentation accordingly. I think it would be ideal to write documentation about the `application-tester` package so that users know how to use it in their application's integration tests. But I would leave that for another PR.


.eslintrc.js
```diff
@@ -28,7 +28,6 @@ module.exports = {
     'linebreak-style': ['error', 'unix'],
     quotes: ['error', 'single', { avoidEscape: true, allowTemplateLiterals: false }],
     semi: ['error', 'never'],
-    '@typescript-eslint/generic-type-naming': ['error', '^T[A-Z][a-zA-Z]+$'],
```
alvaroloes, line 31: After updating `@typescript-eslint` to the latest version, this rule disappeared, so I was getting an error here

.github/workflows/publish.yml
```diff
@@ -59,6 +59,9 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.BOT_AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.BOT_AWS_SECRET_ACCESS_KEY }}
           AWS_REGION: us-east-1
+        # This is not running the integration tests of other providers. Think how this could be done, as they require some other configuration
+        # (as an example, check the k8s tests in "integration-test.yml", it requires a minicube environment)
+        # Maybe running them in different jobs
```
alvaroloes, line 64: This is important. @rdiaz82 not sure if you knew this -> The Kubernetes integration tests are not being run when releasing a new version

packages/application-tester/src/provider-test-helper.ts
```diff
@@ -0,0 +1,22 @@
...
+export interface Queries {
+  events(primaryKey: string): Promise<Array<unknown>>
+  readModels(primaryKey: string, readModelName: string): Promise<Array<unknown>>
+}
+
+export interface ProviderTestHelper {
+  outputs: ApplicationOutputs
+  counters: Counters
+  queries: Queries
+}
```
alvaroloes, line 22: These are the interfaces each provider needs to implement so that they can use the same integration tests we already have

packages/application-tester/src/token-helper.ts
```diff
@@ -0,0 +1,32 @@
+import * as fs from 'fs'
+import * as jwt from 'jsonwebtoken'
+import * as path from 'path'
+
+/**
+ * This helper will create a valid token using a real private key for testing
+ * The keyset file is expecgted to be located in "<package root>/keys/private.key file"
+ */
+export class TokenHelper {
```
alvaroloes, line 9: Refactored directly from the (now deleted) `util.ts` file

packages/framework-integration-tests/integration/helper/app-helper.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function applicationName(): string {
+  return `my-store-${process.env.BOOSTER_APP_SUFFIX}`
+}
+
+export async function getProviderTestHelper(): Promise<ProviderTestHelper> {
+  const provider = process.env.TESTED_PROVIDER
+  const providerHelpers: Record<string, () => Promise<ProviderTestHelper>> = {
+    AWS: () => AWSTestHelper.build(applicationName()),
+  }
```
alvaroloes, line 15: Once a provider has implemented the `ProviderTestHelper`, we can add it here

packages/framework-integration-tests/integration/provider-specific/aws/deployment/deployment.integration.ts
```diff
@@ -0,0 +1,15 @@
...
+import { expect } from '../../../helper/expect'
+
+describe('After deployment', () => {
+  describe('the stack', () => {
+    it('has been created successfully', async () => {
+      // The project must have been deployed by the deploy hook in setup.ts
+      // that scripts uses the cli to do the deployment, so we just check here
+      // that the Cloudformation was run by AWS successfully. For that, we can just
+      // build the AWSHelper. It will throw if the AWS stack is not ready.
+      await expect(AWSTestHelper.build(applicationName())).to.be.eventually.fulfilled
```
alvaroloes, line 12: This test does is provider **specific**, so we can use directly the AWSTesterHelper. 
The deploy and nuke commands are still provider-specific. Maybe we can abstract them too.

juanjoman (changes requested):
I have some minor questions 🙏

packages/application-tester/src/graphql-helper.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  private getApolloHTTPLink(): HttpLink {
+    return new HttpLink({
+      uri: this.providerTestHelper.outputs.graphqlURL,
+      fetch,
+    })
+  }
+
+  private getAuthLink(authToken?: string | (() => string)): ApolloLink {
```
Moneyba, line 64: I really don't understand what a type '() => string' is. Where could I find more information?

packages/framework-integration-tests/integration/helper/app-helper.ts
alvaroloes: 🤔  This are the integration test of the framework itself, so we test those providers we officially support. 
The only way to do that is to add the provider to the source code and add a line here.

BUT! the correct® way to do that is: That someone will include the integration tests in their repo, as it is done with the auth package. They could use the new `application-tester` package to help with that

packages/framework-integration-tests/integration/provider-unaware/cli/cli.project.integration.ts
```diff
@@ -298,7 +298,7 @@ describe('Project', () => {
         })
 
         it('passes linter', async () => {
-          await expect(exec('npm run lint:check', { cwd: projectPath(projectName) })).to.be.eventually.fulfilled
+          await expect(exec('npm run lint:check', { cwd: projectPath(projectName), capture: ['stderr', 'stdout']})).to.be.eventually.fulfilled
```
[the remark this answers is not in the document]
alvaroloes: To be honest, I don't remember having done this change. I also don't know what the `capture` param does. What NIck says makes sense. 

I will verify why this appeared here. Maybe I need a rebase

packages/framework-integration-tests/package.json
```diff
@@ -10,41 +10,41 @@
...
-    "@boostercloud/framework-core": "^0.16.0",
-    "@boostercloud/framework-provider-aws": "^0.16.0",
-    "@boostercloud/framework-provider-kubernetes": "^0.16.0",
-    "@boostercloud/framework-provider-local": "^0.16.0",
-    "@boostercloud/framework-types": "^0.16.0",
+    "@boostercloud/framework-core": "0.16.0",
+    "@boostercloud/framework-provider-aws": "0.16.0",
+    "@boostercloud/framework-provider-kubernetes": "0.16.0",
+    "@boostercloud/framework-provider-local": "0.16.0",
+    "@boostercloud/framework-types": "0.16.0",
```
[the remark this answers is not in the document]
alvaroloes: @NickSeagull `^0.16.0` means `>= 0.16.0 && < 1.0.0`

We had a lot of problems with some dependencies (most of them from AWS) that were updated to newer versions unnoticedly (permitted by the ^) that caused incompatibilities really hard to discover.

Also, the ^ allows that different people had slightly different environments, making troubleshooting a nightmare (and things like "it works on my machine" more likely to appear).

That's why I proposed to use strict versioning as much as possible. If there is a security patch, Github will tell us (or any other tool) and we can update manually.

I wrote a contributing hide to encourage that (I remember other companies also adopted this). Check the final paragraph of this section in the CONTRIBUTING: https://github.com/boostercloud/booster/blob/main/CONTRIBUTING.md#understanding-the-lerna-monorepo-approach-and-how-dependencies-are-structured-in-the-project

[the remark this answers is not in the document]
alvaroloes: @juanjoman I think and hope they are not rewritten! 🤞

packages/application-tester/src/graphql-helper.ts
alvaroloes: Love you asked @Moneyba ! It is a type for a function that takes no parameters and returns a string. More information here: https://www.typescriptlang.org/docs/handbook/2/functions.html

NickSeagull (approved):

juanjoman (approved):

alvaroloes:
There is one integration test failing (kubernetes) that's fixed in PR https://github.com/boostercloud/booster/pull/833
Merging this now

PR #820, opened 2021-05-10
Switch interfaces to classes on new:type command

juanjoman:
## Description
Due to some limitations with the GraphQL API, the commands/read-models property types should be classes instead of interfaces.

## Changes
* Change `boost new:type` CLI command to generate a class instead of an interface
* Modified existing tests
* Added docs for this new limitation

## Checks
- [X] Project Builds


alvaroloes (commented):
Amazing! I left some suggestions in the documentation

PR #825, opened 2021-05-11
Added custom role claim field

charlietfe:
## Description

Instead of forcing the JWT Auth provider to use the `custom:role` field to include the user role inside the `accessToken`, we've changed the configuration to add rolesClaimField. Anyway the `custom:role` will be the fallback value. 

### Example config:

```
  config.appName = 'demoapp'
  config.provider = AWS.Provider
  config.tokenVerifier = {
    jwksUri: 'https://demoapp.auth0.com/.well-known/jwks.json',
    issuer: 'auth0',
    rolesClaim: 'auth0:scope'
  }
})
```

## Checks
- [X] Project Builds


packages/framework-core/src/booster-token-verifier.ts
```diff
@@ -63,13 +63,16 @@ export class BoosterTokenVerifier {
   }
 
   private tokenToUserEnvelope(decodedToken: any): UserEnvelope {
-    const username = decodedToken.email ?? decodedToken.phone_number
+    const username = decodedToken?.email || decodedToken?.phone_number || decodedToken.sub
```
charlietfe, line 66: If for some reason the provider doesn't gives the email or phone_number will use the sub as the username.

packages/framework-integration-tests/src/config/config.ts
```diff
@@ -43,5 +43,6 @@ Booster.configure('production', (config: BoosterConfig): void => {
     issuer: 'booster',
     // Read the content of the public RS256 cert, used to sign the JWT tokens
     publicKey: fs.readFileSync(path.join(__dirname, '..', '..', 'assets', 'certs', 'public.key'), 'utf8'),
+    rolesClaim: 'booster:role'
```
charlietfe, line 46: Just for test a different rolesClaim field.

adriangmweb (approved):
LGTM! 🚀

PR #833, opened 2021-05-13
Automate huge load tests

alvaroloes:
## Description
This PR refactors and adds load tests in a provider unaware way.

As load tests require heavy load, we use Serverless Artillery to generate that load. When we are setting up the load test, we deploy (unless already deployed) the needed infrastructure of Serverless Artillery. Then each load test sends a script with the specification of all the requests they want to send to the currently tested Booster application.
The usage of the scripts and the execution has been abstracted in the class `ArtilleryExecutor` 

Check inlines comments for more information

## Checks
- [x] Project Builds
~- [ ] Updated documentation accordingly~ Not needed


packages/framework-core/src/services/graphql/websocket-protocol/graphql-websocket-protocol.ts
```diff
@@ -150,7 +150,17 @@ export class GraphQLWebsocketHandler {
...
+      await this.connectionManager.sendMessage(this.config, connectionID, new GraphQLComplete(message.id))
+    } catch (e) {
+      // It could be the case that the client already closed the connection without waiting for stop operation to finish
+      // Log this but ignore it
+      this.logger.info(
+        `Received an exception while sending the "complete" message after stopping the GraphQL operation with id ${message.id}.` +
+          'The client probably closed the connection. Ignoring. ',
+        e
+      )
+    }
```
alvaroloes, line 163: This catch reduces several error we were getting when the client had closed the socket early

packages/framework-integration-tests/integration/provider-unaware/load/artillery-executor.ts
```diff
@@ -0,0 +1,63 @@
...
+  duration: number
+  arrivalRate: number
+}
+
+interface OverrideOptions {
+  phases?: Array<Phase>
+  variables?: Record<string, any>
+}
+
+export class ArtilleryExecutor {
```
alvaroloes, line 20: This is a class that can be used from an integration (or better called load) test to execute any Artillery script using Serverless artillery, allowing to execute thousands of requests per second

packages/framework-integration-tests/integration/provider-unaware/load/artillery-executor.ts
```diff
@@ -0,0 +1,63 @@
...
+      console.info('Serverless Artillery stack is already deployed. Skipping redeployment.')
+    } catch (e) {
+      // The CDK returns an exception when the stack is not found. Deploy it in that case
+      await runCommand('.', `slsart deploy --stage ${this.stage}`)
+    }
+  }
+
+  public async executeScript(scriptName: string, overrideOptions: OverrideOptions = {}): Promise<void> {
+    const scriptContent = this.getScriptContent(scriptName, overrideOptions)
+    await runCommand('.', `slsart invoke --stage ${this.stage} --data '${scriptContent}'`)
```
alvaroloes, line 44: The `slsart` dependency is defined as a devDependency in this package, so it is always found

packages/framework-integration-tests/integration/provider-unaware/load/artillery-executor.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  private getScriptContent(scriptName: string, options: OverrideOptions): string {
```
alvaroloes, line 47: This function reads the script and overrides some values and, especially, the variables, before sending the script to Serverless artillery

packages/framework-integration-tests/integration/provider-unaware/load/entities-are-correctly-reduced.load.ts
```diff
@@ -0,0 +1,77 @@
+import { ApolloClient } from 'apollo-client'
+import { NormalizedCacheObject } from 'apollo-cache-inmemory'
+import { internet, random } from 'faker'
+import gql from 'graphql-tag'
+
+import { waitForIt } from '../../helper/sleep'
+import { applicationUnderTest, scriptExecutor } from './setup'
+
+describe('Data consistency on entities', () => {
```
alvaroloes, line 9: This is the first load test: It sends more than 20,000 events and verifies that the entities (although we query ReadModels, they are a mapping one to one to entities) state has not been corrupted

packages/framework-integration-tests/integration/provider-unaware/load/entities-are-correctly-reduced.load.ts
```diff
@@ -0,0 +1,77 @@
...
+  let client: ApolloClient<NormalizedCacheObject>
+  let token: string
+
+  before(async () => {
+    const userEmail = internet.email()
+    token = applicationUnderTest.token.forUser(userEmail, 'Admin')
+    client = applicationUnderTest.graphql.client(token)
```
alvaroloes, line 16: as we are using `applicationUnderTest`, this test is provider unaware

packages/framework-integration-tests/integration/provider-unaware/load/entities-are-correctly-reduced.load.ts
```diff
@@ -0,0 +1,77 @@
...
+          {
+            duration: durationWarmup,
+            arrivalRate: arrivalRateWarmup,
+          },
+          {
+            duration: durationBurst,
+            arrivalRate: arrivalRateBurst,
+          },
+        ],
+      })
```
alvaroloes, line 50: We send those many events in two batches. The first one has a slower rate (only 200 events per second) to give time to lambdas to warm up. The second one will send 2000 events per second

If we don't do this, we will probably get throttling exceptions from the lambda service

packages/framework-integration-tests/integration/provider-unaware/load/read-models-are-correctly-projected.load.ts
```diff
@@ -0,0 +1,54 @@
+import { ApolloClient } from 'apollo-client'
+import { NormalizedCacheObject } from 'apollo-cache-inmemory'
+import { internet, random } from 'faker'
+import { expect } from '../../helper/expect'
+import gql from 'graphql-tag'
+
+import { waitForIt } from '../../helper/sleep'
+import { applicationUnderTest, scriptExecutor } from './setup'
+
+describe('Data consistency on read models', () => {
```
alvaroloes, line 10: This is the second load test. It is focused mainly on data corruption of read models when just one instance aggregates a lot of entities. This exercises the optimistic concurrency

packages/framework-integration-tests/integration/provider-unaware/load/scripts/create-products-same-sku.yml
```diff
@@ -0,0 +1,16 @@
+config:
+  target: <not_set>
```
alvaroloes, line 2: This is the artillery script used in one of the tests. As you can see there are many variables with the value `<not_set>`. The value itself doesn't matter, it is just a marker to know which of those variables will be overridden by the typescript code of the integration test.

alvaroloes: Note: the `target` is always overridden by the ArtilleryExecutor class in a provider unaware way (it just takes the `graphqlURL` of the provider under test)

packages/framework-integration-tests/src/commands/create-product.ts
```diff
@@ -13,12 +13,14 @@ export class CreateProduct {
...
-    readonly currency: string
+    readonly currency: string,
+    readonly productID?: UUID,
   ) {}
 
   public static async handle(command: CreateProduct, register: Register): Promise<void> {
+    const productID = command.productID ?? UUID.generate()
     register.events(
-      new ProductCreated(UUID.generate(), command.sku, command.displayName, command.description, {
+      new ProductCreated(productID, command.sku, command.displayName, command.description, {
```
alvaroloes, line 23: This change was needed to facilitate some tests

packages/framework-integration-tests/src/read-models/stock-read-model.ts
```diff
@@ -0,0 +1,17 @@
...
+})
+export class StockReadModel {
+  public constructor(readonly id: UUID, readonly warehouses: Record<string, number>) {
+  }
+
+  @Projects(Stock, 'id')
+  public static projectStock(entity: Stock): ProjectionResult<StockReadModel> {
+    return new StockReadModel(entity.id, entity.warehouses)
+  }
+}
```
alvaroloes, line 17: This read model is a one-to-one mapping with the Stock entity so that we can check the final status in the load test

packages/framework-integration-tests/integration/provider-unaware/load/artillery-executor.ts
```diff
@@ -0,0 +1,63 @@
...
+    private readonly stage = 'booster'
+  ) {}
+
+  public async ensureDeployed(): Promise<void> {
+    try {
+      await cloudFormation
+        .describeStacks({
+          StackName: `${this.serverlessArtilleryStackPrefix}-${this.stage}`,
+        })
+        .promise()
```
davidverdu, line 34: One question. This code for load tests is provider unaware but for launching the Artillery tests it creates a resource in AWS to run from the cloud, regardless of the provider you are testing?

NickSeagull: Yes, because it is used to run the load test, not the Booster project, it is the same as running the integration tests on GitHub actions

NickSeagull (approved):
Thanks a lot for this!

packages/framework-integration-tests/integration/provider-unaware/load/artillery-executor.ts
alvaroloes: Yeah, the usage of Artillery has nothing to do with the provider you use for Booster. It is just a tool that happens to use AWS infrastructure to run. 

For example, you can execute the load tests using the local provider. This means that the server that will receive the load will be the local provider but the client generating all the requests will be Serverless Artillery, which will be hosted in AWS (which is powerful enough to generate the load).

So, we will use Serverless Artillery hosted in AWS to generate the load needed to execute the load test for any provider.

Does it make sense to you now, @davidverdu ?

davidverdu: It makes sense!

davidverdu (approved):
LGTM. This change it's HUGE! Thanks for it.

davidverdu:
It works in my local machine with node 15 and npm 7+

adriangmweb (approved):
Awesome! 🚀

Thank you a lot for taking the time on adding comments, it really made easier the review when going through those files!

PR #840, opened 2021-05-21
Local provider GraphQL filters compatibility

adriangmweb:
## Description
Modify Local provider searcher-adapter.ts to be able to use the new GraphQL filters like the rest of the providers.

## Changes
Modified the provider's `searcher-adapter.ts` to make it compatible with the new format.

## Checks
- [x] Project Builds
 
## Additional information
Now the local provider GraphQL API is at the same state as the Azure and the K8s ones.

Closes: #601

### DIFF
--- a/packages/framework-provider-local/src/library/read-model-adapter.ts
+++ b/packages/framework-provider-local/src/library/read-model-adapter.ts
@@ -1,6 +1,6 @@
 import {
   BoosterConfig,
-  FilterOld,
+  FilterFor,
   Logger,
   ReadModelEnvelope,
   ReadModelInterface,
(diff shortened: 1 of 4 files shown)


davidverdu (approved):
LGTM.

It makes the booster-blog application work without subscriptions

juanjoman (approved):

alvaroloes (approved):
Great! 

There are no integration tests for this, but I think it is ok as we are now going to use the current provider-unaware ones with the local provider too.

So :shipit:

davidverdu:
> Great!
> 
> There are no integration tests for this, but I think it is ok as we are now going to use the current provider-unaware ones with the local provider too.
> 
> So :shipit:

The provider-unaware tests are scheduled to be added to the Local Provider soon @alvaroloes

PR #842, opened 2021-05-26
GitHub actions nuke AWS stack when was deployed

### LINKED ISSUE #758: AWS Buckets are not being cleaned up and removed when integrations test fails
# Bug Report

## Current Behavior

We run the integration test and when they fail, the buckets created in AWS persist until we manually remove them.

## Expected behavior

When the integration test fails, the buckets are cleaned and removed properly.

### Environment

- Booster version: [e.g. 0.8.2] 
- Node/npm version: [e.g. Node 8/npm 5]
- OS: [e.g. OSX 10.13.4, Windows 10]

adriangmweb:
## Description

Github actions integration tests are defined by consequent jobs with steps. There is a problem where when an integration test fails after deploying the stack to AWS, the rest of the steps get skipped, including the `nuke` one. This is causing that many stacks are created in AWS without being deleted after running the tests, so AWS limits are being hit.

This PR includes a solution that makes the `nuke` job run if the `deployment` step has succeeded, even if any integration test in the middle fails.

## Changes

A conditional expression was added to the GitHub actions `YAML` file to make the nuke job always run when the deployment has succeeded. Also, the property `continue-on-error` is set to true, so it doesn't get skipped if any other job has failed.

## Checks
- [x] Project Builds
 
## Additional information

This change has been tested with the integration tests passing and failing, and it always removes the stack.

Closes #758

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -76,6 +76,8 @@ jobs:
 
   aws-nuke-integration-tests:
     name: AWS Nuke Integration Tests
+    if: always() && needs.aws-deploy-integration-tests.result == 'success'
+    continue-on-error: true
     needs:
       - aws-load-tests
     runs-on: ubuntu-latest


.github/workflows/integration-tests.yml
```diff
@@ -76,6 +76,8 @@ jobs:
 
   aws-nuke-integration-tests:
     name: AWS Nuke Integration Tests
+    if: always() && needs.aws-deploy-integration-tests.result == 'success'
+    continue-on-error: true
     needs:
       - aws-load-tests
```
juanjoman, line 82: This rule has a higher priority than the if above right?

juanjoman (approved):
LGTM!! Just left a question :v:

.github/workflows/integration-tests.yml
adriangmweb: It just let him know that it has to run after the `aws-load-tests` job, but the conditional about plus the `continue-on-error` property set to true, makes it run even when the previous step fails if the deployment job succeeds.

juanjoman: Wonderful 👍

Issue #807, opened 2021-05-03
Add "before" and "after" hooks to ReadModels and, optionally, to Commands


alvaroloes:
## Description

Although the proposed solution for this issue covers several use cases, I'm going to explain the need for this with one use case that has more priority:

Right now, we have the possibility to authorize reading a ReadModel only to specific roles. This is great but not enough for any common application. Consider the following ReadModel:
```typescript
@ReadModel({
  authorize: [User]
})
class UserProfile {
   //...
}
```
Every client that is logged in with the role "User" can access **any instance of UserProfile**. This is normally incorrect and insecure. What we really want it to _allow accessesing the instance only if the current user is the owner of that profile_.

That is currently impossible in Booster, but needed for any application (for example Mayday would need this)

In Commands, this problem is not that important, because the user can check if the current user can access that command inside the command handler and throw an exception if they can't.
However, ReadModels doesn't have any place to add these kinds of checks, and that's what this this issue is for.

## Possible Solution
The proposed solution here adds a general mechanism that solves the above problem, but can also be used to other kind of use cases, like transforming data before sending it to the client, adding extra filters before executing the query, etc.

Proposal: Add two "hooks" to read models:

1. **before** hook: Executed before the query is made. It receives the `FilterFor` object with all the filter the user has specified and the current `UserEnvelope`. It can either throw exceptions or return a new `FilterFor` object to be used in the actual query sent to the provider. Very useful to reject requests under some conditions or further refine the filter before doing the actual query.
2. **after** hook: Executed after the query is made. It receives the array of ReadModels that are about to be sent to the user and the current `UserEnvelope`. It can throw exceptions or return the actual array of read models to be sent back to the user. Very useful for data transformations related to the presentational aspect.

Example:
```typescript
@ReadModel({
  authorize: [User]
})
class UserProfile {
  @Before // <--- This is a "before" hook
  public static checkID(filter: FilterFor<UserProfile>, currentUser?: UserEnvelope): FilterFor<UserProvile> {
    // Role "User" can only access their profile
    if (filter.id !== currentUser.id) {
      throw new NotAuthorizedError("...")
    }
    return filter
  }
  
  @After // <--- This is an "after" hook
  public static useProperCurrency(profiles: Array<UserProfile>, currentUser?: UserEnvelope): UserProfile {
    // We store all payments in cents of euros, but the user can choose the currency they want their money to be deployed in
    for(profile of profiles) {
      switch(profile.currency) {
        case 'dollars':
[...]

PR #850, opened 2021-06-04
Read models' before hooks

juanjoman:
## Description
This is the first part of https://github.com/boostercloud/booster/issues/807 (before hook part). With this feature we want to allow the user to modify read model filters (or throw exceptions if something wrong happens) before fetching the database.

## Changes
* Make `filters` never undefined, but an empty object. This is useful to apply validations for general queries too without setting the filter param as optional.
* Create the before hook for read models.
* Modify `UserEnvelope` and `tokenVerifier` so they can accept custom parameters.

## Checks
- [X] Project Builds

### Additional information
For `UserEnvelope` to correctly process custom parameters, I need to change some logic in the auth rocket, which is my next priority.


packages/framework-core/src/booster-read-models-reader.ts
```diff
@@ -57,12 +61,34 @@ export class BoosterReadModelsReader {
...
+    filters: FilterFor<Class<ReadModelInterface>>,
+    user?: UserEnvelope,
+    beforeHooks?: Array<BeforeFunction>
+  ): FilterFor<ReadModelInterface> {
+    if (beforeHooks) {
+      let currentFilter = filters
+      beforeHooks.map((before: BeforeFunction) => {
+        currentFilter = before(currentFilter, user)
+      })
+      return currentFilter
```
juanjoman, line 86: I'm 90% sure that this can be done with a reduce function but the one I did it's not working (??)

alvaroloes: Try replacing this function body with this:
```typescript
return (beforeHooks ?? []).reduce((currentFilter, before) => before(currentFilter, user), filters)
```

packages/framework-core/src/booster-token-verifier.ts
```diff
@@ -67,14 +67,26 @@ export class BoosterTokenVerifier {
...
       id,
       username,
       role: roleValue?.trim() ?? '',
+      ...customParams,
     }
   }
 
+  private getCustomParameters(decodedToken: Record<string, unknown>): Record<string, unknown> {
+    return Object.keys(decodedToken)
+      .filter((key: string) => key.includes('custom:') && key !== 'custom:role')
```
alvaroloes, line 82: @juanjoman Take into account that the claim where the token is can be different. You should use the above variable (line 67) `rolesClaim`.

Anyway take a look at previous comment

packages/framework-core/src/decorators/read-model.ts
```diff
@@ -18,6 +18,7 @@ export function ReadModel(attributes: RoleAccess): (readModelClass: Class<ReadMo
         class: readModelClass,
         properties: getPropertiesMetadata(readModelClass),
         authorizedRoles: attributes.authorize,
+        before: attributes.before,
```
alvaroloes, line 21: What about doing `before: attributes.before ?? []` ?
This way, we can define the `before` field in the metadata as non-null, simplifying some code (as the one I wrote in THIS comment where the before hooks parameter would not be optional anymore)

packages/framework-integration-tests/src/read-models/cart-read-model.ts
```diff
@@ -18,6 +19,16 @@ export class CartReadModel {
     public cartItemsIds?: Array<string>
   ) {}
 
+  public static beforeFn(filter: FilterFor<CartReadModel>, currentUser?: UserEnvelope): FilterFor<CartReadModel> {
+    if (!filter.id || filter.id?.eq !== 'before-fn-test') return filter
+    return { id: filter.id } as FilterFor<CartReadModel>
```
alvaroloes, line 24: 🤔  Would this return be equivalent to `return filter`?

packages/framework-types/src/concepts/read-model.ts
```diff
@@ -14,4 +14,5 @@ export interface ReadModelMetadata {
   readonly class: Class<ReadModelInterface>
   readonly properties: Array<PropertyMetadata>
   readonly authorizedRoles: RoleAccess['authorize']
+  readonly before?: RoleAccess['before']
```
alvaroloes, line 17: In THIS comment I suggested having this non-optional

packages/framework-types/src/concepts/role.ts
```diff
@@ -14,6 +17,12 @@ export interface AuthMetadata {
   readonly skipConfirmation?: boolean
 }
 
+export type BeforeFunction = (
+  filter: FilterFor<Class<ReadModelInterface>>,
+  currentUser?: UserEnvelope
+) => FilterFor<Class<ReadModelInterface>>
```
alvaroloes, line 23: For me, this type makes more sense to be defined in the `./read-model` file, as it has very little to do with roles.  What do you think?

packages/framework-types/src/concepts/role.ts
```diff
@@ -14,6 +17,12 @@ export interface AuthMetadata {
...
+
 export interface RoleAccess {
   readonly authorize: 'all' | Array<Class<RoleInterface>>
+  readonly before?: Array<BeforeFunction>
```
alvaroloes, line 27: The same with this field, I don't think it belongs to this interface. 
I would create an interface called `FilterHooks` (for example) with just this `before` field (for now). That type could be defined in the same way the decorator `@ReadModel` is defined (in the core package).

Then, in the decorator, you can say that the parameter conforms both interfaces, like this:
```typescript
export function ReadModel(attributes: RoleAccess & FilterHooks): (readModelClass: Class<ReadModelInterface>) => void {
...
```
What do you think?

docs/chapters/04_features.md
```diff
@@ -487,6 +487,52 @@ mutation {
 
 > [!NOTE]  Remember that, in case you want to subscribe to a read model that is restricted to a specific set of roles, you must send the **access token** retrieved upon sign-in. Check ["Authorizing operations"](#authorizing-operations) to know how to do this.
 
+
+### Validating read models' filters
```
alvaroloes, line 491: This `before` hooks can be used to more things than just validating, so I would suggest calling this:
```suggestion
### Adding "before" hooks to your read models
```

docs/chapters/04_features.md
```diff
@@ -487,6 +487,52 @@ mutation {
...
+You can validate read models by using the `before` hook. The functions you add here will be executed **before fetching from your event-store**:
```
alvaroloes, line 492: Take into account that read models are not stored in the event-store, but in independent tables.
Also, when writing documentation, try to think about the reader and give them clues regarding why the thing you are explaining is useful for them. Be empathic with the reader. Remember the hours you have spent reading bad documentation and how you felt!

```suggestion
When you send queries or subscriptions to your read models, you can tell Booster to execute some code before executing the operation. These are called `before` hooks and they receive two parameters: the read model filter sent with the request and an object with information about the currently signed-in user.
These `before` hooks are useful for many things, but mainly:
- To add fine-grained access control. For example, to allow user A access to its Cart read models, but not the ones belonging to user B.
- Modify the read model filters sent with the request based on some criteria. For example, you can narrow the results the users will get by adding extra filters based on the season of the year.

From inside a `before` hook, you can either throw an exception, in which case the query or subscription will be aborted and the error will be sent back to the user, or return final filters to be used in the operation.

This is an example of how you would define a before hook in a read model: 
```

docs/chapters/04_features.md
```diff
@@ -487,6 +487,52 @@ mutation {
...
+  ) {}
+
+  public static validateUser(filter: FilterFor<CartReadModel>, currentUser?: UserEnvelope): FilterFor<CartReadModel> {
+    if (filter?.userId !== currentUser.id) throw NotAuthorizedError("...")
+    return filter
+  }
+}
+```
+
+As an addition, functions can also be chained:
```
alvaroloes, line 512:
```suggestion
You can define more than one `before` hook for a read model. The filter resulting from the call to one hook will be used as a parameter for the next one. Therefore, _the order in which they are specified matters_:
```

docs/chapters/04_features.md
```diff
@@ -487,6 +487,52 @@ mutation {
...
+  before: [CartReadModel.validateUser],
+})
+export class CartReadModel {
+  public constructor(
+    readonly id: UUID,
+    readonly userId: UUID
+  ) {}
+
+  public static validateUser(filter: FilterFor<CartReadModel>, currentUser?: UserEnvelope): FilterFor<CartReadModel> {
+    if (filter?.userId !== currentUser.id) throw NotAuthorizedError("...")
```
alvaroloes, line 506: Remember that filters are no longer optional, so you can remove the `?` after filter

docs/chapters/04_features.md
```diff
@@ -487,6 +487,52 @@ mutation {
...
+    return filter // This filter will be passed as a parameter to the validateEmail function
+  }
+
+  public static validateEmail(filter: FilterFor<CartReadModel>, currentUser?: UserEnvelope): FilterFor<CartReadModel> {
+    if (!filter?.email.includes('myCompanyDomain.com')) throw NotAuthorizedError("...")
```
alvaroloes, line 530: You can remove the `?` here an in the above hook

alvaroloes (changes requested):
Left some suggestions specially in the documentation. 
The PR is great, but remember to also take special care with the documentation. We need to try to be empathic with the reader and help them as much as possible to understand what we are explaining. 

So it is always useful to talk about possible use cases of this feature, tell them suggestions, etc.

In brief, don't forget to be mindful and take documentation seriously ;-)

packages/framework-core/src/booster-token-verifier.ts
```diff
@@ -67,14 +67,26 @@ export class BoosterTokenVerifier {
...
+      .reduce((obj: Record<string, unknown>, key: string) => {
+        const cleanKey = key.split(':')[1]
+        obj[cleanKey] = decodedToken[key]
+        return obj
+      }, {})
+  }
```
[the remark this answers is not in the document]
juanjoman: Hmm I think that carries only role info and nothing else. It's created here I think: https://github.com/boostercloud/rocket-auth-aws-infrastructure/blob/main/src/lambdas/sign-up.ts#L60

At this point, the token has been already verified (that's why I'm using the `decodedToken` object) but let me know if I'm missing something 🙏

packages/framework-core/src/booster-token-verifier.ts
juanjoman: ooooh you're right, thanks!

packages/framework-types/src/concepts/role.ts
juanjoman: Agree! :D

packages/framework-types/src/concepts/role.ts
juanjoman: Definitely, it has nothing to do with roles now that I think about it lol

packages/framework-integration-tests/src/read-models/cart-read-model.ts
juanjoman: You're right! thanks!!

packages/framework-core/src/booster-token-verifier.ts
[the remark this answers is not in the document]
alvaroloes: It carries more info, including any custom claim that the auth provider the user is using might add. This is the reason I think it could be more useful (and easier on our side) to include all.
Take into account that the reason to include only those tockens prefixed with `custom:` is arbitrary. It is true that Cognito uses those, but other auth providers or proxies use others. For example, Hasura uses something like `x-hasura-blahblah`. If we just include those with the `custom:` prefix, we would be limiting a lot the usefulness.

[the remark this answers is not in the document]
alvaroloes: BTW, I didn't fully understand your last sentence

packages/framework-integration-tests/src/read-models/cart-read-model.ts
```diff
@@ -21,6 +22,16 @@ export class CartReadModel {
...
+
+  public static beforeFn(filter: FilterFor<CartReadModel>, currentUser?: UserEnvelope): FilterFor<CartReadModel> {
+    return filter
+  }
+
+  public static beforeFnV2(filter: FilterFor<CartReadModel>, currentUser?: UserEnvelope): FilterFor<CartReadModel> {
+    if (!filter.id || filter.id?.eq !== 'before-fn-test') return filter
+    return { id: { eq: filter.id.eq + '-modified' } } as FilterFor<CartReadModel>
+  }
+
```
davidverdu, line 34: This is an amazing way to test before filters. Just one comment: Don't we need a test case where an exception is thrown inside a before filter, so that the read model query doesn't return results?

davidverdu (changes requested):
LGTM except for two things:
- I miss an integration test testing a thrown exception inside a before filter
- AWS integration tests are failing in my machine. @adriangmweb solved the problem in his PR https://github.com/boostercloud/booster/pull/878, since apparently it was a dependency problem

packages/framework-integration-tests/src/read-models/cart-read-model.ts
juanjoman: Yes! I forgot about that, thanks!

davidverdu (approved):
Now that the exception is being tested, LGTM (Provided that AWS integration tests run properly). I unblock my requested changes, so that you can merge it after having another approval

packages/framework-integration-tests/integration/provider-unaware/end-to-end/subscriptions.integration.ts
```diff
@@ -156,17 +143,52 @@ describe('subscriptions', () => {
...
+      // Wait for for the subscriptions to arrive
+      await waitForIt(countSubscriptions, (newCount) => newCount == originalSubscriptionsCount + 1)
+      // Check we receive data when the read model is modified
+      await cartMutation(client, 'before-fn-test-modified', beforeHookProductId)
+
+      const result = await subscriptionPromise
+      const cart = result.data.CartReadModel
+
+      expect(cart.id).to.equal('before-fn-test-modified')
+    })
```
alvaroloes, line 190: Amazing!

alvaroloes (approved):
Left some minor comments. Approving it anyway!

juanjoman:
K8s tests are passing on local from scratch 😓

packages/framework-types/src/concepts/read-model.ts
```diff
@@ -14,4 +16,14 @@ export interface ReadModelMetadata {
   readonly class: Class<ReadModelInterface>
   readonly properties: Array<PropertyMetadata>
   readonly authorizedRoles: RoleAccess['authorize']
+  readonly before: NonNullable<FilterHooks['before']>
 }
+
+export interface FilterHooks {
+  readonly before?: Array<BeforeFunction>
+}
```
NickSeagull, line 24: # This is the part that should be modified accordingly to my explanation above

NickSeagull (changes requested):
I've left a possible change in the API. I'm sure we can make it work with the better UX and retain the type safety

docs/chapters/04_features.md
```diff
@@ -487,6 +487,60 @@ mutation {
...
+- To add fine-grained access control. For example, to allow user `A` access to its Cart read models, but not the ones belonging to user `B`.
+
+- Modify the read model filters sent with the request based on some criteria. For example, you can narrow the results the users will get by adding extra filters based on the season of the year.
+  From inside a `before` hook, you can either throw an exception, in which case the query or subscription will be aborted, and the error will be sent back to the user, or return final filters to be used in the operation.
+  This is an example of how you would define a before hook in a read model:
+
+```typescript
+@ReadModel({
+  authorize: [User],
+  before: [CartReadModel.validateUser],
```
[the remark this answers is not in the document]
juanjoman: Hmm the problem I see about this solution is user experience. To me is much easier to let the IDE autocomplete the function based on the class while typing. Maybe when I introduce the string, it's not right and I have to recheck if I mistyped it. Also, it simplifies how is passed to the core as it's a normal function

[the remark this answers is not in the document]
alvaroloes: One of the ideas of the before hooks is that you can put their functions that are defined anywhere. They can be functions in other modules of your project or functions that are in a third-party library. This is very powerful because we can publish a module with common before hooks or you can use a module with authorization hooks that is defined in a specific Auth rocket, for example.

The example that Juanjo used here happens to use static functions defined inside the read model class, but this is just a coincidence.

If we use strings, we would loose the ability to use any function.

docs/chapters/04_features.md
```diff
@@ -487,6 +487,60 @@ mutation {
...
+  ) {}
+
+  public static validateUser(filter: FilterFor<CartReadModel>, currentUser?: UserEnvelope): FilterFor<CartReadModel> {
+    if (filter?.userId !== currentUser.id) throw NotAuthorizedError("...")
+    return filter
+  }
+}
+```
+
+You can define more than one `before` hook for a read model. The filter resulting from the call to one hook will be used as a parameter for the next one. Therefore, _the order in which they are specified matters_:
```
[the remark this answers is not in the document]
alvaroloes: Agree, maybe we can use an information box or whatever its called, I don't know the syntax to do that, but I have seen that some parts of the documentation use those boxes

packages/framework-types/src/concepts/read-model.ts
alvaroloes: 😂 😂 😂 😂 😂  It seems you are shouting! 

Anyway, check my response to that comment 😉

docs/chapters/04_features.md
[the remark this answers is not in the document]
NickSeagull: If its the case for using any kind of function then I'd rephrase the example with the hooks outside of the class instead of being static methods.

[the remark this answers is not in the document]
juanjoman: I finally mixed the example with static and external methods 👯

docs/chapters/04_features.md
```diff
@@ -487,6 +487,65 @@ mutation {
...
+
+> [!NOTE] The order in which filters are specified matters.
+
+```typescript
+import { checkAuth } from '../../auth' // You can also use external functions!
+
+@ReadModel({
+  authorize: [User],
+  before: [CartReadModel.validateUser, CartReadModel.validateEmail, checkAuth],
+})
```
NickSeagull, line 530: What I meant is to avoid showing the usage of static methods altogether. Put a regular function outside of the class and use it. As the static thing is more verbose

NickSeagull (approved):
Lovely, thanks!

juanjoman:
Just in case someone asks: Load tests failing is a known issue.

Issue #797, opened 2021-04-22
Create real class instances when passing events, entities and read models to user's reducer and projections


alvaroloes:
# Bug Report
Right now, when we are reducing the events and we call the user-defined reducer function, we do it this way:
https://github.com/boostercloud/booster/blob/64c1dc984a602efbf54f2fc5b50c41d1b33d27db/packages/framework-core/src/services/event-store.ts#L104-L105

And when we are calling the user-defined projection function, we do it this way:
https://github.com/boostercloud/booster/blob/64c1dc984a602efbf54f2fc5b50c41d1b33d27db/packages/framework-core/src/services/read-model-store.ts#L73-L75

In both cases, the parameters we are passing to the functions (the event and current entity for the reducer, and the entity and the current read model for the projection) are **_raw javascript objects and not instances of their respective classes_** which is incorrect.

This leads to unexpected bugs in the user's code if they define instance methods in their event, entity, or read model classes and call them inside the reducers or projections, as those methods won't be found in those raw javascript objects.

For example if I have an event and entity like the following, the method calls in the reducer will produce errors:
```typescript
@Event
class PersonCreated {
  // ...
  public fullName(): string {
    return this.firstName + ' ' + this.lastName
  }
}

@Entity
class Person {
  //...
  @Reduces(PersonCreated)
  public static reducePersonCreated(event: PersonCreated, currentEntity: Person {
    //...
    event.fullName() // Error: Method `fullName` not found in "Object" (as "event" is not a real instance of "PersonCreated")
    currentEntity.getAgeInDays() // Error: Method `getAgeInDays` not found in "Object" (as "currentEntity" is not a real instance of "Person")
    //...
    
  }
  
  public getAgeInDays(): number {
    //...
  }
}
```
## Current Behavior
The method calls showed in the previous example throw errors because those methods are not found. The same happens in read-model projections if we call methods on the entity or current read model.

## Expected behavior
The methods should be found, as its receiver should be a proper class instance

## Possible Solution
The solution is to create proper instances for the event, entity, and read model and call the user's reducers and projections with those instances as parameters. 
The solution is pretty similar (we can do exactly the same) as we are doing right now when we call command handlers:
https://github.com/boostercloud/booster/blob/64c1dc984a602efbf54f2fc5b50c41d1b33d27db/packages/framework-core/src/booster-command-dispatcher.ts#L34-L36

## Additional information
[...]

PR #854, opened 2021-06-04
Fix/real class instances

Moneyba:
## Description
Currently, when you pass events, entities and read models to user's reducer and projections , you get a raw javascript object instead of an instance of their respective classes.
We have the same bug when calling event handlers: we are passing as a parameter the raw object stored in the event envelope, instead of the instance.
fixes: https://github.com/boostercloud/booster/issues/797
## Changes
Create proper instances for the event, entity, read model and event handlers.

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-core/src/booster-command-dispatcher.ts
+++ b/packages/framework-core/src/booster-command-dispatcher.ts
@@ -9,6 +9,7 @@ import {
 } from '@boostercloud/framework-types'
 import { BoosterAuth } from './booster-auth'
 import { RegisterHandler } from './booster-register-handler'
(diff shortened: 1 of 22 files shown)


NickSeagull:
It didn't get linked for some reason here, but FYI : https://github.com/boostercloud/booster/pull/855/files

davidverdu (approved):
LGTM. I would rebase just to make merging softer (changes in package.json @types/express dependencies are already merged in a previos PR)

alvaroloes (approved):
Thanks for the changes!

PR #856, opened 2021-06-07
Fix incompatibility of node versions 15.5+ and serverless-artillery causing "Empty lambda zip" errors in load tests

alvaroloes:
## Description
This is a "dependency-hell" chained error:

1. From node v15.6.0 onwards, the behavior of `.end` on `zlib.DeflateRaw` [was changed](https://github.com/nodejs/node/issues/37027)
2. This caused the library `archiver` to [fail](https://github.com/archiverjs/node-archiver/issues/491). Only the newest version has it fixed.
    - This library is used by the AWS CDK, so we started getting an error when deploying because the lambda's `.zip` were empty. We fixed this by updating the AWS CDK that used a newer version of `archiver`
3. `serverless-artillery` uses `serverless` to deploy the lambda that executes de load test. `serverless`, in turn, uses `archiver` to zip the lambda's code. This means that we were getting the "empty lambda zip" error when deploying serverless-artillery for the load tests.
4. The newest version of `serverless` uses the fixed version of `archiver`, but we can't use it because it is incompatible with the version we are using of `serverless-artillery`.
5. We can't update `serverless-artillery` because there is not newer version :trollface: 

## Changes
The solution is to force the version of `serverless` we are using (that's compatible with `serverless-artillery`) to use a newer version of `archiver`.
The official way to do this (which is called "override dependencies") has still not reached the official version of node and NPM (official issue [here](https://github.com/npm/statusboard/issues/343))

The "un-official" way to do this is what is done in this PR:
- We are hoisting all the dependencies of the project. This means to put all the dependencies of each package in the root `node_modules` folder (instead of inside each package)
- Therefore, the good version of `archiver` is located at `<root>/node_modules/archiver`
- The bad version, the one that `serverless` is using, is located in a transitive dependency in `<root>/node_modules/serverless/node_modules/archiver`
- The solution is to remove the bad module. 
- This way, when the `serverless` module looks for `archiver` and it doesn't find it in its local "node_modules" folder, it will search for it in the parent directory (this is how every node module lookup is done), finding the good `archiver`

## Checks
- [x] Project Builds
 
## Additional information
image


packages/framework-integration-tests/integration/provider-unaware/load/artillery-executor.ts
```diff
@@ -26,17 +27,28 @@ export class ArtilleryExecutor {
...
-      await runCommand('.', `slsart deploy --stage ${this.stage}`)
+      // The CDK returns an exception when the stack is not found. Ignore it and deploy
+    }
+    const validStatuses = ['CREATE_COMPLETE', 'UPDATE_COMPLETE']
+    if (slsartStack) {
+      if (validStatuses.includes(slsartStack?.StackStatus)) {
+        console.info('Serverless Artillery stack is already deployed. Skipping redeployment.')
+        return
+      }
+      console.info(`Serverless Artillery stack is in a wrong state: ${slsartStack?.StackStatus}. Redeploying.`)
```
alvaroloes, line 49: I enhanced the check we were doing to see if we need to redeploy the `serverless-artillery` stack

packages/framework-integration-tests/package.json
```diff
@@ -59,6 +59,7 @@
     "lint:fix": "eslint --quiet --fix --ext '.js,.ts' **/*.ts",
     "compile": "ttsc -b tsconfig.json",
     "clean": "rimraf ./dist tsconfig.tsbuildinfo",
+    "postinstall": "rimraf ../../node_modules/serverless/node_modules/archiver && echo 'Deleted buggy \"archiver\" module from \"serveless\" dependencies to use the newer hoisted one'",
```
alvaroloes, line 62: This is the main change in this PR

NickSeagull (approved):
Nice one! Thanks :D

juanjoman (approved):
Amazing 🙌

Moneyba (approved):

PR #860, opened 2021-06-08
Feature/812 cli command to create test files

### LINKED ISSUE #812: New CLI command to create test files
## Feature Request
New CLI command to create test files

## Description
As I know there is no commands to create test file. When I was looking at new docs testing section I though that it has some code that will be repeated in each test file. 

```ts
const config = new BoosterConfig('test')
config.appName = 'testing-time'
``` 

Booster CLI command would create the test file to help developers to start writing new tests without copy-pasting same code.

## Possible Solution
Create command that will generate test file under `tests/` folder.

- Command: `boost new:test [name]`

- Minimal skeleton content of created test file:

```ts
import { expect } from '../expect'

const config = new BoosterConfig('test')
config.appName = 'testing-time' // 'testing' + [app name] ?

describe('[name]', () => {
  it('should...', () => {
    expect(true).to.be.true
  })
})

```

Fecony:
## Description
CLI command to create example test file. 

```bash
boost new:test SimpleTest
```

Will create test file (`simple-test.test.ts`) with following content in `test/` folder:

```ts
import { BoosterConfig } from '@boostercloud/framework-types'
import { expect } from 'chai'

const config = new BoosterConfig('test')
config.appName = 'testing-booster-app' // <- app name

describe('[SimpleTestTest]', () => {
  it('should be true', () => {
    expect(true).to.be.true
  })
})
```

### Test files use `chai` that is not installed by default in new projects. 

> Maybe it should be included or should print message after test file was created. "Run `npm install -D chai @types/chai` to install required dependencies" ?

## Changes
- Add `new:test` command to generate test file
- Add new template for test files
- Tests to test simple cases that test file is generated

## Checks
- [x] Project Builds
~- [] Updated documentation accordingly~

## Additional Information
Closes #812 

This command may be expanded to: 
- allow users create tests in subfolders
- check if user has recommended `expect.ts` file
- use another testing tools (as flag ?)


NickSeagull:
Thanks for this @Fecony ! Actually this is a feature that the core team has been discussing internally, so this pushes it forward!

I'd like to hear more from the @boostercloud/booster-core team. Maybe @juanjoman can shed some light on his thoughts (I believe you were the one working on the draft on testing, right Juanjo?)

packages/cli/src/commands/new/test.ts
```diff
@@ -0,0 +1,62 @@
...
+  return [
+    {
+      packagePath: '@boostercloud/framework-types',
+      commaSeparatedComponents: 'BoosterConfig',
+    },
+    {
+      packagePath: 'chai',
+      commaSeparatedComponents: 'expect',
+    },
+  ]
```
juanjoman, line 48: Just have a question, does the project's `package.json` includes chai as a dependency? 😮

juanjoman (changes requested):
Amazing work @Fecony!! I would just like to check that we're providing `chai` as we're importing it on the generated test. Could you check it? 🙏

packages/cli/src/commands/new/test.ts
Fecony: Hey @juanjoman. That is something I mention in PR description "Test files use chai that is not installed by default in new projects." 😄 

I think we should include `chai` as dependency because Booster docs use `chai` but we don't provide it by default? 😅 

+ if `chai` is not in the `project.json` dependencies, what we should use as `expect` assertion instead?

Fecony: @juanjoman I though about importing `expect` from `./expect` if it exists, as we might expect it to be here (because it is [recommended file](http://docs.booster.cloud/chapters/06_testing-booster-applications?id=recommended-files))

If not, check if `chai` exists and import it instead. But if user doesn't have them? 🤔

juanjoman: oooh sorry @Fecony, I didn't read it 😓. I think it's ok then to import it from `./expect` and let the user choose their test library 😄

Fecony:
Closing, as I have no time/knowledge which path to take. 😬 

- We could take the path of creating automated tests, as @ClaytonFarr [did investigate](https://github.com/ClaytonFarr/booster-testing-utilities), that was mentioned on [Discord here](https://discord.com/channels/763753198388510780/781960568541806593/966081220935036988)
- Take simple path and assume that user has `expect.ts`
- Force user to use *[insert testing library here]*
- Create Booster-specific test matchers `expect(...).to.registerEvents(...)`

Issue #656, opened 2021-03-04
Hoppscotch recommendation for subscribing to read models


adriangmweb:
# Documentation bug

## Description

`Hoppscotch`, formerly `Postwoman`, is being recommended in the documentation to subscribe to a read model. But after trying without success and investigating a bit, it seems that it doesn't support sending headers on WebSockets. This doesn't allow connections to use the `Sec-WebSocket-Protocol` to send the `graphql-ws` header. So it's not possible to use it to connect to a Booster subscription.

I think we should remove the [reference in the docs](https://github.com/boostercloud/booster/tree/main/docs#subscribing-to-read-models)

UdattaChowdhury:
Hey!! I would like to work on this issue. I am a beginner so can you help me out where I can find this issue?

adriangmweb:
hey @UdattaChowdhury! Of course, the reference in the docs[ is here](https://github.com/boostercloud/booster/blob/main/docs/chapters/04_features.md#subscribing-to-read-models). I also updated the link on the description.

If you need any help with the issue, don't hesitate to reach me out on Discord and I'll give you a hand!

UdattaChowdhury:
Sure, @adriangmweb. Thanks.

PR #867, opened 2021-06-09
feat: Local Environment Configuration

NickSeagull:
## Description
In this PR, I'm adding a core function called `configureLocal`. This function allows to define Booster configurations used for development purposes that use the local provider. This is specially useful, because the local provider package shouldn't be uploaded together with the production code, so it can be added as a `devDependency`.

The `configureLocal` function will check for the `BOOSTER_ENV` environment variable, and only then require the local provider package.

Together with this change I've made the appropriate changes to the documentation so users can already know how to use it, and unit/integration tests. 🎉 

## Checks
- [x] Project Builds


packages/framework-types/src/booster-app.ts
```diff
@@ -8,6 +8,7 @@ export interface BoosterApp {
   start(projectPath: string): void
   config: BoosterConfig
   configure(environment: string, configurator: (config: BoosterConfig) => void): void
+  configureLocal(environment: string, configurator: (config: BoosterConfig) => void): void
```
juanjoman, line 11: can't we reuse `configure`?

juanjoman (commented):
I'm not 100% sure, but I think we can already do `config.provider = Local.Provider()`. Why's different from the Kubernetes one? https://github.com/boostercloud/booster/pull/867/files#diff-d9e7660b5dae564a1f1a3daec6c8a2f5f8b74d520e4c48750dc5fa218ebbedf7R14

NickSeagull:
@juanjoman : It's different because in the case of the local provider, it will be in `devDependencies`, and not `dependencies` meaning that we cannot import it.

If we reuse configure, what would happen is that when Booster loads `config.ts` it will throw in runtime saying that it cannot find the module

alvaroloes (changes requested):
@NickSeagull I don't think making such a huge special case for a specific provider using a specific typescript package is a good solution to this.

We are hardcoding the local provider into the core. 
What if we create a different local provider that uses other technology?
What if the user wants to add rockets to the local provider (like using the AWS auth or any other thing)

Also, this problem is happening with any other provider. An ideal solution would work for any provider so that when you deploy AWS, the Kubernetes or Azure dependencies are not loaded (and they are right now).

There is an issue where we are discussing this in a way that can be done for any provider (taking also into account the pacakge.json). Do you mind keep discussing there and add your idea?

---

Being said that, I think that the approach you followed here is opening a new path. 

_What if we force the user to add the `require` inside the `configure` block?_ 
We would get exactly the same effect you got here with the local provider, but for any provider.
Maybe this could be a bit awkward at the beginning, but if we say in the docs it is the recommended way, maybe we can live with it.

This, however, still doesn't fix the fact that the dependencies of other providers will be always inside the `node_modules` of the Lambdas (because they are listed in `package.json`). This would need a bit more thinking.

NickSeagull:
Agree with this, let me rework this in a better way

alvaroloes:
@NickSeagull I was editing my comment and, eventually, added a lot of things. Please re-read 🙏 

I'm also linking this to the document where we are discussing this, for reference

NickSeagull:
I'm gonna go with a variation of option one and a mix of what you say regarding the `config.rockets` field. I'm not going to get into the field of pruning unnecessary dependencies, as that is a task on itself.

NickSeagull:
With the latest changes, now we don't need to have a `configureLocal` but rather, we have a `config.providerPackage` field (I've left the `config.provider` field for backwards compatibility, but if the team thinks we should use the new one only, let's remove it).

By using `config.providerPackage`, Booster will only perform a `require` when the `provider` is used (lazy loading).

Also, I've added a `config.rockets` field that will be used to allow passing rockets to the provider.

docs/chapters/04_features.md
```diff
@@ -250,18 +282,20 @@ While it is OK to know how to manually send GraphQL request, you normally don't
 
 To have a great developer experience, we **strongly recommend** to use a GraphQL client for your platform of choice. Here are some great ones:
 
-- **[Hoppscotch (formerly Postwoman)](https://hoppscotch.io/)**: Ideal for testing sending manual requests, getting the schema, etc.
+- **[Altair](https://altair.sirmuel.design/)**: Ideal for testing sending manual requests, getting the schema, etc.
```
NickSeagull, line 285: Switched to Altair, because it is the easiest GraphQL client that allows testing localhost

packages/framework-core/test/booster-graphql-dispatcher.test.ts
```diff
@@ -20,8 +19,7 @@ import { GraphQLWebsocketHandler } from '../src/services/graphql/websocket-proto
 import { ExecutionResult } from 'graphql/execution/execute'
 import { GraphQLError } from 'graphql'
 import { BoosterTokenVerifier } from '../src/booster-token-verifier'
-
-const logger: Logger = console
+import { noopLogger as logger } from './helpers/logger-helper'
```
NickSeagull, line 22: There were tests that leaked logs, I switched their logger to a noop logger

packages/framework-types/src/config.ts
```diff
@@ -102,6 +105,11 @@ export class BoosterConfig {
   }
 
   public get provider(): ProviderLibrary {
+    if (!this._provider && this.providerPackage) {
+      const rockets = this.rockets ?? []
+      const provider = require(this.providerPackage)
+      this._provider = provider.Provider(rockets)
+    }
```
NickSeagull, line 112: Here's where the magic happens. The rest of the codebase stays the same, because when accessing `config.provider` as a getter, this code will get trigger, and if the provider is not set, it will `require` by using the `config.providerPackage` and `config.rockets` fields

docs/chapters/04_features.md
adriangmweb: Does Altair support adding headers on WebSockets? Because then we could also close #656

docs/chapters/04_features.md
```diff
@@ -382,15 +416,15 @@ And we would get the following as response:
...
 
 Doing this process manually is a bit cumbersome. _You will probably never need to do this_, as GraphQL clients like [Apollo](#using-apollo-client) abstract this process away. However, we will explain how to do it for learning purposes.
 
-Before sending any subscription, you need to _connect_ to the WebSocket to open the two-way communication channel. This connection is done differently depending on the client/library you use to manage web sockets. In this section, we will show examples using the [`wscat`](https://github.com/websockets/wscat) command line program. You can also use the online tool [Hoppscotch (formerly Postwoman)](https://hoppscotch.io/)
+Before sending any subscription, you need to _connect_ to the WebSocket to open the two-way communication channel. This connection is done differently depending on the client/library you use to manage web sockets. In this section, we will show examples using the [`wscat`](https://github.com/websockets/wscat) command line program. You can also use the online tool [Altair](https://altair.sirmuel.design/)
```
adriangmweb, line 427: I guess that this answers the question about WebSockets, check my previous comment ;)

docs/chapters/04_features.md
NickSeagull: If by headers you mean connection params, then yes!

davidverdu (approved):
LGTM.

docs/chapters/04_features.md
imolorhe: I don't think the user agents support specifying headers as part of a web socket upgrade request, and that's why the connection params was added as part of the graphql subscriptions over websocket specification, to be able to authenticate the subscription requests AFAIK. I detailed a bit more about it in this issue: https://github.com/imolorhe/altair/issues/1419#issuecomment-759148821

packages/framework-core/test/booster-graphql-dispatcher.test.ts
alvaroloes: Just to consider: Not always is good to silence the logs in the tests. It can help to troubleshoot a failing test, as you can see the logs of the tested code printed out. But I would leave this decision to the maker of each test.

packages/framework-integration-tests/src/config/config.ts
```diff
@@ -5,15 +5,10 @@ import * as Kubernetes from '@boostercloud/framework-provider-kubernetes'
...
-
-  Booster.configure('local', (config: BoosterConfig): void => {
-    config.appName = 'my-store'
-    config.provider = Local.Provider()
-  })
-}
+Booster.configure('local', (config: BoosterConfig): void => {
+  config.appName = 'my-store'
+  config.providerPackage = '@boostercloud/framework-provider-local'
+})
```
alvaroloes, line 11: 💥

packages/framework-types/src/config.ts
alvaroloes: I like it!

docs/chapters/04_features.md
```diff
@@ -1,5 +1,34 @@
...
+})
+```
+
+Note how, contrary to the rest of the environments, the configuration of the local environment is using the `config.providerPackage` field,
+and not the `config.provider`.
+
+This is a special field that allows deferring the loading of the
+provider to later, so you don't get dependencies errors when you
+upload your app to production, given that the local provider lives in
+the `devDependencies` section of your `package.json`
```
alvaroloes, line 22: @NickSeagull 
🤔  I think we should bet on only one option. To be honest, I like your solution of providing the package name and the rockets as separate properties. After thinking about it, I don't see drawbacks. 

So I would write in the documentation only this way of specifying the provider and the rockets. Besides, I would rewrite all the configuration environments we have in the integration tests, etc. to the new way and mark the `provider` property as deprecated with a message.

Do you see any inconvenience in doing this? Any possible design flaw?

alvaroloes (commented):
I love how the change turned out to be. Thanks a lot for your hard work on the docs!

I left one comment related to adopting this change fully.

packages/framework-core/test/booster-graphql-dispatcher.test.ts
NickSeagull: I believe that logging should only be enabled when troubleshooting, and then shutdown again

docs/chapters/04_features.md
NickSeagull: No, my decision was more geared towards backwards compatibility, but I would just get one as well

packages/framework-types/src/config.ts
```diff
@@ -105,6 +105,16 @@ export class BoosterConfig {
...
+    if (this._provider) {
+      console.warn(`
+      The usage of the 'config.provider' field is deprecated,
+      please use 'config.providerPackage' instead.
+
+      For more information, check out the docs:
+
+      https://docs.booster.cloud/chapters/05_going-deeper?id=configuration-and-environments
+      `)
+    }
```
alvaroloes, line 117: Wouldn't this error appear every time we want to get the provider in Booster core?

If I understood this properly, this message should go (without the if) in the setter. Correct?

NickSeagull: Correct! Nice catch, thanks!!

alvaroloes (approved):

PR #868, opened 2021-06-10
include custom object types like enums

adriangmweb:
## Description

Fix to avoid excluding enums as `GraphQLJSONObject` types

## Changes

Now checks when the type at compilation time is just an `object` type, setting then the `GraphQLJSONObject` by default

## Checks
- [x] Project Builds


charlietfe (approved):
I

rdiaz82 (approved):
LGTM

Issue #849, opened 2021-06-03
Create real classes instances of entities, read models and events in the Booster class.


Moneyba:
## Current Behavior
Currently, when you get entities, read models and events using the Booster class, you get a raw javascript object instead of an instance of their respective classes.
https://github.com/boostercloud/booster/blob/6448db061ba7d11bd91bbd6525e4b646fb8205a9/packages/framework-core/src/booster.ts#L70
https://github.com/boostercloud/booster/blob/6448db061ba7d11bd91bbd6525e4b646fb8205a9/packages/framework-core/src/booster.ts#L77
https://github.com/boostercloud/booster/blob/6448db061ba7d11bd91bbd6525e4b646fb8205a9/packages/framework-core/src/booster.ts#L105
This leads to unexpected bugs in the user's code if they define instance methods in their entity, read model and event classes and call them, as those methods won't be found in those raw javascript objects.

## Expected behavior

The methods should be found, as its receiver should be a proper class instance

## Possible Solution

The solution is to create proper instances for entities, read models and events of the Booster class.
It is a similar solution that was made in issue https://github.com/boostercloud/booster/issues/797

## Additional information
Related issue: https://github.com/boostercloud/booster/issues/797

### Environment

- Booster version: This bug is in all current published versions. 
- Node/npm version: Node 16.2.0/npm 7.13.0
- OS: OSX 11.1

PR #875, opened 2021-06-14
fix: create real classes instances of entities, read models and events in the Booster class

Moneyba:
## Description
This PR fixes https://github.com/boostercloud/booster/issues/849

## Changes

- Create proper instances for  entities, read models and events in the Booster class.
- Remove the deprecated method `fetchEntitySnapshot`

## Checks
- [x] Project Builds


packages/framework-core/test/booster.test.ts
```diff
@@ -81,35 +72,145 @@ describe('the `Booster` class', () => {
...
+          } as EventInterface,
+        },
+      ]
+      const providerEventsSearch = fake.returns(searchResult)
+      Booster.configureCurrentEnv((config) => {
+        config.provider = {
+          events: {
+            search: providerEventsSearch,
+          },
+        } as unknown as ProviderLibrary
```
juanjoman, line 130: is this double `as` on purpose? 😮

packages/framework-core/test/booster.test.ts
```diff
@@ -81,35 +72,145 @@ describe('the `Booster` class', () => {
...
 
-    expect(Booster.serveGraphQL).to.have.been.calledOnceWith(message)
+      expect(Booster.serveGraphQL).to.have.been.calledOnceWith(message)
+    })
+  })
+
+  describe('The `entity` method', () => {
+    afterEach(() => {
+      restore()
+    })
```
juanjoman, line 181: This is at the top of the file too, is it necessary?

packages/framework-core/test/booster.test.ts
```diff
@@ -81,35 +72,145 @@ describe('the `Booster` class', () => {
...
+
+    context('given a BoosterConfig', () => {
+      const config = new BoosterConfig('test')
+      config.provider = {} as ProviderLibrary
+
+      it('the `entity` function calls to the `fetchEntitySnapshot` method in the EventStore', async () => {
+        replace(EventStore.prototype, 'fetchEntitySnapshot', fake.returns({ id: '42' }))
```
juanjoman, line 188: Just to confirm, this `fetchEntitySnapshot` is not the one we wanted to deprecate right? Is it the one from providers?

juanjoman (approved):
LGTM!! Just left some comments but nothing big, nice work 🎉

packages/framework-common-helpers/src/instances.ts
```diff
@@ -0,0 +1,15 @@
...
+  Object.assign(instance, rawObject)
+  return instance
+}
+
+export function createInstances<T>(instanceClass: Class<T>, rawObjects: Array<Record<string, any>>): T[] {
+  return rawObjects.map((rawObject) => {
+    const instance = new instanceClass()
+    Object.assign(instance, rawObject)
+    return instance
+  })
```
alvaroloes, line 14: What about reusing the above method?
```suggestion
  return rawObjects.map((rawObject) => createInstance(instanceClass, rawObject))
```

packages/framework-core/test/booster.test.ts
alvaroloes: I think it is the only way to bypass the compiler in this case

packages/framework-core/test/booster.test.ts
```diff
@@ -81,35 +72,141 @@ describe('the `Booster` class', () => {
...
+
+      for (const event of events) {
+        let eventValue
+        switch (event.type) {
+          case TestEvent.name:
+            eventValue = event.value as TestEvent
+            expect(eventValue.getId()).to.not.throw
+            break
+          default:
+            console.log('Default')
```
alvaroloes, line 147: Is this log required?

packages/framework-core/test/booster.test.ts
```diff
@@ -81,35 +72,141 @@ describe('the `Booster` class', () => {
...
+          type: TestEvent.name,
+          entity: random.alpha(),
+          entityID: random.uuid(),
+          createdAt: random.alphaNumeric(),
+          value: {
+            id: '1',
+            entityID: () => UUID.generate(),
+          } as EventInterface,
+        },
+      ]
```
alvaroloes, line 123: Would you mind including another raw event of another type to test that an array of events of different types is correctly handled, please?

alvaroloes (commented):
Left some comments, but the rest LGTM!

packages/framework-core/test/booster.test.ts
Moneyba: Yes, that's it. The other option would be as `any`. Thanks @alvaroloes!

packages/framework-core/test/booster.test.ts
Moneyba: `fetchEntitySnapshot` is a function inside of the class `EventStore`.

alvaroloes (approved):

packages/framework-core/test/booster.test.ts
juanjoman: Awesome! 😄

PR #876, opened 2021-06-17
Commands' before hooks

juanjoman:
**Related PR:** https://github.com/boostercloud/booster/pull/850

**Note:** PR above is the complementary version of this one and will need to be merged first since it's bigger and I'm probably going to have conflicts.

## Description
This is the second part of the https://github.com/boostercloud/booster/issues/807 issue (commands side). With this new feature, we'll introduce a `beforeCommand` parameter, that will allow us to do some extra actions before a command handler is called. This is the type definition for this new function:
```typescript
export interface FilterHooks {
  readonly beforeCommand?: Array<CommandBeforeFunction>
}
export type CommandInput = Record<string, any>
// This one below, the ones above are just informative
export type CommandBeforeFunction = (input: CommandInput, currentUser?: UserEnvelope) => CommandInput
```

## Changes
- Add new types for the before hooks (only for commands)
- Modify the current input type for a more reusable one `export type CommandInput = Record<string, any>`
- Add the before functions logic to command handlers' processing steps

## Checks
- [X] Project Builds


docs/chapters/04_features.md
```diff
@@ -317,6 +317,69 @@ And this would be the response:
 
 > [!NOTE] Remember to set the proper **access token** for secured commands, check ["Authorizing operations"](#authorizing-operations).
 
+### Adding "before" hooks to your commands
+As seen above, when you send commands, you can include input parameters to be received for the command handler. Although, there are some cases where you'd want to do some actions just before the `handle` function gets called in your command, like:
+- Validating parameters
```
juanjoman, line 322: Not sure if we can reuse the docs in the previous PR https://github.com/boostercloud/booster/pull/850 so it's everything in the same chapter

davidverdu: I'm not sure. I'd keep them separated in each section, since the hook arguments are different between ReadModels and Commands

davidverdu (commented):
LGTM except for two things:

- I miss an integration test testing a thrown exception inside a before filter
- AWS integration tests are failing in my machine. @adriangmweb solved the problem in his PR #878, since apparently it was a dependency problem

davidverdu (approved):
LGTM

docs/chapters/04_features.md
NickSeagull: Agree with @davidverdu

docs/chapters/04_features.md
```diff
@@ -317,6 +317,69 @@ And this would be the response:
...
+- Change command inputs on the fly with custom logic
+- Deny access to concrete users
```
NickSeagull, line 324:
```suggestion
- Deny access to specific users
```

docs/chapters/04_features.md
```diff
@@ -317,6 +317,69 @@ And this would be the response:
...
+
+For these cases (and more), you can use the commands before hooks. 
+
+These hooks can be added in your @Command decorator through the `beforeCommand` parameter, as an array of functions:
+```typescript
+@Command({
+  authorize: [User],
+  beforeCommand: [ChangeCartItem.beforeFn],
```
NickSeagull, line 332: Same as in #850 , I'd make this an array of keys

packages/framework-core/src/booster-command-dispatcher.ts
```diff
@@ -31,7 +34,14 @@ export class BoosterCommandDispatcher {
...
+
+    const commandInput = this.applyBeforeFunctions(
+      commandEnvelope.value,
+      commandMetadata.before,
+      commandEnvelope.currentUser
+    )
+
+    const commandInstance = createInstance(commandClass, commandInput)
     // TODO: Here we could call "command.validate()" so that the user can prevalidate
     // the command inputted by the user.
```
NickSeagull, line 46: I believe this comment is not necessary given we have hooks now, no?

packages/framework-core/test/booster-command-dispatcher.test.ts
```diff
@@ -193,5 +195,77 @@ describe('the `BoosterCommandsDispatcher`', () => {
...
+        return input
+      }
+
+      it('transforms the input if a before hook function is passed', async () => {
+        let transformedInput = {}
+        @Command({ authorize: 'all', beforeCommand: [beforeFn] })
+        class PostComment {
+          public constructor(readonly comment: string) {}
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          public static async handle(command: PostComment, _register: Register): Promise<void> {
```
NickSeagull, line 218: I believe you can remove the register arg
```suggestion
          public static async handle(command: PostComment): Promise<void> {
```

packages/framework-core/test/booster-command-dispatcher.test.ts
```diff
@@ -193,5 +195,77 @@ describe('the `BoosterCommandsDispatcher`', () => {
...
+        expect(transformedInput).to.deep.equal(new PostComment(newComment))
+      })
+
+      it('transforms the input when more than one before hook function is passed', async () => {
+        let transformedInput = {}
+        @Command({ authorize: 'all', beforeCommand: [beforeFn, beforeFnV2] })
+        class PostComment {
+          public constructor(readonly comment: string) {}
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          public static async handle(command: PostComment, _register: Register): Promise<void> {
```
NickSeagull, line 247: same as above

NickSeagull (changes requested):
Left a suggestion similar to the one I left at #850

packages/framework-core/src/booster-command-dispatcher.ts
juanjoman: You're right!

docs/chapters/04_features.md
```diff
@@ -317,6 +317,69 @@ And this would be the response:
...
+But this is not all, remember that the `beforeCommand` parameter is an array of functions. That is because these functions are chainable:
+1. Imagine that we have 2 functions: `beforeFn` and `beforeFnV2`
+2. When sending a command, the `beforeFn` above will be called, and we could change the `input` variable (for example)
+3. That changed `input` variable will then be passed to the next function `beforeFnV2`
+
+Let's see it through an example:
+```typescript
+@Command({
+  authorize: [User],
+  beforeCommand: [ChangeCartItem.beforeFn, ChangeCartItem.beforeFnV2],
```
alvaroloes, line 358: Wondering: Why chaning the API here? I mean, in ReadModels we use `before`, but here we need to use `beforeCommand`, which is also redundant. Can't we use just use `before`?

packages/framework-types/src/concepts/filter-hooks.ts
```diff
@@ -0,0 +1,8 @@
+import { UserEnvelope } from '../envelope'
+import { CommandInput } from './command'
+
+export interface FilterHooks {
```
alvaroloes, line 4: As the function this interface represents is tightly coupled with Commands, we can call it `CommandFilterHooks`. Also, we can call the "FilterHooks" we are using in read models as `ReadModelFilterHooks`, as it is only useful for read models (because of the parameters the "before" function receives).

What do you think?

alvaroloes (commented):
Left one question and one suggestion. The main concern is the question (related to the name "beforeCommand")

Thanks for this!

docs/chapters/04_features.md
juanjoman: I just wanted to reuse the same `FilterHooks` interface for both but now looking at your second comment I think It's much clearer to keep it as `before` renaming also the interfaces. Thanks!!

alvaroloes (approved):

PR #878, opened 2021-06-17
GraphQL Pagination on queries

### LINKED ISSUE #835: Move Query filters doc to GraphQL API section
## Description

The GraphQL API new filters are explained in the [booster architecture](https://docs.booster.cloud/#/chapters/03_booster-architecture?id=filtering-a-read-model) section of the docs. The ideal would be to move them to the [GraphQL API](https://docs.booster.cloud/#/chapters/04_features?id=graphql-api) section, so it makes more sense.

As you can see in the following image, on the actual section there is a link to see more information on the GraphQL API section, so it just makes sense to move them there. 
image

adriangmweb:
## Description

This PR includes the GraphQL pagination as it was decided on [the confluence proposal](a list of carts using complex filters). 

- Adds a new Read Model with the `List` prefix which returns a `Connection` type. This type returns a `cursor` and an `items` property.
    - `cursor` includes the object to perform pagination, if present, on the next call. If there are no more items, it will be null.
    - `items` is the array of items returned by the query
    
- The new `List` prefixed Read Model includes two new optional parameters; `limit` and `afterCursor`.
    - `limit` is used to set a maximum number of results to be returned by the query.
    - `afterCursor` can be used when there is pagination, and its value should be the one returned on the property `cursor` on the previous query.
    
- Retro compatibility between the old and the new read models has been ensured to avoid breaking changes.
- Some integration tests have been included to ensure stability
- Includes documentations and moves the read model filters one to the GraphQL section, so it closes #835 

## Checks
- [x] Project Builds


packages/framework-core/src/booster-read-models-reader.ts
```diff
@@ -60,6 +60,10 @@ export class BoosterReadModelsReader {
     if (readModelRequest.filters) {
       searcher.filter(readModelRequest.filters)
     }
+    searcher
+      .limit(readModelRequest.limit)
+      .afterCursor(readModelRequest.afterCursor)
+      .paginatedVersion(readModelRequest.paginatedVersion)
```
juanjoman, line 66: what's the `paginatedVersion` for?

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -78,9 +79,13 @@ export class GraphQLGenerator {
...
     readModelClass: AnyClass
-  ): GraphQLFieldResolver<any, GraphQLResolverContext, { filter: Record<string, ReadModelPropertyFilter> }> {
+  ): GraphQLFieldResolver<any, GraphQLResolverContext, ReadModelRequestArgs> {
     return (parent, args, context, info) => {
-      const readModelEnvelope = toReadModelRequestEnvelope(readModelClass.name, args, context)
+      let isPaginated = false
+      if (info.fieldName && info.fieldName.startsWith('List') && info.returnType.toString().endsWith('Connection')) {
+        isPaginated = true
+      }
+      const readModelEnvelope = toReadModelRequestEnvelope(readModelClass.name, args, context, isPaginated)
```
juanjoman, line 88:
```suggestion
      const isPaginated = info.fieldName && info.fieldName.startsWith('List') && info.returnType.toString().endsWith('Connection'))
      const readModelEnvelope = toReadModelRequestEnvelope(readModelClass.name, args, context, isPaginated)
```

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -145,6 +166,23 @@ export class GraphQLQueryGenerator {
...
   }
+  public generateFilterListedQueriesFields(name: string, type: TargetTypeMetadata): GraphQLFieldConfigArgumentMap {
+    const filterArguments = this.generateFilterArguments(type)
+    const filter: GraphQLInputObjectType = new GraphQLInputObjectType({
+      name: `List${name}Filter`,
+      fields: () => ({
+        ...filterArguments,
+        and: { type: new GraphQLList(filter) },
+        or: { type: new GraphQLList(filter) },
+        not: { type: filter },
```
juanjoman, line 177: if this type is `filter`, is it also `GraphQLInputObjectType`? Where does `filter` come from?

packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
```diff
@@ -136,6 +136,192 @@ describe('Read models end-to-end tests', () => {
...
+              `,
+            })
+          },
+          (result) => result?.data?.CartReadModels?.length >= 1
+        )
+
+        const cartData = queryResult.data.CartReadModels
+
+        expect(cartData).to.be.an('array')
+        expect(cartData.length).to.equal(1)
```
juanjoman, line 239: Shouldn't the check be `result?.data?.CartReadModels?.length === 1`? As we're comparing the same below

packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
```diff
@@ -136,6 +136,192 @@ describe('Read models end-to-end tests', () => {
...
+              `,
+            })
+          },
+          (result) => result?.data?.ListCartReadModels?.items.length >= 1
+        )
+
+        const cartData = queryResult.data.ListCartReadModels.items
+
+        expect(cartData).to.be.an('array')
+        expect(cartData.length).to.equal(1)
```
juanjoman, line 292: same here

packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
```diff
@@ -136,6 +136,192 @@ describe('Read models end-to-end tests', () => {
...
```
juanjoman, line 321: here too

packages/framework-provider-aws/src/library/read-models-searcher-adapter.ts
```diff
@@ -24,13 +34,28 @@ export async function searchReadModel(
       ExpressionAttributeValues: buildExpressionAttributeValues(filters),
     }
   }
+  if (limit) {
+    params = { ...params, Limit: limit }
+  }
+  if (afterCursor) {
+    params = { ...params, ExclusiveStartKey: afterCursor }
+  }
+
```
juanjoman, line 43: can't we add this to the initial assignment like this?
```typescript
let params: DocumentClient.ScanInput = {    
    TableName: config.resourceNames.forReadModel(readModelName),    
    ConsistentRead: true,
    Limit: limit,
    ExclusiveStartKey: afterCursor
}
```

If I recall correctly, these params will be `undefined` anyways if they're not specified

packages/framework-types/src/envelope.ts
```diff
@@ -68,14 +68,31 @@ export interface ReadModelEnvelope {
...
+  items: Array<unknown>
+  count?: number
+  cursor?: Record<string, string>
+}
+
 export interface ReadModelRequestEnvelope extends Envelope {
   typeName: string
   version: number
-  filters?: Record<string, ReadModelPropertyFilter>
+  filters?: ReadModelRequestProperties
```
juanjoman, line 80: Not necessary in this PR, but FYI we found what it could be a bug with this type: https://github.com/boostercloud/booster/issues/879

packages/framework-core/src/services/graphql/graphql-query-generator.ts
davidverdu: It's weird that you are declaring the const filter and using it when you construct class instances inside the assignment

davidverdu (commented):
LGTM except for the comment I left. I also agree with Juanjo's comments. The unit and integration tests run properly in my machine. Let me know when you review the comments to approve the PR

packages/framework-core/src/booster-read-models-reader.ts
adriangmweb: Is just a temporal property to support paginated and not paginated Read Models. In the future, we will deprecate the non-paginated and remove it eventually, so it won't be necessary.

packages/framework-core/src/services/graphql/graphql-generator.ts
adriangmweb: I changed to a cleaner version, but I'd prefer to leave the conditional, so it get easier to recognise for removing it on the future

packages/framework-core/src/services/graphql/graphql-query-generator.ts
adriangmweb: `filter` is the actual const declared being used recursively to build the schema. It's weird but you can see how it's used [here](https://graphql.org/graphql-js/type/#graphqlobjecttype).

packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
adriangmweb: I leave like that to open it to the possibility of adding more cases, but it should work like that as well.

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -78,6 +79,26 @@ export class GraphQLQueryGenerator {
...
+    for (const name in this.targetTypes) {
+      const type = this.targetTypes[name]
+      const graphQLType = this.typeInformer.getGraphQLTypeFor(type.class)
+      queries[`List${inflected.pluralize(name)}`] = {
+        type: new GraphQLObjectType({
+          name: `${type.class.name}Connection`,
+          fields: {
+            items: { type: new GraphQLList(graphQLType) },
+            cursor: { type: GraphQLJSONObject },
+          },
```
davidverdu, line 93: Is the recursive reference to `graphQLType` working even though `fields` is not a function?

davidverdu (approved):
LGTM

packages/framework-types/src/envelope.ts
adriangmweb: I'll take into account for future PRs as it requires some work

packages/framework-provider-aws/src/library/read-models-searcher-adapter.ts
adriangmweb: That's a great suggestion, thanks @juanjoman! I just changed it 🚀

juanjoman (approved):
LGTM! 🙌

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -78,9 +80,13 @@ export class GraphQLGenerator {
 
   public readModelResolverBuilder(
     readModelClass: AnyClass
-  ): GraphQLFieldResolver<any, GraphQLResolverContext, { filter: Record<string, ReadModelPropertyFilter> }> {
+  ): GraphQLFieldResolver<any, GraphQLResolverContext, ReadModelRequestArgs> {
     return (parent, args, context, info) => {
-      const readModelEnvelope = toReadModelRequestEnvelope(readModelClass.name, args, context)
+      let isPaginated = false
+      if (info.fieldName && info.fieldName === `List${pluralize(readModelClass.name)}`) {
```
alvaroloes, line 86: I think you can do just:
```suggestion
      if (info?.fieldName === `List${pluralize(readModelClass.name)}`) {
```

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -78,6 +79,26 @@ export class GraphQLQueryGenerator {
...
   }
 
+  private generateFilterListedQueries(): GraphQLFieldConfigMap<unknown, GraphQLResolverContext> {
+    const queries: GraphQLFieldConfigMap<unknown, GraphQLResolverContext> = {}
+    for (const name in this.targetTypes) {
+      const type = this.targetTypes[name]
+      const graphQLType = this.typeInformer.getGraphQLTypeFor(type.class)
+      queries[`List${inflected.pluralize(name)}`] = {
+        type: new GraphQLObjectType({
+          name: `${type.class.name}Connection`,
```
alvaroloes, line 89: Just wondering because I'm having a hard time understanding what the `Connection` name means here. What does it refer to? Is this a connection to something? 

Wouldn't using "Result" convey a closer meaning to what it is?

alvaroloes: Also, I think you can use just `${name}...` instead of `${type.class.name}`

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -78,6 +79,26 @@ export class GraphQLQueryGenerator {
     return queries
   }
 
+  private generateFilterListedQueries(): GraphQLFieldConfigMap<unknown, GraphQLResolverContext> {
```
alvaroloes, line 82: Nitpicking. What about calling this `generateListedQueries`? It is shorter and we can infer any list can be filtered 😊

packages/framework-core/src/services/graphql/graphql-query-generator.ts
alvaroloes: I was surprised too, but it is how you can define recursive types in the GraphQL schema when using the programmatic API 🤯

packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
```diff
@@ -136,6 +136,192 @@ describe('Read models end-to-end tests', () => {
...
+        expect(cartData[0].id).to.equal(mockCartId)
+      })
+    })
   })
```
alvaroloes, line 325: I'm missing tests using the `limit` parameter and `afterCursor`. I mean, reading several pages and checking it works

packages/framework-types/src/envelope.ts
```diff
@@ -68,14 +68,31 @@ export interface ReadModelEnvelope {
   value: ReadModelInterface
 }
 
+export interface ReadModelResult {
```
alvaroloes, line 71: I think the name `ReadModelListResult` would be more accurate as if not, it seems this is a result of only one read model. 
Not a hard opinion though ;-)

packages/framework-types/src/searcher.ts
```diff
@@ -1,8 +1,15 @@
...
+import { ReadModelResult } from './envelope'
 import { Class } from './typelevel'
 
-export type SearcherFunction<TObject> = (className: string, filters: FilterFor<TObject>) => Promise<Array<any>>
+export type SearcherFunction<TObject> = (
+  className: string,
+  filters: FilterFor<TObject>,
+  limit?: number,
+  afterCursor?: any,
+  paginatedVersion?: boolean
```
alvaroloes, line 11: Would you mind creating a ticket to remove the `paginatedVersion` related code in the next version?

packages/framework-types/src/searcher.ts
```diff
@@ -76,60 +73,17 @@ export class Searcher<TObject> {
...
-  : TType extends string
-  ? EnumToUnion<typeof StringOperations>
-  : TType extends boolean
-  ? EnumToUnion<typeof BooleanOperations>
-  : never
-
-type EnumToUnion<TEnum> = keyof TEnum
-
-// ----------------------------------------------------------------------------------------------------
-
```
alvaroloes, line 132: Thanks!!!!

packages/framework-types/src/searcher.ts
```diff
@@ -11,10 +18,10 @@ export type SearcherFunction<TObject> = (className: string, filters: FilterFor<T
  */
 export class Searcher<TObject> {
   // private offset?: number
-  // private limit?: number
+  private maxItems?: number
```
alvaroloes, line 21: I think I know why you call this `maxItems` and not limit (because there was a conflict with the function limit(), correct?)

If so, there is a patter that we are following in other classes like Booster config: using the same name but with an underscore prefix: `_limit`. 

Let me know if you like it.

alvaroloes (changes requested):
LOVE this change. I can't believe it is finally happening.

I left several comments, but I leave this as "Requested changes" mainly because:
- The comment about missing tests
- Documentation? I know that you say "pending until it is extended to events", but I don't think we need to wait for that. We can create a section of "Filter & pagination" or whatever and talk only about Read models. When we add it to events, we can add it there.

packages/framework-core/src/services/graphql/graphql-query-generator.ts
adriangmweb: Hey @alvaroloes, the idea of using `Connection` is to match the standard GraphQL type used for the result of a paginated query https://graphql.org/learn/pagination/#complete-connection-model

packages/framework-types/src/searcher.ts
adriangmweb: I'm letting you know that I like more that approach and I'm adding the change 😂 thanks!!

packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
adriangmweb: Tests are included now 🚀

adriangmweb:
hey @alvaroloes! Documentation has been added, and the section regarding read models filters has been moved as well to the GraphQL API section. So it closes #835

packages/framework-core/src/services/graphql/graphql-query-generator.ts
alvaroloes: I still don't understand why GraphQL called it Connection. But it has some "upvotes" because it is the standard in GraphQL, I  think it is not a bad idea.

docs/chapters/04_features.md
```diff
@@ -773,6 +773,36 @@ export class GetProductsCount {
 
 
 
+#### Using pagination
+
+The current Booster GraphQL API includes an alternative type for your read models that stands for `List{"your-read-model-name"}`, which includes the functionality to work with pagination.
```
alvaroloes, line 778: I think it is better to say that the `List` queries are the official way to filter read models, and the other way is deprecated and will be deleted in future versions

docs/chapters/04_features.md
```diff
@@ -773,6 +773,36 @@ export class GetProductsCount {
...
+
+The Read Model List type includes some new parameters that can be used on queries:
+ * `limit`; an integer that specifies the maximum number of items to be returned.
+ * `afterCursor`; an object that includes the key of the last retrieved item, so the next query will return the next page of results.
```
alvaroloes, line 782: This is not accurate and it could confuse the user, as the they should not use the key of the read model, but the cursor returned in the previous query. 

The cursor should be opaque, the user should not need to know its internals, as it can change from one provider to another (this is commented in the GraphQL documentation too)

So we should say here that in `afterCursor` we should put the result of what was returned in `cursor` in the previous query (if not null)

docs/chapters/04_features.md
```diff
@@ -773,6 +773,36 @@ export class GetProductsCount {
...
+  ) {
+    id
+    sku
+    availability
+    price
+  }
+}
+```
+
+Besides the parameters, this new type also returns a different result than the regular one. Called `{your-read-model-name}Connection`, it includes the following properties:
```
alvaroloes, line 800: Similar to a previous comment: I would not call the other query the "regular one", as it will disappear. I would try not to mention the other kind of query whatsoever. Just saying on the top that it is deprecated is enough 😉

docs/chapters/04_features.md
```diff
@@ -773,6 +773,36 @@ export class GetProductsCount {
...
+ * `afterCursor`; if there are more results to paginate, it will return the object to pass to the `cursor` parameter on the next query. If there aren't more items to be shown, it will be undefined.
```
alvaroloes, line 801: 🤔 Isn't this inverted? The result returns `cursor`, not "afterCursor", and it should say "...  the object to pass to the `afterCursor` parameter in the next query..."

docs/chapters/04_features.md
```diff
@@ -773,6 +773,36 @@ export class GetProductsCount {
...
+ * `items`; the list of items returned by the query, if there aren't any, it will be an empty list.
+
+_Note: currenty it just works with Read Models on the AWS Provider_
```
alvaroloes, line 804: I wouldn't include this sentence. There are many more things that only work with AWS Provider, but it is not said. I'm saying this to avoid the need to go over all the places where we say this thing when we add this feature to the local provider, for example.

We already say at the beginning of the documentation that AWS is full supported and the others are experimental.

alvaroloes (commented):
Thanks a lot for the tests and documentation!!

I have made some documentation suggestions.

packages/framework-types/src/searcher.ts
```diff
@@ -76,60 +73,17 @@ export class Searcher<TObject> {
...
   public async search(): Promise<Array<TObject>> {
-    const searchResult = await this.searcherFunction(this.objectClass.name, this.filters)
+    const searchResult = await this.searcherFunction(
+      this.objectClass.name,
+      this.filters,
+      this._limit,
+      this._afterCursor,
+      this._paginatedVersion
+    )
     return searchResult as Array<TObject>
```
alvaroloes, line 83: Would this fail if the search was paginated? The search result would be the `ReadModelListResult`, and not an array.

alvaroloes:
One more thing! I think [this comment](https://github.com/boostercloud/booster/pull/878#discussion_r659939296) was missed

packages/framework-types/src/searcher.ts
alvaroloes: Also, the user should get the cursor back so that they can use it in `Searcher.afterCursor`

docs/chapters/04_features.md
adriangmweb: You're right!

packages/framework-types/src/searcher.ts
adriangmweb: Sorry, I missed this comment. Is on #888

alvaroloes (approved):
Thanks a lot for accepting the suggestions. This change is HUGE! 🚀

Issue #597, opened 2021-02-01
Show rocket errors on deploy time


juanjoman:
## Current Behavior

This is an example from the [backup rocket](https://github.com/boostercloud/booster/tree/master/packages/rocket-backup-aws-infrastructure), that will help you to reproduce this bug:

1. This rocket [throws an error](https://github.com/boostercloud/booster/blob/master/packages/rocket-backup-aws-infrastructure/src/backup-stack.ts#L35) if the `backupType` parameter is not 'POINT_IN_TIME' or 'ON_DEMAND'.
2. Set `backupType: 'WHATEVER'` and run `boost deploy -e <your_environment>`
```typescript
Booster.configure('development', (config: BoosterConfig): void => {
  config.appName = 'my-app-dev'
  config.provider = AWS.Provider([
    {
      packageName: '@boostercloud/rocket-backup-aws-infrastructure',
      parameters: {
        backupType: 'WHATEVER',
      },
    },
  ])
})
```
3. Deployment fails but no error is shown, but an "x" icon shows up in the output after deploying the `toolkit` bucket:
```
boost deploy -e development
ℹ boost deploy [development] 🚀
ℹ Bootstraping the following environment: {"name":"Default environment","account":"<id>","region":"eu-west-1"}
 ⏳  Bootstrapping environment aws://<id>/eu-west-1...
my-app-dev-toolkit: creating CloudFormation changeset...
[██████████████████████████████████████████████████████████] (3/3)



 ✅  Environment aws://<id>/eu-west-1 bootstrapped.
ℹ Deploying my-app-dev on environment development
✖
✔ Deploying
✔ Cleaning up deployment files
ℹ Deployment complete!

```

## Expected behavior
1. The rocket error should be shown at deployment time. 
2. [Optional] - The `toolkit` bucket shouldn't be deployed if this fails.

juanjoman:
FYI, this is what's happening (AWS provider example):
1. [cli] - Everything starts here: https://github.com/boostercloud/booster/blob/main/packages/cli/src/commands/deploy.ts#L19
2. [aws-provider-infra] - `index.ts`calls the deploy, which calls `mountStack` from the rocket. The thing is that the thrown exception is captured by this file and sent through the logger: https://github.com/boostercloud/booster/blob/main/packages/framework-provider-aws-infrastructure/src/index.ts#L16
[...]

PR #881, opened 2021-06-25
Show rocket errors on the CLI

juanjoman:
## Description
Related issue: https://github.com/boostercloud/booster/issues/597

Currently, when deploying a booster application, and if a rocket throws an error, the CLI just keeps printing the next steps, when the deployment failed (it just shows an "X"):
```bash
boost deploy -e development
ℹ boost deploy [development] 🚀
ℹ Bootstraping the following environment: {"name":"Default environment","account":"<id>","region":"eu-west-1"}
 ⏳  Bootstrapping environment aws://<id>/eu-west-1...
my-app-dev-toolkit: creating CloudFormation changeset...
[██████████████████████████████████████████████████████████] (3/3)



 ✅  Environment aws://<id>/eu-west-1 bootstrapped.
ℹ Deploying my-app-dev on environment development
✖ <------ IT ONLY SHOWS THIS "X"
✔ Deploying
✔ Cleaning up deployment files
ℹ Deployment complete!
```
This is better explained in the issue above, but it's happening because the `Script.step` function is treating the `logger.error` that it's thrown in the `aws-infrastructure` package as a "success" call in the `tryCatch` statement: https://github.com/boostercloud/booster/blob/main/packages/cli/src/common/script.ts#L56

**The solutions I thought about were:**
1. Just letting the `deploy` and `nuke` methods throw, so the program is interrupted
2. Modify the logger implementation so `logger.error` always throws.

I decided to go for the **1st option** because `logger.error` is used in different places and I preferred to go with the safe approach.

## Changes
- Remove try-catch wrapper on the `aws-infrastructure` package
- Modified test check that these methods throw exceptions

## Checks
- [X] Project Builds
- [] Updated documentation accordingly --- Not needed

### DIFF
--- a/packages/framework-provider-aws-infrastructure/src/index.ts
+++ b/packages/framework-provider-aws-infrastructure/src/index.ts
@@ -9,19 +9,7 @@ export * from './test-helper/aws-test-helper'
 export const Infrastructure = (rocketDescriptors?: RocketDescriptor[]): ProviderInfrastructure => {
   const rockets = rocketDescriptors?.map(loadRocket)
   return {
-    deploy: async (config: BoosterConfig, logger: Logger) => {
-      try {
-        await deploy(config, logger, rockets)
-      } catch (error) {
-        logger.error(error)
-      }
-    },
-    nuke: async (config: BoosterConfig, logger: Logger) => {
-      try {
-        await nuke(config, logger, rockets)
-      } catch (error) {
-        logger.error(error)
-      }
-    },
+    deploy: async (config: BoosterConfig, logger: Logger) => await deploy(config, logger, rockets),
+    nuke: async (config: BoosterConfig, logger: Logger) => await nuke(config, logger, rockets),
   }
 }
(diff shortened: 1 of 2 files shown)


packages/framework-provider-aws-infrastructure/src/index.ts
```diff
@@ -9,19 +9,7 @@ export * from './test-helper/aws-test-helper'
...
-    },
-    nuke: async (config: BoosterConfig, logger: Logger) => {
-      try {
-        await nuke(config, logger, rockets)
-      } catch (error) {
-        logger.error(error)
-      }
-    },
+    deploy: async (config: BoosterConfig, logger: Logger) => await deploy(config, logger, rockets),
+    nuke: async (config: BoosterConfig, logger: Logger) => await nuke(config, logger, rockets),
```
juanjoman, line 13: I'm letting these throw if there's an error on the `deploy` or `nuke` processes. That way, it will be then captured here: https://github.com/boostercloud/booster/blob/main/packages/cli/src/common/script.ts#L58

NickSeagull (approved):
Removing code, and getting more features?!!?

image

packages/framework-provider-aws-infrastructure/src/index.ts
alvaroloes: To be honest, I thought we had already done this change!

alvaroloes (approved):
Thanks!

PR #890, opened 2021-07-04
feat: create nested array fields in GraphQL schema based on metadata

thomas-advantitge:
## Description
Currently, nested array fields are transformed into `JSONObject` after generation of the GraphQL schema:
```GraphQL
type PartyReadModel {
  id: String
  type: String
  addresses: JSONObject
}
```

## Changes
This PR generates proper GraphQL types for nested array fields, similarly as with nested non-array fields:
```GraphQL
type PartyReadModel {
  id: String
  type: String
  addresses: [Address]
}

type Address {
  street: String
  postalCode: String
  city: String
  country: String
}
```

## Checks
- [x] Project Builds
 
## Additional information
* This PR relies on an update of `metadata-booster`: https://github.com/boostercloud/metadata-booster/pull/1.
* This PR is put in draft as these changes have impact on the filtering of read models. Although it doesn't break existing functionality, filtering on nested array fields is not working. A suggestion was done by updating `ReadModelPubSub`, but provider specific implementations still need an update in order to be able to filter properly from the GraphQL API.
* No new tests were written yet.

### DIFF
--- a/packages/framework-core/src/services/graphql/graphql-type-informer.ts
+++ b/packages/framework-core/src/services/graphql/graphql-type-informer.ts
@@ -57,6 +57,11 @@ export class GraphQLTypeInformer {
         fields[prop.name] = {
           type: GraphQLList(graphQLPropType),
         }
+        // TODO: awaiting update of metadata-booster package to access .isArray boolean
+      } else if ((prop.typeInfo as any).isArray) {
+        fields[prop.name] = {
+          type: GraphQLList(this.getGraphQLTypeFor(prop.typeInfo.type)),
+        }
       } else {
         fields[prop.name] = { type: this.getGraphQLTypeFor(prop.typeInfo.type) }
       }
--- a/packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
+++ b/packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
@@ -44,8 +44,12 @@ function filterReadModel(readModel: Record<string, any>, filters?: Record<string
     return true
   }
   for (const filteredProp in filters) {
-    const readModelPropValue = readModel[filteredProp]
-    return filterByOperation(filters[filteredProp], readModelPropValue)
+    if (Array.isArray(readModel[filteredProp])) {
+      return readModel[filteredProp].some((readModelFilteredProp: any) =>
+        filterByOperation(filters[filteredProp], readModelFilteredProp)
+      )
+    }
+    return filterByOperation(filters[filteredProp], readModel[filteredProp])
   }
   return true
 }


packages/framework-core/src/services/graphql/graphql-type-informer.ts
```diff
@@ -57,6 +57,11 @@ export class GraphQLTypeInformer {
         fields[prop.name] = {
           type: GraphQLList(graphQLPropType),
         }
+        // TODO: awaiting update of metadata-booster package to access .isArray boolean
+      } else if ((prop.typeInfo as any).isArray) {
+        fields[prop.name] = {
+          type: GraphQLList(this.getGraphQLTypeFor(prop.typeInfo.type)),
+        }
```
NickSeagull, line 64: If we manage to do the change that @alvaroloes suggested [here](https://github.com/boostercloud/metadata-booster/pull/1#pullrequestreview-698995075) the `if` could be simplified to a simple name check

thomas-advantitge: It seems there was already an array check above the code block I've added. However, it behaves differently the first time a type enters the if block because the metadata isn't generated yet. It causes the generated GraphQL type to be `JSONObject` for the first reference. 

Not sure if the proposed change is the best solution, feedback is welcome!

alvaroloes:
@adriangmweb Do you mind taking a look at the changes here, please?

adriangmweb (approved):
LGTM!

Thanks, @thomas-advantitge, these kinds of changes are really useful as the more types get registered on the GraphQL Schema, the more secure and robust is the API.

packages/framework-core/src/services/graphql/graphql-type-informer.ts
```diff
@@ -47,11 +47,12 @@ export class GraphQLTypeInformer {
 
       if (primitiveType === Array) {
         const param = prop.typeInfo.parameters[0]
-        const graphQLPropType = this.getGraphQLTypeFor(param.type)
+        let graphQLPropType = this.getGraphQLTypeFor(param.type)
 
         if (!this.isGraphQLScalarType(graphQLPropType) && param.type) {
           const properties = getPropertiesMetadata(param.type)
           this.generateGraphQLTypeFromMetadata({ class: param.type, properties })
+          graphQLPropType = this.getGraphQLTypeFor(param.type)
```
javiertoledo, line 55: Isn't this line (55) doing the same as line 50? Is this call be redundant maybe?

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
```diff
@@ -44,8 +44,12 @@ function filterReadModel(readModel: Record<string, any>, filters?: Record<string
...
   }
   for (const filteredProp in filters) {
-    const readModelPropValue = readModel[filteredProp]
-    return filterByOperation(filters[filteredProp], readModelPropValue)
+    if (Array.isArray(readModel[filteredProp])) {
+      return readModel[filteredProp].some((readModelFilteredProp: any) =>
+        filterByOperation(filters[filteredProp], readModelFilteredProp)
+      )
+    }
+    return filterByOperation(filters[filteredProp], readModel[filteredProp])
```
javiertoledo, line 52: It would be nice to add a couple of tests for this new scenario...

javiertoledo (changes requested):
Thanks a lot for the change @thomas-advantitge. Please check my comments and let me know what you think ;-)

packages/framework-core/src/services/graphql/graphql-type-informer.ts
thomas-advantitge: It's already a while ago but this "duplication" was on purpose. If I'm correct the main problem is `generateGraphQLTypeFromMetadata` which adds information into a map, information which is needed by `getGraphQLTypeFor` to generate the correct type. Without the second `getGraphQLTypeFor`, some types result in JSONObject. 

It's not the most elegant solution, it's rather a fix as larger restructuring will probably take place related to https://github.com/boostercloud/metadata-booster/pull/2.

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
thomas-advantitge: Good point. Allow me to remove this rather unrelated part from the PR due to the lack of time to extend this part right now, at least if the main fix (`graphQLPropType = this.getGraphQLTypeFor(param.type)`) would still be useful to merge.

packages/framework-core/src/services/graphql/graphql-type-informer.ts
javiertoledo: Oh, got it, then this is definitely calling for restructuring, having methods mutating stuff gives me goosebumps...  Thanks for the explanation!

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
javiertoledo: As an alternative, what if you create an issue to implement these tests? Maybe there's someone in my team that can address them...

javiertoledo (approved):
Approving after clarifications.

thomas-advantitge:
Fixed in https://github.com/boostercloud/booster/pull/1071

PR #895, opened 2021-07-15
Allow the local provider to use all the authorization features

alvaroloes:
## Description
With this PR you can now use authorization in your local provider. You can use any auth rocket for that.

## Changes
The local provider was ignoring the token that was being sent in the request. After passing it through the core, all the authorization features are available.

## Checks
- [x] Project Builds

### DIFF
--- a/packages/application-tester/src/index.ts
+++ b/packages/application-tester/src/index.ts
@@ -1,3 +1,4 @@
 export * from './application-tester'
 export * from './provider-test-helper'
 export * from './graphql-helper'
+export * from './token-helper'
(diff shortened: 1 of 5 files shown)


packages/application-tester/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './application-tester'
 export * from './provider-test-helper'
 export * from './graphql-helper'
+export * from './token-helper'
```
alvaroloes, line 4: This was just missing

NickSeagull (approved):
Thanks!!

juanjoman (approved):

PR #897, opened 2021-07-15
Enable CORS in local provider

alvaroloes:
## Description
Using the local provider from a web browser was giving CORS errors. This PR fixes that

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-provider-local-infrastructure/src/index.ts
+++ b/packages/framework-provider-local-infrastructure/src/index.ts
@@ -43,7 +43,7 @@ export const Infrastructure = {
     expressServer.use(express.json())
     expressServer.use(cors())
     expressServer.use(function (req, res, next) {
-      res.header('Access-Control-Allow-Origin', '*')
+      res.header('Access-Control-Allow-Origin', '127.0.0.1')
       res.header('Access-Control-Allow-Headers', 'Origin, X-Requested-With, Content-Type, Accept')
       next()
     })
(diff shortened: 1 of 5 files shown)


NickSeagull (approved):
💣

packages/framework-provider-local-infrastructure/src/index.ts
```diff
@@ -40,6 +41,12 @@ export const Infrastructure = {
     const graphQLService = new GraphQLService(userProject)
     router.use('/graphql', new GraphQLController(graphQLService).router)
     expressServer.use(express.json())
+    expressServer.use(cors())
+    expressServer.use(function (req, res, next) {
+      res.header('Access-Control-Allow-Origin', '*')
```
juanjoman, line 46:
```suggestion
      res.header('Access-Control-Allow-Origin', '127.0.0.1')
```
Wouldn't it be better to just allow your machine to do this?

juanjoman (approved):
Amazing! Just left a comment/suggestion

packages/framework-provider-local-infrastructure/src/index.ts
alvaroloes: Never thought of this! But one question: If I write `'*'`, Can another machine access it? I mean, that other machine would be in another network and the developer should allow incoming requests through a specific port...

juanjoman: Only if you port forward the application on local (which is not done by default) or if you expose a container with the local env on a docker for example (as a normal application, exposed to the internet). It's not a super big deal, but something nitpicky haha

alvaroloes: Ok, I'll change it and we will see if in the future we need some configuration about this

alvaroloes: Eventually this didn't worked and I went back to to the `*`

PR #912, opened 2021-07-27
Fix ambiguous dependencies

thomas-advantitge:
When using yarn 2 with PnP support, dependencies are treated more strictly. 
It exposes dependencies being used at runtime but not specified in the package.json manifest.


packages/framework-core/package.json
```diff
@@ -55,6 +55,7 @@
     "faker": "5.1.0",
     "metadata-booster": "0.3.1",
     "mock-jwks": "1.0.0",
+    "nock": "11.8.2",
```
juanjoman, line 58: why do we need `nock` here?

juanjoman (approved):
LGTM! Just left a question below

packages/framework-core/package.json
thomas-advantitge: It's a peer dependency of `mock-jwks`: https://github.com/Levino/mock-jwks/blob/main/package.json
(Sadly, the package also brings a lot of dev dependencies like Koa which seem not even be used.)

rdoria1 (approved):
LGTM

juanjoman:
Hey @thomas-advantitge, while trying to merge your branch I received this error. 

image



Are you able to run the `scripts/check-all-the-things.sh` script? I'm facing an error on your branch

thomas-advantitge:
> lerna run integration --stream

Hi @juanjoman, could not reproduce the error: 
image

Could you give me some more input if you think this is really related?
Running integration tests right now.

juanjoman:
Hmm, could you rebase with `main` so I can check again? 🤔  @thomas-advantitge

Issue #903, opened 2021-07-27
Nested queries and relation DSL for Read Models


javiertoledo:
## Description

GraphQL allows expressing relations by nesting types like this:

```graphql
type User {
  id: ID!
  name: String!
  chatGroups: [ChatGroup] // A user can be in zero or more chat groups
}

type ChatGroup {
  id: ID!
  name: String!
  users: [User] // A chat group can have many users
  category: Category // A chat group can be nested in a category
}

type Category {
  id: ID!
  name: String!
}
```

It would be a nice addition to support this kind of semantics while defining read models. Also, as we're currently using DynamoDB as the read models database in AWS, even making the required queries manually is not very performant.

## Possible Solution

Implement some kind of relation-aware decorators in read models to define one-to-one, one-to-many, and many-to-many relationships. For example, using property decorators:

```typescript
@ReadModel
export class User {
  @hasMany
  public chatGroups: Array<ChatGroup>

  public constructor(readonly id: UUID, readonly name: string) {}

  @Projects(...)
  public static projectionName(...) { ... }
}
```

At this moment, I'm not sure if we could avoid the extra decorators inferring relations at compile time when we're using another read model class as a type, but it's definitely an idea to consider.

Also, it would be interesting that relations are properly supported by the underlying database so nested queries are performant too.

AdrianLorenzoDev:
I love the proposed solution. I think that the interface is easy to understand and simple.

I want to raise some related concerns that I think should be taken into account soon or later:

* With some providers, like AWS, **there is no limitation when creating Global Secondary Indexes (GSI) when you create the stack for the first time using CloudFormation**. However, if you **update an already created stack, you can only make one GSI update (creation/deletion) per template update**. It would be nice to have a way of doing those updates in the future without that kind of compromise, maybe by doing partial updates.
* I think that **Booster should also support nested queries pagination**, because there are a lot of use cases that would benefit from it when using the `@HasMany({referenceKey: <fieldName>})` decorator.

alvaroloes:
Great DSL Javi. Some things to take into consideration in case they help:

### Regarding `@HasMany`

[...]

PR #913, opened 2021-07-28
Nested queries and Read Models relation DSL

javiertoledo:
This PR adds a set of decorators to define relationships between read models to enable nested queries. You can read more details in #903.


javiertoledo:
Closing, as there's an open discussion about this in #903

PR #916, opened 2021-07-29
feat: Return stuff from command handlers

NickSeagull:
## Description

Allows returning stuff from command handlers 

## Changes

* Added an optional `@Returns` decorator that allows specifying the return type of a command handler
* Wired the type into the GraphQL Mutation Generator
* Got result back in core so we can return it through the resolver

## Checks
- [x] Project Builds
 
## Additional Information

Here are some screenshots from tests that I've made:

Usage:
image

Return:
image


.eslintrc.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'prettier',
     'plugin:prettier/recommended', // Enables eslint-plugin-prettier and eslint-config-prettier. This will display prettier errors as ESLint errors. Make sure this is always the last configuration in the extends array.
-    'plugin:import/warnings'
+    'plugin:import/warnings',
```
AdrianLorenzoDev, line 12: **Minor detail**: Is this trailing comma needed?

.eslintrc.js
```diff
@@ -45,6 +44,6 @@ module.exports = {
         allowHigherOrderFunctions: true,
       },
     ],
-    "import/no-extraneous-dependencies": ["error"]
+    'import/no-extraneous-dependencies': ['error'],
```
AdrianLorenzoDev, line 47: **Minor detail**: same here about trailing comma

AdrianLorenzoDev (approved):
LGTM! Simple but awesome solution. Just added comments to minor details, can be ignored!

.eslintrc.js
NickSeagull: Yup it's an ESLint rule, I didn't add it, it was added automatically by it

juanjoman (changes requested):
Love it! I'm just missing tests 🙏

package.json
```diff
@@ -6,11 +6,14 @@
...
   "private": true,
   "scripts": {
+    "build": "lerna run compile",
+    "clean-compile": "lerna run clean && lerna clean --yes && lerna bootstrap && lerna run compile",
     "update-tocs": "npm run update-documentation-toc && npm run update-contributing-toc",
     "update-documentation-toc": "markdown-toc -i --maxdepth 4 docs/README.md",
     "update-contributing-toc": "markdown-toc -i --bullets=- --maxdepth 4 CONTRIBUTING.md"
   },
   "devDependencies": {
+    "@boostercloud/framework-provider-local": "file:packages/framework-provider-local",
```
javiertoledo, line 16: Why do you need to add this to `devDependencies`?

packages/framework-core/src/decorators/command.ts
```diff
@@ -17,12 +18,49 @@ export function Command(
...
+  commandClass: Class<TCommand>,
+  methodName: string,
+  methodDescriptor: TypedPropertyDescriptor<
+    (command: TCommand, register: Register) => Promise<PrimitiveTypeOf<TReturn>>
+  >
+) => void {
+  return (commandClass) => {
+    Booster.configureCurrentEnv((config): void => {
+      if (config.commandHandlerReturnTypes[commandClass.name]) {
+        throw new Error(`A command handler for the command ${commandClass.name} is already registered`)
```
javiertoledo, line 52:
```suggestion
        throw new Error(`A command handler return type for the command ${commandClass.name} is already registered`)
```

javiertoledo (approved):
It's a pity that there's not a way to get the return type automatically here. I wonder if it would be worth redesigning the handler as an instance method to be able to use `Reflect.getMetadata`...

package.json
NickSeagull: Thanks, this was a test

NickSeagull:
Will add tests later

Issue #921, opened 2021-08-03
Add support for a `@sortedBy` constructor parameter annotation in Read Models to set a sort key in DynamoDB


javiertoledo:
# Feature Request

We’ll add support for an optional `@sortedBy` decorator in Booster to add a sort key to the underlying database table. It would work like this:

```typescript
@ReadModel
export class SomeReadModel {
  public constructor(
    readonly id: UUID
    @sortedBy readonly timestamp: string
  ) {}
}
```

With this in place, the framework will set a compound key where the partition key is the read model id, as usual, but we add a sortkey that can be a timestamp or other values to sort the table. This way we can get performant and paginated requests in an order that has a real-world sematic (Currently, paginated requests are sorted by ID, which can yield funny results if you don't take that into account while projecting your entities).

This compound key also has an interesting side effect: it allows efficient read model filtering for nested objects. For example, if you want to represent a group of people, but you want to query by the group id, this makes it possible to get a list of people in the group by setting the ID of the group as the partition key and the ID of the person as part of the sort key.

PR #922, opened 2021-08-04
Add support for a new `@sequencedBy` decorator to allow consuming Read Models as time series

javiertoledo:
## Description

This pull request solves #921 for the AWS provider by adding a `@sequencedBy` parameter decorator to read model constructors. This decorator activates time-series queries and adds as a sortKey to the underlying DynamoDB table for better query performance.

## Changes

- [ ] Created the `@sequencedBy` decorator, which injects metadata in the `Booster.config` object.
- [ ] The AWS infrastructure package picks the metadata and generates the proper sortKey parameter in the table configuration
- [ ] Change the GraphQL queries to allow sort operations and getting a list of read models that share a common primary key
- [ ] Implement integration tests for the new feature

## Checks
- [ ] Project Builds


alvaroloes:
One question @javiertoledo 
Wouldn't it be better to define de `@Index` decorator, instead of sort by, so that it can be used also to do efficient queries, not only sort by?
This way, the user can create several indexes and do queries to those indexes, sorting by them if they need it. I mean, calling it "sortBy" makes the user think it is only for "sorting", but it is an index, so it can also be used to do efficient queries without scanning the whole table. What is more, we have plans to stop using DynamoDB for ReadModels and use something like a relational DB or ElasticSearch. In those technologies, the indexing capabilities are much more flexible.

Another point: I think it also makes sense to allow sorting _in the query_. I mean: we now have the fields `afterCursor` and `pageSize` (I think). We can also add the `sortBy` parameter, so that the user can specify the "index" he wants to sort by.

javiertoledo:
Reworked in PR #924

PR #924, opened 2021-08-06
Add support for a new @sequencedBy decorator to allow consuming Read Models as time series

javiertoledo:
## Description

This pull request solves #921 for the AWS provider by adding a `@sequencedBy` parameter decorator to read model constructors. This decorator activates time-series queries and adds as a sortKey to the underlying DynamoDB table for better query performance.

## Changes

- [ ] Created the `@sequencedBy` decorator, which injects metadata in the `Booster.config` object.
- [ ] The AWS infrastructure package picks the metadata and generates the proper sortKey parameter in the table configuration
- [ ] Change the GraphQL queries to allow sort operations and getting a list of read models that share a common primary key
- [ ] Implement integration tests for the new feature

## Checks
- [ ] Project Builds


packages/framework-core/src/booster-read-models-reader.ts
```diff
@@ -18,9 +20,39 @@ import { getReadModelFilters } from './services/filter-helpers'
...
+    const filters = getReadModelFilters({}, readModelMetadata.before, readModelByIdRequestEnvelope.currentUser)
+
+    searcher.filter(filters)
+
+    return searcher.findById(readModelByIdRequestEnvelope.id, readModelByIdRequestEnvelope.sequenceKey)
+  }
+
+  public async search(
+    readModelRequest: ReadModelRequestEnvelope
+  ): Promise<ReadModelInterface[] | ReadModelListResult<ReadModelInterface>> {
```
NickSeagull, line 37: Nitpick, usually we use the following form of array type throughout the codebase:
```suggestion
  ): Promise<Array<ReadModelInterface> | ReadModelListResult<ReadModelInterface>> {
```

packages/framework-core/src/booster.ts
```diff
@@ -93,12 +95,29 @@ export class Booster {
       if (!Array.isArray(searchResult)) {
         return {
           ...searchResult,
-          items: searchResult ? createInstances(readModelClass, searchResult.items as Array<any>) : [],
+          items: createInstances(readModelClass, searchResult.items),
         }
+      } else {
+        return createInstances(readModelClass, searchResult)
```
NickSeagull, line 101: Maybe we can omit the `else` here?

packages/framework-core/src/decorators/metadata.ts
```diff
@@ -10,3 +10,33 @@ export function getPropertiesMetadata(classType: AnyClass): Array<PropertyMetada
...
+ * borrowed from the promisify-node code and can be found in the following link:
+ * https://github.com/nodegit/promisify-node/blob/02fc47cfc00146a533193bc4740e2e3e3be81c6f/utils/args.js
+ *
+ * @param {Function} func - The function to parse.
+ * @returns {Array} arg - List of arguments in the function.
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function getFunctionArguments(func: Function): Array<string> {
+  // First match everything inside the function argument parens.
+  const args = func.toString().match(/([^(])*\(([^)]*)\)/)?.[2]
```
NickSeagull, line 27: I would replace the regex code for a parser with [`parser-ts`](https://github.com/gcanti/parser-ts) but we can leave this for a future iteration

packages/framework-core/src/decorators/sequenced-by.ts
```diff
@@ -0,0 +1,21 @@
+import { Booster } from '../booster'
+import { Class, ReadModelInterface } from '@boostercloud/framework-types'
+import { getFunctionArguments } from './metadata'
+
+export function sequencedBy(klass: Class<ReadModelInterface>, _functionName: string, parameterIndex: number): void {
```
NickSeagull, line 5: Maybe this should be  `@SequencedBy`? For some reason camelCase looks weird to me here...

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -144,6 +153,28 @@ export class GraphQLGenerator {
...
+  args: ReadModelByIdRequestArgs,
+  context: GraphQLResolverContext,
+  sequenceKeyName?: string
+): ReadModelByIdRequestEnvelope {
+  const sequenceKey = sequenceKeyName
+    ? {
+        name: sequenceKeyName,
+        value: args[sequenceKeyName] as TimeKey,
+      }
+    : undefined
```
NickSeagull, line 167: It could be possible that `args[sequenceKeyName]` is `undefined` here. Can you add a check please?

```suggestion
  const sequenceValue = sequenceKeyName ? args[sequenceKeyName] : undefined
  const sequenceKey = sequenceValue
    ? {
        name: sequenceKeyName,
        value: sequenceValue as TimeKey,
      }
    : undefined
```

NickSeagull: Now that I think of it, we're parsing `unknown` values into Booster internal structures everywhere. Perhaps it is time to use a package that does this for us like [`morphic-ts`](https://github.com/sledorze/morphic-ts) ?

packages/framework-core/src/services/graphql/graphql-query-generator.ts
```diff
@@ -310,3 +313,16 @@ export class GraphQLQueryGenerator {
...
+  if (sequenceKeyName) {
+    return {
+      id: { type: new GraphQLNonNull(GraphQLID) },
+      sequenceKeyName: { type: new GraphQLNonNull(GraphQLID) },
+    }
+  } else {
+    return {
+      id: { type: new GraphQLNonNull(GraphQLID) },
+    }
+  }
```
NickSeagull, line 327: Maybe we can transform this into an early return?

packages/framework-core/src/services/read-model-store.ts
```diff
@@ -39,38 +40,61 @@ export class ReadModelStore {
...
+    entity: EntityInterface,
+    projectionMetadata: ProjectionMetadata
+  ): SequenceKey | undefined {
+    const sequenceKeyName = this.config.readModelSequenceKeys[projectionMetadata.class.name]
+    const sequenceKeyValue = (entity as any)[sequenceKeyName]
+    if (sequenceKeyName && sequenceKeyValue) {
+      return { name: sequenceKeyName, value: sequenceKeyValue }
+    } else {
+      return undefined
+    }
```
NickSeagull, line 87: No need for this `else`
```suggestion
    }
```

packages/framework-core/test/services/graphql/graphql-generator.test.ts
```diff
@@ -186,43 +186,33 @@ describe('GraphQL generator', () => {
       })
     })
 
-    describe('readModelByIDResolverBuilder', () => {
-      let mockReadModels: Array<ReadModelInterface>
+    // describe('readModelByIDResolverBuilder', () => {
```
NickSeagull, line 189: I guess this commented code is fixed in #936

packages/framework-provider-aws/src/library/read-models-adapter.ts
```diff
@@ -59,7 +62,7 @@ export async function storeReadModel(
         },
       })
       .promise()
-  } catch (e) {
+  } catch (e: any) {
```
NickSeagull, line 65: Since when this is possible? In theory TS syntax doesn't allow type annotations in `catch` clauses

packages/framework-provider-aws/src/library/read-models-searcher-adapter.ts
```diff
@@ -21,7 +21,7 @@ export async function searchReadModel(
   limit?: number,
   afterCursor?: DynamoDB.DocumentClient.Key | undefined,
   paginatedVersion = false
-): Promise<Array<any> | ReadModelListResult> {
+): Promise<Array<any> | ReadModelListResult<any>> {
```
NickSeagull, line 24: Maybe better `unknown` for both?

packages/framework-types/src/envelope.ts
```diff
@@ -1,11 +1,12 @@
...
+import { CommandInput, EntityInterface, EventInterface, ReadModelInterface, SequenceKey, UUID } from './concepts'
 import { GraphQLClientMessage } from './graphql-websocket-messages'
 import { FilterFor } from './searcher'
 
 /**
  * An `Envelope` carries a command/event body together with the name
  * of its class. This is important information for the `Distributor` to
- * work. Each provider has to implement their own `Envelope`.
+ * work. Each provider haimport { SequenceKey } from './concepts/sequence-metadata';
+s to implement their own `Envelope`.
```
NickSeagull, line 9:
```suggestion
 * work. Each provider has to implement their own `Envelope`.
```

packages/framework-types/src/envelope.ts
```diff
@@ -83,12 +84,24 @@ export interface ReadModelRequestEnvelope extends Envelope {
...
+
 export interface ReadModelRequestArgs {
   filter?: ReadModelRequestProperties
   limit?: number
   afterCursor?: unknown
 }
 
+export interface ReadModelByIdRequestArgs {
+  id: string
+  [sequenceKey: string]: string | undefined
```
NickSeagull, line 102: This is a [index signature](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures), not a declaration for a single field. Not sure if this is what you meant, this means:

> ReadModelByIdRequestArgs has multiple fields that have a `string` name, and will have of value `string | undefined`

Doesn't matter too much as it works, but take into account that this would accept any kind of field, even multiple ones

NickSeagull (approved):
Left some suggestions, but overall looking good

PR #925, opened 2021-08-07
Simplifications in the GraphQLGenerator class

javiertoledo:
This is the first change of a series of changes to redesign the GraphQL generation service as a more functional module. The current implementation is becoming hard to understand because we're starting to mix the implementations of different kinds of GraphQL queries and overusig injections. We should aim for a more modular design in which functionality is properly isolated and untangled and root classes don't have knowledge of implementation details of leaf classes. For instance, if we differentiate between regular or paginated queries, it would be desirable that we have separate generators for each. This will make it easier to change, improve and test these classes. This first change cleans a small amount of redundancy from the `GraphQLGenerator` class by avoiding having an explicit singleton instance. In upcoming changes, I would like to make the `GraphQLGenerator` root class less aware of what the other classes are doing.

## Checks
- [x] Project Builds
~- [ ] Updated documentation accordingly~ (There are no API changes in this PR)


packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -26,7 +26,7 @@ export class BoosterGraphQLDispatcher {
 
   public constructor(private config: BoosterConfig, private logger: Logger) {
     this.readModelDispatcher = new BoosterReadModelsReader(config, logger)
-    this.graphQLSchema = GraphQLGenerator.build(config, logger).generateSchema()
+    this.graphQLSchema = GraphQLGenerator.generateSchema(config, logger)
```
javiertoledo, line 29: I think that in future changes we should move the logger to the Booster singleton so it becomes accessible as `Booster.logger`. Once that's done we could avoid the need of injecting these instances all over the place, as the `Booster` instance is globally accessible.

NickSeagull (approved):
Awesome, thanks!

PR #935, opened 2021-08-23
Allow getting multiple and single items in a sequence

NickSeagull:
## Description

**BREAKING CHANGE**: Now the "by id" query returns a non empty list of elements, instead of a single one.

This PR adds the possibility of querying Read Models that have the `@sequencedBy` decorator used in their fields.

For example, for this read model:

```ts
@ReadModel({
  authorize: 'all',
})
export class ProductReadModel {
  public constructor(
    readonly id: UUID,
    @sequencedBy readonly sku: SKU,     // <-- Note this line
    readonly displayName: string,
    readonly description: string,
    readonly availability: number,
    public deleted: boolean,
    readonly price?: Money
  ) {}

  // ... projections and so on...

}
```

And three products created, with ID `shir`, each one with an SKU `001`, `002` and `003`.

One can query by the ID like:

```graphql
query {
  ProductReadModel(id: "shir"){
    id
    sku
    displayName
  }
}
```

And get the items in that sequence:

```json
{
  "data": {
    "ProductReadModel": [
      {
        "id": "shir",
        "sku": "001",
        "displayName": "TShirt"
      },
      {
        "id": "shir",
        "sku": "002",
        "displayName": "TShirt"
      },
      {
        "id": "shir",
        "sku": "003",
        "displayName": "TShirt"
      }
    ]
  }
}
```

But also, one can provide a sku to get a single item in the sequence:

```graphql
query {
  ProductReadModel(id: "shir", sku: "001"){
    id
    sku
    displayName
  }
}
```

Which returns the item:

```json
{
  "data": {
    "ProductReadModel": [
      {
        "id": "shir",
        "sku": "001",
        "displayName": "TShirt"
      }
    ]
  }
}
```

## Changes

1. Modified the GraphQL generator in order to allow returning a query
2. Changed the `getItem` call to DynamoDB in the AWS provider to `query`

## Checks
- [x] Project Builds
 
## Additional information

Will work on tests and docs in the following PR as this one isn't going to be merged into `main`


javiertoledo:
Before applying this breaking change and thinking twice from a final-user experience standpoint, I think that it would make sense to differentiate two scenarios and keep backward compatibility for non-sequenced read models.

I mean, for non-sequenced read models, we generate the regular API that yields a single read model instance, as usual, but when we add the `@sequencedBy` decorator to a read model, it becomes a "sequenced read model" and both the request by ID and the request by ID + sequenceKey yield the proposed Array-like response.
 
Some arguments in favor of doing this are:

1. Obviously backward compatibility. For users not using the sequences feature, it's going to feel weird having to change all their API clients for no apparent benefit. It's also more semantic IMO.
2. We know at deploy time whether a specific read model is sequenced or not because we can check the metadata anytime.
3. The user will also know and can easily check when a read model is sequenced as they can clearly see the `@sequencedBy` decorator.
4. It's understandable that a query by ID of a sequenced read model yields an array of objects (the list of objects in the sequence)
5. It doesn't make that much sense that requesting a specific object yields an array with a single object in it.
6. When you're implementing a front-end app, it's likely that you know that the model is sequenced and you will act accordingly, and a model will rarely change from being sequenced to not being sequenced, and if it does, it will probably be a reasoned decision.
7. When generating the API types, we can easily differentiate the sequenced vs non-sequenced scenarios and generate different types. The front-end teams will have a consistent type-checked API in any case.

packages/framework-provider-aws/src/library/read-models-adapter.ts
```diff
@@ -29,18 +30,36 @@ export async function fetchReadModel(
   readModelName: string,
   readModelID: UUID,
   sequenceKey?: SequenceKey
-): Promise<ReadModelInterface> {
-  const params: DynamoDB.DocumentClient.GetItemInput = {
+): Promise<ReadOnlyNonEmptyArray<ReadModelInterface>> {
+  // We're using query for get single read models too as the difference in performance
+  // between get-item and query is none for a single item.
+  // Source: https://forums.aws.amazon.com/thread.jspa?threadID=93743
```
javiertoledo, line 36: Interesting... 🤔

javiertoledo (approved):
This looks great! Did you use some kind of import-reorganizing plugin?

NickSeagull:
Agree with all the points @javiertoledo , I'm gonna rework the API generation favoring backwards compatibility as it makes a lot of sense!

javiertoledo:
Merging this to the main branch in #924

PR #936, opened 2021-08-23
Unit tests for findById queries in the new `@sequencedBy` decorator feature

javiertoledo:
This PR implements the unit tests for the `findById` feature implemented in PR #924 for non-sequenced and sequenced read model requests. It also improves the overall quality of some tests on that module.

## Checks
- [x] Project Builds
- ~[ ] Updated documentation accordingly~ _Not needed_
 
## Additional information

This is also related to the work in PR #935 which extends the `findById` method to query sequences.


packages/framework-types/src/searcher.ts
```diff
@@ -77,8 +77,14 @@ export class Searcher<TObject> {
...
   public async searchOne(): Promise<TObject> {
     // TODO: If there is only an ID filter with one value, this should call to `findById`
-    const searchResult = await this.search()
-    return Array.isArray(searchResult) ? searchResult[0] : searchResult.items[0]
+    const searchResult = await this.searcherFunction(
+      this.objectClass.name,
+      this.filters,
+      1, // Forces limit 1
+      this._afterCursor,
+      false // It doesn't make sense to paginate a single result, as pagination metadata would be discarded
```
NickSeagull, line 85: Now thinking about it, if we use `getByID` to get sequences as well, pagination might be requried

NickSeagull (approved):

packages/framework-types/src/searcher.ts
javiertoledo: Sequences are naturally paginated if we support filtering the sequence by sequence keys. If you have a list of items and you need the next "page" you only need to request objects where `sequenceKey` > `biggestKnownSequenceKey`

PR #937, opened 2021-08-24
fix: fix graphql type recognition for complex nested fields

karel-advantitge:
This is a fix related https://github.com/boostercloud/booster/pull/890 but for non-array fields. It causes nested fields to have generated types rather than using the fallback JSONObject. Note that the type informer class could potentially be simplified in order to have a cleaner solution than this.


javiertoledo:
Thanks @karel-advantitge! I've been working a bit with the GraphQL module lately and it's definitely needing a refactor, we'll probably address that soon :-)

javiertoledo (approved):

karel-advantitge:
Fixed in https://github.com/boostercloud/booster/pull/1071

PR #942, opened 2021-08-29
fix(core): Ensure Booster.entity returns an entity instead of envelope

### LINKED ISSUE #896: Booster.entity returns an entity envelope and not the entity itself
# Bug Report

## Current Behavior

In a command handler, I get an entity, and then save it in an event to be reduced afterwards by another entity. The result is that what's saved in the event is the entity snapshot, and not the entity itself.

Here's the code of the command handler:

```ts
@Command({
  authorize: [User],
})
export class AnswerQuestion {
  public constructor(readonly questionId: UUID, readonly reviewId: UUID, readonly answer: string) {}

  public static async handle(command: AnswerQuestion, register: Register): Promise<void> {
    const question = await Booster.entity(Question, command.questionId)
    // Down here I'm saving the 'Question' entity
    register.events(new QuestionAnswered(command.reviewId, question, JSON.parse(command.answer)))
  }
}
```

After all reductions, etc... The read model that I get is:

```json
{
  "data": {
    "ReviewReadModels": [
      {
        // ...other fields...
        "answers": [
          {
            "question": {
              "snapshottedEventCreatedAt": "2021-07-15T12:18:05.928Z",
              "entityTypeName_kind": "Question-snapshot",
              "entityTypeName_entityID_kind": "Question-ade15543-f9a3-4cf7-8828-6defb37f069b-snapshot",
              "createdAt": "2021-07-15T12:18:06.824Z",
              "deleted": false,
              "kind": "snapshot",
              "requestID": "946cc0e2-3738-4f0a-971c-f36bcd16bf5a",
              "typeName": "Question",
              "entityTypeName": "Question",
              "entityID": "ade15543-f9a3-4cf7-8828-6defb37f069b",
              "version": 1,
              "value": {
                "attachments": [],
                "deleted": false,
                "name": "How is this going?",
                "options": [
                  "good",
                  "bad",
                  "ugly"
                ],
                "description": "Tell us your opinion",
                "id": "ade15543-f9a3-4cf7-8828-6defb37f069b",
                "questionType": "SINGLE_SELECTION"
              }
            },
            "answer": "yes"
          }
        ]
      }
    ]
  }
}
```

## Expected behavior

An envelope should NEVER reach user code, so this is a bug. What should be saved is the `.value` field of the envelope

## Possible Solution

Add a `.value` in this method 
https://github.com/boostercloud/booster/blob/ca350d092bb814ae770fc3401e7d01afbfd77a0b/packages/framework-core/src/booster.ts#L128

### Environment

- Booster version: 0.17

### LINKED ISSUE #918: Booster.entity does not return the entity
# Bug Report

## Current Behavior

When the application runs the `Booster.entity()` asynchronous function, instead of returning the expected entity, it returns and object which contains the entity in the `value` attribute:

```json
{
    "entityTypeName_entityID_kind": "User-EXAMPLE_ID-snapshot",
    "version": 1,
    "value": {
        "id": "EXAMPLE_ID",
    },
    "snapshottedEventCreatedAt": "2021-07-30T12:20:07.459Z",
    "kind": "snapshot",
    "entityTypeName": "User",
    "entityTypeName_kind": "User-snapshot",
    "typeName": "User",
    "entityID": "EXAMPLE_ID",
    "createdAt": "2021-07-30T12:20:09.200Z",
    "requestID": "REQUEST_ID"
}
```

## Expected behavior

`Booster.entity()` should return the entity.

## Possible Solution

Using the `value` attribute from the object that is returned by this function. 

### Environment

- Booster version: 0.20.2
- Node/npm version: v16.5.0
- OS: macOS Big Sur 11.4

jmbrito01:
## Description
Closes #896 
Closes #918 
Currently, Booster.entity returns a `EventEnvelope` when it should actually return the `Entity`.

## Changes
Changes src/booster.ts:149 from `createInstance(entityClass, entitySnapshotEnvelope)` to `createInstance(entityClass, entitySnapshotEnvelope.value)`

Changes tests/booster.test.ts:213 from `replace(EventStore.prototype, 'fetchEntitySnapshot', fake.returns({  id: '42'  }))` to `replace(EventStore.prototype, 'fetchEntitySnapshot', fake.returns({ value: { id: '42' } }))` to fix core tests

## Checks
- [ ] Project Builds

### DIFF
--- a/packages/framework-core/src/booster.ts
+++ b/packages/framework-core/src/booster.ts
@@ -146,7 +146,7 @@ export class Booster {
   ): Promise<TEntity | undefined> {
     const eventStore = new EventStore(this.config, this.logger)
     const entitySnapshotEnvelope = await eventStore.fetchEntitySnapshot(entityClass.name, entityID)
-    return entitySnapshotEnvelope ? createInstance(entityClass, entitySnapshotEnvelope) : undefined
+    return entitySnapshotEnvelope ? createInstance(entityClass, entitySnapshotEnvelope.value) : undefined
   }
 
   /**
--- a/packages/framework-core/test/booster.test.ts
+++ b/packages/framework-core/test/booster.test.ts
@@ -210,7 +210,7 @@ describe('the `Booster` class', () => {
       config.provider = {} as ProviderLibrary
 
       it('the `entity` function calls to the `fetchEntitySnapshot` method in the EventStore', async () => {
-        replace(EventStore.prototype, 'fetchEntitySnapshot', fake.returns({ id: '42' }))
+        replace(EventStore.prototype, 'fetchEntitySnapshot', fake.returns({ value: { id: '42' } }))
 
         class SomeEntity {
           public constructor(readonly id: UUID) {}


NickSeagull (approved):
Looks awesome! Thanks, will run some tests now 😄

NickSeagull:
Integration tests passing, merging as it is a very short PR

PR #943, opened 2021-09-04
Expose count on paginated queries

thomas-advantitge:
Field is already available on the AWS provider, but not exposed via the GraphQL generator.


javiertoledo (approved):
Does this change require any change in the resolver? 
It would be nice to add a sample using the count field in the integration tests.

thomas-advantitge:
> Does this change require any change in the resolver? It would be nice to add a sample using the count field in the integration tests.

The field is managed by the provider implementation: https://github.com/boostercloud/booster/blob/main/packages/framework-provider-aws/src/library/read-models-searcher-adapter.ts#L49

I'll add an integration test.

javiertoledo:
Closing, as the changes were merged in a different PR

PR #965, opened 2021-10-29
feat(azure): Add Azure Scheduled Commands support

gonzalogarciajaubert:
## Description
There’s a Scheduled Commands feature in AWS that takes a typical cron string and triggers a command on a schedule instead of making it wait for a request. This feature is now supported in Azure

## Changes

* Add PackageAzureFunction classes for: GraphQL, EventHandler and ScheduleCommands function
* Refactor Azure/ApplicationStackBuilder class
* Refactor Azure/FunctionDefinition
* Add Azure rawScheduledInputToEnvelope
* Update Azure Provider to include Azure/rawScheduledInputToEnvelope call

## Checks
- [X] Project Builds

### DIFF
--- a/packages/framework-provider-azure-infrastructure/src/infrastructure/stacks/schedules-functions.ts
+++ b/packages/framework-provider-azure-infrastructure/src/infrastructure/stacks/schedules-functions.ts
@@ -10,15 +10,13 @@ export class SchedulesFunctions {
   public constructor(readonly config: BoosterConfig) {}
 
   public getFunctionDefinitions(): Array<FunctionDefinition> | undefined {
-    if (Object.keys(this.config.scheduledCommandHandlers).length) {
-      return Object.keys(this.config.scheduledCommandHandlers)
-        .map((scheduledCommandName) => this.buildScheduledCommandInfo(scheduledCommandName))
-        .filter((scheduledCommandInfo) => !SchedulesFunctions.isEmpty(scheduledCommandInfo.metadata.scheduledOn))
-        .map((scheduledCommandInfo) =>
-          this.scheduledCommandInfoToTimeTriggerFunction(scheduledCommandInfo)
-        ) as Array<FunctionDefinition>
-    }
-    return undefined
+    if (SchedulesFunctions.isEmpty(this.config.scheduledCommandHandlers)) return
+    return Object.keys(this.config.scheduledCommandHandlers)
+      .map((scheduledCommandName) => this.buildScheduledCommandInfo(scheduledCommandName))
+      .filter((scheduledCommandInfo) => !SchedulesFunctions.isEmpty(scheduledCommandInfo.metadata.scheduledOn))
+      .map((scheduledCommandInfo) =>
+        this.scheduledCommandInfoToTimeTriggerFunction(scheduledCommandInfo)
+      ) as Array<FunctionDefinition>
   }
 
   private scheduledCommandInfoToTimeTriggerFunction(
@@ -42,7 +40,7 @@ export class SchedulesFunctions {
     }
   }
 
-  private static isEmpty(value: ScheduleInterface): boolean {
+  private static isEmpty(value: Record<string, unknown> | ScheduleInterface): boolean {
     return !Object.keys(value).length
   }
 
--- a/packages/framework-provider-azure-infrastructure/src/infrastructure/stacks/schedules-functions.ts
+++ b/packages/framework-provider-azure-infrastructure/src/infrastructure/stacks/schedules-functions.ts
@@ -1,20 +1,9 @@
-import { BoosterConfig, ScheduleInterface } from '@boostercloud/framework-types'
-import { FunctionDefinition } from '../types/functionDefinition'
+import { BoosterConfig, ScheduledCommandMetadata, ScheduleInterface } from '@boostercloud/framework-types'
+import { FunctionDefinition, ScheduleFunctionDefinition } from '../types/functionDefinition'
 
-interface TimeTriggerFunction {
+interface ScheduledCommandInfo {
   name: string
-  config: {
-    bindings: [
-      {
-        type: string
-        name: string
-        direction: string
-        schedule: string
-      }
-    ]
-    scriptFile: string
-    entryPoint: string
-  }
+  metadata: ScheduledCommandMetadata
 }
 
 export class SchedulesFunctions {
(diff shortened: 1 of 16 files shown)


packages/framework-provider-azure-infrastructure/src/infrastructure/stacks/schedules-functions.ts
```diff
@@ -0,0 +1,61 @@
...
+   * An NCRONTAB expression is similar to a CRON expression except that it includes an additional sixth field at the beginning to use for time precision in seconds:
+   * {second} {minute} {hour} {day} {month} {day-of-week}
+   * @param scheduledCommandMetadata
+   * @private
+   */
+  private static createCronExpression(scheduledCommandMetadata: ScheduleInterface): string {
+    const { minute = '*', hour = '*', day = '*', month = '*', weekDay = '*' } = scheduledCommandMetadata
+    const expression = `0 ${minute} ${hour} ${day} ${month} ${weekDay}`
+    const neverRunByDefault = '0 0 5 31 2 ? *'
+    return `${expression !== '* * * ? * *' ? expression : neverRunByDefault}`
```
javiertoledo, line 59: I don't know if I fully understand these two last lines. Some comments here:
* `neverRunByDefault` has 7 items, is that correct or the last asterisk should not be there? 
* I get that the idea is to set an impossible date (February 31th) to avoid the function from running, but why is `'* * * ? * *'` invalid? Doesn't that mean "run every second"? Also, I'm not an expert in corn syntax, but I think that this other pattern means the same: ``'* * * * * ?'`.
* Ideally, I'd prefer that invalid cron expressions just avoid creating the resource than creating the functions with an impossible date.

packages/framework-provider-azure-infrastructure/src/infrastructure/types/functionDefinition.ts
```diff
@@ -0,0 +1,4 @@
+export interface FunctionDefinition {
+  name: string
+  config: object
```
javiertoledo, line 3: Is there any way to have a more specific type for the config property? Maybe you can import the type from the Azure package or something...

packages/framework-provider-azure/src/index.ts
```diff
@@ -67,7 +68,7 @@ export const Provider = (): ProviderLibrary => ({
   },
   // ScheduledCommandsLibrary
   scheduled: {
-    rawToEnvelope: undefined as any,
+    rawToEnvelope: rawScheduledInputToEnvelope,
```
javiertoledo, line 71: 🎉

javiertoledo (changes requested):
Looking great so far, but I left a few questions in the way cron expressions are calculated to make sure that what's there is intended.

It would also be nice to have some unit tests for the feature.

packages/framework-provider-azure-infrastructure/src/infrastructure/stacks/schedules-functions.ts
gonzalogarciajaubert: You are right, I copy/paste it from the AWS version and didn't check this use case. As you said, if we don't want to run it then is better to avoid creating the resource. Let me fix it

gonzalogarciajaubert: About why it could be an invalid configuration I guess it was because it's the default value: 

`const { minute = '*', hour = '*', day = '*', month = '*', weekDay = '*' } = scheduledCommandMetadata`

so we should check if all the fields are `*` but this is valid. I mean, this configuration `0 * * * * *` means to run it each minute which is valid (you can check it here: https://ncrontab.swimburger.net/). 

I would change it to add it always. Thoughts?

packages/framework-provider-azure-infrastructure/src/infrastructure/types/functionDefinition.ts
gonzalogarciajaubert: Let me see if I'm able to figure out the definition (there is not an azure package for it)

gonzalogarciajaubert:
@javiertoledo Added missing tests and improved the types. Could you take a new look? thanks

packages/framework-provider-azure-infrastructure/src/infrastructure/types/functionDefinition.ts
```diff
@@ -0,0 +1,37 @@
+export interface IBinding {
+  type: string
+  name: string
+  direction: string
+}
+
+export interface IScheduleBinding extends IBinding {
+  schedule: string
+}
```
NickSeagull, line 9: It's better use the following style, as it's used throughout the project:

```suggestion
export type IScheduleBinding = IBinding & {
  schedule: string
}
```

packages/framework-provider-azure-infrastructure/test/infrastructure/stacks/schedules-functions.test.ts
```diff
@@ -0,0 +1,117 @@
+import { expect } from '../../expect'
+import { SchedulesFunctions } from '../../../src/infrastructure/stacks/schedules-functions'
+import { BoosterConfig, ScheduledCommandInterface, ScheduleInterface } from '@boostercloud/framework-types'
+import { describe } from 'mocha'
+import { ScheduleFunctionDefinition } from '../../../src/infrastructure/types/functionDefinition'
+
+describe('Creating schedules-functions', () => {
```
NickSeagull, line 7: This should be **scheduleD** instead of **scheduleS**, no?

packages/framework-provider-azure/src/library/scheduled-adapter.ts
```diff
@@ -0,0 +1,26 @@
+import { Logger, ScheduledCommandEnvelope, UUID } from '@boostercloud/framework-types'
+
+interface AzureScheduledCommandEnvelope {
+  bindings: {
+    [name: string]: object
```
NickSeagull, line 5: I'd avoid using `object` as it is equivalent to `any`. Better to use `unknown`

```suggestion
    [name: string]: unknown
```

NickSeagull (changes requested):

packages/framework-provider-azure-infrastructure/src/infrastructure/stacks/schedules-functions.ts
```diff
@@ -0,0 +1,67 @@
...
+  public getFunctionDefinitions(): Array<FunctionDefinition> | undefined {
+    if (Object.keys(this.config.scheduledCommandHandlers).length) {
+      return Object.keys(this.config.scheduledCommandHandlers)
+        .map((scheduledCommandName) => this.buildScheduledCommandInfo(scheduledCommandName))
+        .filter((scheduledCommandInfo) => !SchedulesFunctions.isEmpty(scheduledCommandInfo.metadata.scheduledOn))
+        .map((scheduledCommandInfo) =>
+          this.scheduledCommandInfoToTimeTriggerFunction(scheduledCommandInfo)
+        ) as Array<FunctionDefinition>
+    }
+    return undefined
```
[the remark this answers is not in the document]
gonzalogarciajaubert: Yeah, thanks. I have to do some changes so I'm not going to commit your suggestion directly but create a new one

packages/framework-provider-azure-infrastructure/src/infrastructure/types/functionDefinition.ts
```diff
@@ -0,0 +1,37 @@
+export interface IBinding {
```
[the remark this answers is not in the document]
gonzalogarciajaubert: Sure!

packages/framework-provider-azure-infrastructure/src/infrastructure/types/functionDefinition.ts
gonzalogarciajaubert: Cool, thanks

packages/framework-provider-azure-infrastructure/test/infrastructure/stacks/schedules-functions.test.ts
gonzalogarciajaubert: ouch haha, thanks

NickSeagull (approved):

PR #968, opened 2021-11-03
fix(cli): print failed command

### LINKED ISSUE #637: New project errors out when local Git is not configured with user's email address
Hey team!

Wanted to give Booster a try in my laptop and, when trying to create a new project, it **errored out** because it couldn't find Git's email address from Git's config (indeed, it's not configured in my system). Error description:
```
⠋ Initializing git repositoryError: Error: Could not initialize git repository
Initialized empty Git repository in /home/mykra/code/booster/booster-demo/.git/

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: unable to auto-detect email address (got 'mykra@Lindelap.(none)')

```
Besides, the process **hangs** loading with the `Initializing Git repository` message. 

This is low priority because the message is really clear but just wanted to point out that this exists.

Fecony:
## Description
This PR should fix #637 issue and possibly other issues that might appear when CLI command is stuck

## Changes
- log `.fail` message if it fails (it will exit process and mark step as failed)
- colorize git initialization command error message

## Checks
- [x] Project Builds
~Updated documentation accordingly~

## Additional information
Fixes #637

Screenshot 2021-11-03 at 15 41 22

### DIFF
--- a/packages/cli/src/services/project-initializer.ts
+++ b/packages/cli/src/services/project-initializer.ts
@@ -14,6 +14,7 @@ import * as prettierRc from '../templates/project/prettierrc-yaml'
 import * as mochaRc from '../templates/project/mocharc-yml'
 import { wrapExecError } from '../common/errors'
 import { installAllDependencies } from './dependencies'
+import Brand from '../common/brand'
 
 export async function generateConfigFiles(config: ProjectInitializerConfig): Promise<void> {
   await Promise.all(filesToGenerate.map(renderToFile(config)))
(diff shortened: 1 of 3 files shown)


packages/cli/src/services/project-initializer.ts
```diff
@@ -42,7 +43,7 @@ export async function initializeGit(config: ProjectInitializerConfig): Promise<v
   try {
     await exec('git init && git add -A && git commit -m "Initial commit"', { cwd: projectDir(config) })
   } catch (e) {
-    throw wrapExecError(e, 'Could not initialize git repository')
+    throw wrapExecError(e, Brand.dangerize('Could not initialize git repository'))
```
javiertoledo, line 46: Maybe we could make the `wrapExecError` function apply the `Brand.dangerize` to any string passed, so we get a consistent way to print script errors, what do you think?

javiertoledo (approved):
Very nice touch for Developer Experience 👏🏼

packages/cli/src/services/project-initializer.ts
Fecony: Sounds good @javiertoledo to keep it consistent. I have updated `wrapExecError` function to _dangerize_ any prefix passed to it ✨

PR #976, opened 2021-11-05
Rdiaz82/task/update tsc version

rdiaz82:
## Description
Updated the version of tsc to the latest version and fixed new check types for exceptions. 

## Changes

updated `package.json` and casted all exceptions 

## Checks
- [x] Project Builds


packages/cli/src/common/script.ts
```diff
@@ -126,7 +126,8 @@ export class Script<TContext> {
           throw err
         }, constVoid)
       )
-    } catch (err) {
+    } catch (error) {
+      const err = error as Error
```
javiertoledo, line 130: If this is now required, I wonder if they enabled type annotations in the catch error type:

```suggestion
    } catch (error: Error) {
```

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -66,7 +66,8 @@ export class BoosterGraphQLDispatcher {
       try {
         this.logger.debug(`Decoding current user from auth token: ${envelope.token}`)
         envelope.currentUser = await this.boosterTokenVerifier.verify(envelope.token)
-      } catch (e) {
+      } catch (err) {
+        const e = err as string
```
javiertoledo, line 70: 🤔  Is this intended? Even while you can throw a string, I think that the language wraps it in an Error object...

javiertoledo (approved):
I guess that this is what's needed, but it looks weird to me that we need those casts, I've left a couple of comments just in case there's an alternative. Otherwise, I'm ok with this.

packages/cli/src/common/script.ts
rdiaz82: yeahh!! that what I expected but not! In fact, it is even weirder you can use `error:any` but you can't use any other type. This is what the official documentation says:

https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables

packages/framework-core/src/booster-graphql-dispatcher.ts
rdiaz82: not now! now errors are `unkown` and you are forced to define the type. In that particular case the error was a string.

javiertoledo: That's weird, maybe we should try to fix that and make sure errors are always wrapped in an `Error` object 🤔

rdiaz82: Fixed!! I checked and the function throwing the error was generating a normal `Error` so simply was a matter of changing the streaming error from simply pass the error to `err.message`

rdiaz82:
This was fixed in anothe PR adding a flag, so closing this one!

PR #979, opened 2021-11-10
Azure optimistic concurrency

gonzalogarciajaubert:
## Description
Support optimistc concurrency on Azure Provider

## Changes

- Update tests
- Add optimisticConcurrencyValue to ReadModelInterface
- Add/Update store read models method to handle concurrency with ReadModelInterface

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-provider-aws/src/library/read-models-adapter.ts
+++ b/packages/framework-provider-aws/src/library/read-models-adapter.ts
@@ -69,7 +69,21 @@ export async function fetchReadModel(
     `[ReadModelAdapter#fetchReadModel] Loaded read model ${readModelName} with ID ${readModelID} with result:`,
     response.Items
   )
-  return response.Items as unknown as ReadOnlyNonEmptyArray<ReadModelInterface>
+
+  if (!response.Items || response.Items?.length === 0) {
+    return response.Items as unknown as ReadOnlyNonEmptyArray<ReadModelInterface>
+  }
+
+  return (response.Items as unknown as ReadOnlyNonEmptyArray<ReadModelInterface>).map((value) => {
+    const optimisticConcurrencyValue = generateOptimisticConcurrencyValue(value)
+    return {
+      ...value,
+      boosterMetadata: {
+        ...value?.boosterMetadata,
+        optimisticConcurrencyValue: optimisticConcurrencyValue,
+      },
+    } as ReadModelInterface
+  }) as unknown as ReadOnlyNonEmptyArray<ReadModelInterface>
 }
 
 export async function storeReadModel(
@@ -85,10 +99,9 @@ export async function storeReadModel(
       .put({
         TableName: config.resourceNames.forReadModel(readModelName),
         Item: readModel,
-        ConditionExpression: 'attribute_not_exists(boosterMetadata.version) OR boosterMetadata.version = :version',
-        ExpressionAttributeValues: {
-          ':version': expectedCurrentVersion,
-        },
+        ConditionExpression:
+          'attribute_not_exists(boosterMetadata.optimisticConcurrencyValue) OR boosterMetadata.optimisticConcurrencyValue = :optimisticConcurrencyValue',
+        ExpressionAttributeValues: { ':optimisticConcurrencyValue': expectedCurrentVersion },
       })
       .promise()
   } catch (e) {
@@ -140,4 +153,12 @@ function buildKey(
 ): DynamoDB.DocumentClient.Key {
   if (sequenceKeyName && sequenceKeyValue) return { id: readModelID, [sequenceKeyName]: sequenceKeyValue }
   else return { id: readModelID }
+}
+
+function generateOptimisticConcurrencyValue(value: ReadModelInterface): number {
+  let optimisticConcurrencyValue = 1 // if there is not version then we need to persist the first one
+  if (value.boosterMetadata?.version) {
+    optimisticConcurrencyValue = value.boosterMetadata?.version + 1 // the next version number that we are going to persist
+  }
+  return optimisticConcurrencyValue
 }
--- a/packages/framework-provider-aws/src/library/read-models-adapter.ts
+++ b/packages/framework-provider-aws/src/library/read-models-adapter.ts
@@ -80,7 +80,7 @@ export async function fetchReadModel(
       ...value,
       boosterMetadata: {
         ...value?.boosterMetadata,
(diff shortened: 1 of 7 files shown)


packages/framework-provider-aws/src/library/read-models-adapter.ts
```diff
@@ -69,7 +69,21 @@ export async function fetchReadModel(
...
+    return response.Items as unknown as ReadOnlyNonEmptyArray<ReadModelInterface>
+  }
+
+  return (response.Items as unknown as ReadOnlyNonEmptyArray<ReadModelInterface>).map((value) => {
+    const optimisticConcurrencyValue = generateOptimisticConcurrencyValue(value)
+    return {
+      ...value,
+      boosterMetadata: {
+        ...value?.boosterMetadata,
+        optimisticConcurrencyValue: optimisticConcurrencyValue,
```
javiertoledo, line 83: I guess you can use this syntax here:

```suggestion
        optimisticConcurrencyValue,
```

packages/framework-provider-aws/src/library/read-models-adapter.ts
```diff
@@ -141,3 +154,11 @@ function buildKey(
   if (sequenceKeyName && sequenceKeyValue) return { id: readModelID, [sequenceKeyName]: sequenceKeyValue }
   else return { id: readModelID }
 }
+
+function generateOptimisticConcurrencyValue(value: ReadModelInterface): number {
+  let optimisticConcurrencyValue = 1 // if there is not version then we need to persist the first one
+  if (value.boosterMetadata?.version) {
+    optimisticConcurrencyValue = value.boosterMetadata?.version + 1 // the next version number that we are going to persist
+  }
+  return optimisticConcurrencyValue
```
javiertoledo, line 163: The Rubyist inside of me yields this :trollface: (totally optional, the clearer the better!)

```suggestion
  return (value.boosterMetadata?.version ?? 0) + 1 // the next version number that we are going to persist
```

packages/framework-provider-azure/src/library/read-model-adapter.ts
```diff
@@ -18,7 +25,72 @@ export async function fetchReadModel(
...
+  } as ItemDefinition
+  try {
+    await db
+      .database(config.resourceNames.applicationStack)
+      .container(config.resourceNames.forReadModel(readModelName))
+      .items.create(itemModel)
+    logger.debug('[ReadModelAdapter#insertReadModel] Read model inserted')
+  } catch (err) {
+    // In case of conflict (The ID provided for a resource on a PUT or POST operation has been taken by an existing resource) we should retry it
+    if (err.code == 409) {
```
javiertoledo, line 61: Maybe setting a named constant for this code? There might be a `constants.ts` file somewhere..

```suggestion
    if (err.code == AZURE_CONFLICT_ERROR_CODE) {
```

packages/framework-provider-azure/src/library/read-model-adapter.ts
```diff
@@ -18,7 +25,72 @@ export async function fetchReadModel(
...
+  } as RequestOptions
+  try {
+    await db
+      .database(config.resourceNames.applicationStack)
+      .container(config.resourceNames.forReadModel(readModelName))
+      .items.upsert(readModel, options)
+    logger.debug('[ReadModelAdapter#updateReadModel] Read model updated')
+  } catch (err) {
+    // If there is a precondition failure then we should retry it
+    if (err.code == 412) {
```
javiertoledo, line 88: Same here, maybe...

```suggestion
    if (err.code == AZURE_PRECONDITION_FAILED_ERROR) {
```

Maybe there's some Azure library that provides an easier to read enum for these numbers?

packages/framework-provider-azure/src/library/read-model-adapter.ts
```diff
@@ -28,9 +100,9 @@ export async function storeReadModel(
   readModelName: string,
   readModel: ReadModelInterface
 ): Promise<void> {
-  await db
-    .database(config.resourceNames.applicationStack)
-    .container(config.resourceNames.forReadModel(readModelName))
-    .items.upsert(readModel)
-  logger.debug('[ReadModelAdapter#storeReadModel] Read model stored')
+  const version = readModel.boosterMetadata?.version || 0
```
javiertoledo, line 103: In TypeScript it's more common to use the `??` operator:

```suggestion
  const version = readModel.boosterMetadata?.version ?? 0
```

packages/framework-types/src/concepts/read-model.ts
```diff
@@ -6,6 +6,7 @@ export interface ReadModelInterface {
   id: UUID
   boosterMetadata?: {
     version: number
+    optimisticConcurrencyValue?: string | number
```
javiertoledo, line 9: Would it make sense to always use strings and cast them to numbers on the providers that require them?

javiertoledo (approved):
It looks great! Only left minor comments and suggestions

packages/framework-provider-aws/src/library/read-models-adapter.ts
NickSeagull: I'd say that this one is better because its more idiomatic + we avoid having a `let` which is mostly always bad :)

packages/framework-provider-azure/src/library/read-model-adapter.ts
NickSeagull: Agree

packages/framework-provider-azure/src/library/read-model-adapter.ts
NickSeagull: Yep, always use `??` instead of `||` as the latter will accept falsy values, and `??` checks for nullables

packages/framework-types/src/concepts/read-model.ts
NickSeagull: I'd leave it like this to avoid the casting tbh

NickSeagull (approved):
Just Javi's comments

gonzalogarciajaubert:
Awesome feedback @NickSeagull @javiertoledo I'll take care of it ASAP

packages/framework-provider-aws/src/library/read-models-adapter.ts
gonzalogarciajaubert: Awesome, I didn't know it

PR #1000, opened 2021-12-07
Docs: Clarify how to use multiple AWS profiles

ClaytonFarr:
Many users may not be familiar with how or where to work with AWS environment variables. For example, how are these different than .env set variables?)

Updated documentation to hopefully –

* make it clearer on where to set a variable like AWS_PROFILE
* add detail in warning about how / when to check and unset AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID variables
* added detail that setting AWS_PROFILE needs to be manually done for each deploy/nuke
* added notes for a workaround to tie a profile to an environment via package scripts

## Description

## Checks
- [ ] Project Builds

### DIFF
--- a/docs/chapters/02_getting-started.md
+++ b/docs/chapters/02_getting-started.md
@@ -145,9 +145,11 @@ aws_secret_access_key = <YOUR SECRET ACCESS KEY>
 
 #### Multiple AWS Accounts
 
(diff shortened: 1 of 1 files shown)


javiertoledo (approved):
Great improvements, thanks @ClaytonFarr!

PR #1005, opened 2021-12-24
Local rockets multiprovider

gonzalogarciajaubert:
**Note:** Base branch of this PR is feature/azure-rockets-multiprovider. Please update it if needed

## Description
Add Rockets 2.0 support to Local Provider

## Changes
 - add rockets to the provider

## Checks
- [x] Project Builds


packages/framework-provider-local/src/library/events-adapter.ts
```diff
@@ -88,8 +88,6 @@ export async function storeEvents(
     )
   }
   logger.debug('[EventsAdapter#storeEvents] EventEnvelopes stored')
-
-  await userApp.boosterEventDispatcher(eventEnvelopes)
```
gonzalogarciajaubert, line 92: I'm not 100% sure if this could be removed but if we call this method a class is needed to handle the dispatcher. Could anyone confirm if it's needed or not?

packages/framework-provider-local-infrastructure/src/index.ts
```diff
@@ -41,6 +44,11 @@ export const Infrastructure = (): ProviderInfrastructure => {
       const userProject: UserApp = require(path.join(process.cwd(), 'dist', 'index.js'))
       const graphQLService = new GraphQLService(userProject)
       router.use('/graphql', new GraphQLController(graphQLService).router)
+      if (rockets && rockets.length > 0) {
+        rockets.map((rocket) => {
```
alvaroloes, line 48: Suggestion: this could be a `.forEach`as you are not transforming the array 😉

alvaroloes (approved):
Just one suggestion

packages/framework-provider-local/src/library/events-adapter.ts
gonzalogarciajaubert: I just restore it as it was needed for the Local Provider due to this is a sync process

packages/framework-provider-local-infrastructure/src/index.ts
gonzalogarciajaubert: Good catch!

Issue #1010, opened 2022-01-14
lerna run compile fails at framework-common-helpers package


iyorozuya:
# Bug Report

## Current Behavior

When I run ```lerna run compile``` then I get the following error:
```
lerna ERR! npm run compile exited 1 in '@boostercloud/framework-common-helpers'
lerna ERR! npm run compile stdout:

> @boostercloud/framework-common-helpers@0.24.2 compile /home/shiroyasha/Documents/community/booster/packages/framework-common-helpers
> tsc -b tsconfig.json

src/retrier.ts(18,23): error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'Error'.
src/retrier.ts(19,7): error TS2322: Type 'unknown' is not assignable to type 'Error | undefined'.
  Type 'unknown' is not assignable to type 'Error'.
```

## Expected behavior

```lerna run compile``` shouldn't give an error

### Environment

- Node/npm version: Node v14.18.1/npm 6.14.15
- OS: Ubuntu 21.10

NickSeagull:
You are right! This is due to a new feature in TypeScript :)

You can do either that or cast it like:

```ts
...
} catch (err) {
  const e = err as Error
  ...
}

jesusAlvSoto:
Hey @iyorozuya, do you want to open a PR solving this issue? That would be great ❤️

iyorozuya:
I was fixing other compile errors when running ```lerna run compile```. Definitely opening PR for fixing this issue.

PR #1011, opened 2022-01-22
fix: typecasted e from unknown to Error

iyorozuya:
## Description
Addresses Issue #1010

## Changes
- typecasted error from unknown to Error in retries.ts present in package framework-common-helpers


javiertoledo (approved):
Thanks!

PR #1013, opened 2022-01-23
feat: Added request information to commands + extra JWT token verification function

charlietfe:
## Description

In this PR we have addressed the following useful features:

* The Register object in the Commands now contains the raw request. By now we have only added the `body` and the `headers`. Take into account that the headers will vary between different providers.
* Added a new `extraValidation` function inside the`TokenVerifierConfig` to perform custom validations against the token header or token payload. These validations will be executed always after the JWT standard validations, expiration, issuers checks, and so on. 

* This is the new signature for `TokenVerifierConfig`:

```
export type TokenVerifierConfig = {
  issuer: string
  jwksUri?: string
  publicKey?: string
  rolesClaim?: string
  extraValidation?: (jwtToken: unknown, rawToken: string) => void
}
```

Note: The headers and the payload has been specified as `unknown` since we don't know the content of the token until it has been decrypted.

* Example config:

```
    const configWithExtraValidation = new BoosterConfig('test with extra validation')
    configWithExtraValidation.tokenVerifiers = [
      {
       issuer: 'auth0',
        jwksUri: 'https://myauth0app.auth0.com/.well-known/jwks.json',
        extraValidation: (jwtToken, _rawToken) => {

         if ((jwtToken.headers as any)?.alg !== 'RS512') {
            throw 'Invalid token encoding'
          }

          if ((jwtToken.payload as any)?.['custom:role'] !== 'Admin') {
            throw 'Unauthorized'
          }
        },
      },
    ]
```

## Changes
 
- Changed Register interface
- Changed graphQL adapters on each provider to support request
- Changed Booster configuration object.
- Changed token verifier logic.
- Activated JWT options to include token headers. 
- Added tests.

## Checks
- [X] Project Builds


gonzalogarciajaubert (approved):
LGTM

rdiaz82 (approved):
LGTM

packages/framework-core/src/booster-token-verifier.ts
```diff
@@ -54,12 +55,20 @@ class TokenVerifierClient {
...
-        return resolve(this.tokenToUserEnvelope(decoded))
+        const jwtToken = decoded as any
+        if (this.tokenVerifierConfig?.extraValidation) {
+          try {
+            this.tokenVerifierConfig?.extraValidation(jwtToken?.header, jwtToken?.payload)
+          } catch (err) {
+            reject(err)
+          }
+        }
+        return resolve(this.tokenPayloadToUserEnvelope(jwtToken?.payload))
```
javiertoledo, line 66: Why not add the header to the user envelope too? I don't know the reason at the moment, but it could be useful for security logging and it's cheap to add it.

```suggestion
        return resolve(this.tokenToUserEnvelope(jwtToken))
```

packages/framework-types/src/envelope.ts
```diff
@@ -137,3 +138,8 @@ export interface UserEnvelope {
   role: string
   claims: Record<string, unknown>
 }
+
+export interface RequestEnvelope {
+  headers: unknown
+  body: unknown
```
javiertoledo, line 144: Let's not restrict this only to the header and the body of the request, but include all the context we have, including data from the cloud provider. If we want to highlight some important fields with types it's ok, but I'd add a placeholder to access other fields if necessary. I'd rather have the entire context than only the header and the body:

```suggestion
  request: {
    headers: unknown
    body: unknown
    [key: string] : unknown
  }
  [key: string] : unknown
```

javiertoledo (changes requested):
I think that for this it's better not to decide which fields we show. Let's inject all the data we have to make it available for the devs to decide how to deal with it.

packages/framework-core/src/booster-token-verifier.ts
```diff
@@ -54,12 +55,20 @@ class TokenVerifierClient {
         if (err) {
           return reject(err)
         }
-        return resolve(this.tokenToUserEnvelope(decoded))
+        const jwtToken = decoded as any
+        if (this.tokenVerifierConfig?.extraValidation) {
+          try {
+            this.tokenVerifierConfig?.extraValidation(jwtToken?.header, jwtToken?.payload)
```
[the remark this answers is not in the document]
charlietfe: @javiertoledo Regarding your questions:

* Is there anything in the token apart from the header and the payload? As far as I know, the information is divided into three sections header, payload, and signature. I think we could include them.
* Would it make sense to inject the raw token too? Umm, maybe if the user wants to use the raw token for an external issuer validation or whatever reason it could make sense.

packages/framework-types/src/envelope.ts
charlietfe: @javiertoledo The reason behind the constraint of only adding the header and the body is because the context might be quite long and I don't know if that could cause any problem. Makes sense?

charlietfe: Fixed in d43ea9cb9848e447a39abaa483e34388613d0524

packages/framework-core/src/booster-token-verifier.ts
charlietfe: Fixed in d43ea9cb9848e447a39abaa483e34388613d0524

packages/framework-core/src/booster-token-verifier.ts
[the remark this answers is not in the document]
charlietfe: Fixed in d43ea9cb9848e447a39abaa483e34388613d0524

packages/framework-core/src/booster-token-verifier.ts
```diff
@@ -54,23 +55,33 @@ class TokenVerifierClient {
         if (err) {
           return reject(err)
         }
-        return resolve(this.tokenToUserEnvelope(decoded))
+        const jwtToken = decoded as any
+        if (this.tokenVerifierConfig?.extraValidation) {
+          try {
+            this.tokenVerifierConfig?.extraValidation(jwtToken, token)
```
javiertoledo, line 61: ❤️

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -125,6 +125,7 @@ export class BoosterGraphQLDispatcher {
         },
         pubSub: new NoopReadModelPubSub(),
         storeSubscriptions: true,
+        context: envelope.context,
```
javiertoledo, line 128: ❤️

javiertoledo (approved):
This is awesome, thanks @charlietfe!

PR #1021, opened 2022-02-10
Allow empty commands

charlietfe:
## Description

In this PR, we could create commands without parameters. Until now, at least one parameter was mandatory in commands. With this change, a command could be specified in this way.

```typescript
@Command({
  authorize: 'all',
})
export class NotifyFireAlarm {
  public constructor() {}

  public static async handle(command: NotifyFireAlarm, register: Register): Promise<void> {
    register.events(new FireAlarmNotified(UUID.generate(), 'Fire alarm!'))
  }
}
```

## Changes

We have modified the mutations graphQL generator to not generate a required input if the command class doesn't contain any property.


gonzalogarciajaubert (approved):
LGTM

rdiaz82 (approved):
LGTM, but would be nice to have at least one test

javiertoledo (approved):
🚢 🇮🇹

PR #1037, opened 2022-03-15
Throw specific token error if expired or notBefore exception + Add events to booster-app

gonzalogarciajaubert:
## Description
* Throw specific token error if expired or notBefore exception
* Add events to booster-app.ts 

## Checks
- [x] Project Builds


gonzalogarciajaubert:
Closed. Two new PR's will be generated

PR #1038, opened 2022-03-16
Publish `Booster.logger` for easy access along the core and apps

javiertoledo:
## Description

Minor change to publish the configured `Booster.logger` object (with a fallback to console.log)

## Changes

- Replaced the private static property `logger` in the `Booster` singleton by a getter that proxies the logger defined in the current `Booster.config` object.
- It falls back to `console.log` when no logger is configured.

## Checks
- [x] Project Builds


gonzalogarciajaubert (approved):
LGTM

rdiaz82 (approved):
LGTM

PR #1048, opened 2022-03-30
removed error if join key is not valid and avoid to pass empty instance to readmodels

rdiaz82:
Fixes two bugs:

Removed error and just log a warning if a joining key doesn't exists in read models
Avoid passing an empty object instance if the entity doesn't exists into the read models

## Checks
- [X] Project Builds


packages/framework-core/src/services/read-model-store.ts
```diff
@@ -41,6 +40,12 @@ export class ReadModelStore {
         const entityInstance = createInstance(entityMetadata.class, entitySnapshotEnvelope.value)
         const readModelID = this.joinKeyForProjection(entityInstance, projectionMetadata)
         const sequenceKey = this.sequenceKeyForProjection(entityInstance, projectionMetadata)
+        if (!readModelID) {
+          this.logger.debug(
```
gonzalogarciajaubert, line 44: What do you think about using a warn here?

PR #1049, opened 2022-03-31
Add GlobalErrorHandler

gonzalogarciajaubert:
## Description
Add a GlobalErrorHandler

## Checks
- [X] Project Builds


packages/framework-core/src/booster-command-dispatcher.ts
```diff
@@ -33,17 +38,23 @@ export class BoosterCommandDispatcher {
...
+        commandEnvelope.currentUser
+      )
+
+      const commandInstance = createInstance(commandClass, commandInput)
+
+      register = new Register(commandEnvelope.requestID, commandEnvelope.currentUser, commandEnvelope.context)
+      this.logger.debug('Calling "handle" method on command: ', commandClass)
+      result = await commandClass.handle(commandInstance, register)
+    } catch (e) {
+      throw await this.globalErrorDispatcher.dispatch(new CommandHandlerGlobalError(commandEnvelope, e))
```
javiertoledo, line 56: Would it make sense to provide a way to avoid re-throwing the error if it can be handled in the global handler? For instance, we allow the global error dispatcher to return `undefined`, and in this case, we don't throw an error here, but silently ignore the error. If it's different than `undefined` we keep the behavior defined here and re-throw it.

packages/framework-core/src/booster-event-dispatcher.ts
```diff
@@ -96,7 +98,12 @@ export class BoosterEventDispatcher {
           const eventInstance = createInstance(eventClass.class, eventEnvelope.value)
           const register = new Register(eventEnvelope.requestID, eventEnvelope.currentUser)
           logger.debug('Calling "handle" method on event handler: ', eventHandler)
-          await eventHandler.handle(eventInstance, register)
+          try {
+            await eventHandler.handle(eventInstance, register)
+          } catch (e) {
+            const globalErrorDispatcher = new BoosterGlobalErrorDispatcher(config, logger)
+            throw await globalErrorDispatcher.dispatch(new EventHandlerGlobalError(eventInstance, e))
```
javiertoledo, line 105: Same here, should we allow ignoring an error?

javiertoledo (approved):
Great work! Just left a couple of comments.

packages/framework-core/src/booster-command-dispatcher.ts
gonzalogarciajaubert: It makes sense, in fact, the calls to the dispatch method should be without the `throw` as the dispatch method is the one throwing the exception (I'll fix that):

```typescript
  await this.globalErrorDispatcher.dispatch(new CommandHandlerGlobalError(commandEnvelope, e))
```

The problem with ignoring the error is that it can cause unexpected behaviours. For instances, if there is an error creating the instance in this method, the `register` variable could be undefined and the next call to: `await RegisterHandler.handle(this.config, this.logger, register)` could fail because of that.

gonzalogarciajaubert: I'm going to move the register creation outside the try/catch to avoid this problem

docs/chapters/04_features.md
```diff
@@ -1305,6 +1305,90 @@ boost nuke -e <environment name>
...
+ 
+
+## Error handling
+
+Booster includes a global error handler annotation `@GlobalErrorHandler` that will catch all errors that are thrown by:
+* **Command Handling Errors**: Errors thrown by the `handle` method of the command.
+  **Program handling errors**: Errors thrown by the ScheduledCommand `handle` method.
+  **Event Handle errors**: Errors thrown by the `Event Handle` method.
+* **Reducer errors**: Errors thrown by the `@Reduces` method of the entity.
+* **Projection errors: Errors thrown in the ReadModel `@Projects` method.
```
rdiaz82, line 1317: I would say that is missing `**` for closing the bold text

rdiaz82 (approved):
LGTM, just a small comment

Issue #1053, opened 2022-04-07
Use Array<UUID> field as joinKey in Read Model projection annotation


damagode:
## Description

Sometimes I need to use an Array of UUIDs as `joinKey` in Read Model Projections. This is not supported right now and I think it could be a great addition. 

In my mind, the wanted behavior would be to apply the same projection to all of the ids inside the array. 

Something like:

_Entity:_
```
array: [id1, id2]
```

_Projection:_
```
@Projects(Entity, 'array')
```

_Behavior:_
```
Projection logic applied to Read Model with id1
Projection logic applied to Read Model with id2
```

## Possible Solution

In `ReadModelStore` class, we can make the `joinKeyForProjection` to return an array and apply the projection logic to all of the elements individually.

javiertoledo:
Closing, as it's done and merged 🚀

PR #1054, opened 2022-04-07
Issue 1053: Add Array joinKey support in read model projections

damagode:
## Description
This PR is related to the need of using array fields in the `joinKey` for read model projections. The implemented behavior is to execute the projection for all the UUIDs in the array independently. So we will call the projection logic for each read model in the array.

The change is backward compatible with previous projections methods, so no changes are required in projects using the previous version.

## Changes

- Update `joinKeyForProjection` method to always return `Array<UUID>` or `undefined` in case the key is missing
- Updated `project` method to process and call `applyProjectionToReadModel` for each UUID in the `joinKey` array
- Add an optional parameter to the `projectionFunction` call in order to pass the current readModelID we are processing from the array
- Adding tests with array `joinKey`
- Adding integration tests with array `joinKey`
- Updating documentation

## Checks
- [x] Project Builds
- [x] Backward compatibility

## Additional information
- Related Issue: https://github.com/boostercloud/booster/issues/1053


packages/framework-core/src/services/read-model-store.ts
```diff
@@ -35,42 +35,52 @@ export class ReadModelStore {
...
+              this.applyProjectionToReadModel(
+                entityInstance,
+                projectionMetadata,
+                readModelName,
+                readModelID,
+                sequenceKey
+              ),
+            OptimisticConcurrencyUnexpectedVersionError
+          )
+        })
```
javiertoledo, line 70: I **think** that `async`/`await` has no effect in this code, as the `map` operation does not await the callback function. As you're flattening the results and feeding them to the `Promise.allSettledAndFulfilled`, it's safe to remove them.

```suggestion
        return readModelIDList.map((readModelID: UUID) => {
          this.logger.debug(
            '[ReadModelStore#project] Projecting entity snapshot ',
            entitySnapshotEnvelope,
            ` to build new state of read model ${readModelName} with ID ${readModelID}`,
            sequenceKey ? ` sequencing by ${sequenceKey.name} with value ${sequenceKey.value}` : ''
          )

          return retryIfError(
            this.logger,
            () =>
              this.applyProjectionToReadModel(
                entityInstance,
                projectionMetadata,
                readModelName,
                readModelID,
                sequenceKey
              ),
            OptimisticConcurrencyUnexpectedVersionError
          )
        })
```

javiertoledo (approved):
This is great, thanks @damagode!

I'm pre-approving this PR, but before merging this new feature, we should:
- [ ] Add some documentation that explains how to use it
- [ ] Implement an integration test that checks the new scenario

packages/framework-integration-tests/src/entities/pack.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * An order object represents a completed order that's ready to be delivered
+ */
+@Entity
+export class Pack {
+  public constructor(readonly id: UUID, readonly name: string, readonly products: Array<UUID>) {}
+
+  @Reduces(PackCreated)
+  public static createOrder(event: PackCreated): Pack {
```
gonzalogarciajaubert, line 13:
```suggestion
  public static createPack(event: PackCreated): Pack {
```

packages/framework-integration-tests/src/entities/pack.ts
```diff
@@ -0,0 +1,16 @@
+import { Entity, Reduces } from '@boostercloud/framework-core'
+import { UUID } from '@boostercloud/framework-types'
+import { PackCreated } from '../events/pack-created'
+
+/**
+ * An order object represents a completed order that's ready to be delivered
```
gonzalogarciajaubert, line 6: I think you can fix this comment or remove it

damagode:
Moved to https://github.com/boostercloud/booster/pull/1061

PR #1060, opened 2022-04-15
Logger refactor, fixes and improvements

javiertoledo:
This change generalizes the Booster logger by exposing a new `getLogger` function that allows any Booster module (or a Booster application) to initialize a custom logger with the desired prefix and module name. The function also makes sure that you're loading the logger configured in `Booster.config`. Example:

```typescript
// It takes the prefix from `config.logPrefix` or falls back to 'Booster'
const logger = getLogger(config, 'nameOfMyModule')

logger.debug('This is a log')
// => [Booster]|nameOfMyModule: This is a log

// The prefix can also be overridden on demand
const logger2 = getLogger(config, 'nameOfMyModule', 'MyAppName')

logger2.debug('This is a log')
// => [MyAppName]|nameOfMyModule: This is a log

```

As part of the change, all log calls have been replaced to use the new logger, removing the explicit logger passing from all framework functions. Thanks to this change, all log messages produced by the framework will include a customizable global prefix, as well as the class/module name producing the log for easier message grouping and filtering when troubleshooting errors.

## Checks
- [x] Project Builds


gonzalogarciajaubert (approved):
Great work, thanks! LGTM

PR #1061, opened 2022-04-18
Issue 1053: Add Array joinKey support in read model projections

damagode:
## Description
This PR is related to the need of using array fields in the `joinKey` for read model projections. The implemented behavior is to execute the projection for all the UUIDs in the array independently. So we will call the projection logic for each read model in the array.

The change is backward compatible with previous projections methods, so no changes are required in projects using the previous version.

## Changes

- Update `joinKeyForProjection` method to always return `Array<UUID>` or `undefined` in case the key is missing
- Updated `project` method to process and call `applyProjectionToReadModel` for each UUID in the `joinKey` array
- Add an optional parameter to the `projectionFunction` call in order to pass the current readModelID we are processing from the array
- Adding type checking in projection methods (Compiler will show errors in the projection function signature)
- Adding tests with array `joinKey`
- Adding integration tests with array `joinKey`
- Updating documentation

## Checks
- [x] Project Builds
- [x] Backward compatibility

## Additional information
- Related Issue: https://github.com/boostercloud/booster/issues/1053

### DIFF
--- a/docs/chapters/03_booster-architecture.md
+++ b/docs/chapters/03_booster-architecture.md
@@ -675,14 +675,24 @@ export class UserReadModel {
   public constructor(readonly username: string, /* ...(other interesting fields from users)... */) {}
 
   @Projects(User, 'id')
-  public static projectUser(entity: User, current?: ProjectionResult<UserReadModel>) { // Here we update the user fields}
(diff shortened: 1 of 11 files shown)


gonzalogarciajaubert (approved):
LGTM

javiertoledo (approved):

damagode:
Fixed some integration tests here: https://github.com/boostercloud/booster/pull/1062

PR #1064, opened 2022-04-20
feat: allow response headers to be set from command

tpellegrin:
## Description

## Checks
- [ ] Project Builds


packages/framework-provider-local/src/library/api-adapter.ts
```diff
@@ -3,11 +3,16 @@ import { httpStatusCodeFor, toClassTitle } from '@boostercloud/framework-types'
...
 
-export async function requestSucceeded(body?: any): Promise<APIResult> {
+export async function requestSucceeded(body?: any, headers?: Record<string, string>): Promise<APIResult> {
   return {
     status: 'success',
     result: body,
+    headers: {
+      'Access-Control-Allow-Origin': '*',
+      ...headers,
+    },
```
javiertoledo, line 15: I think that it's better if we set any default headers in the file `booster-graphql-dispatcher.ts` when the `GraphQLResolverContext` object is created, I'll leave another comment there.

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -118,6 +127,7 @@ export class BoosterGraphQLDispatcher {
       }
       const resolverContext: GraphQLResolverContext = {
         connectionID: envelope.connectionID,
+        responseHeaders: {},
```
javiertoledo, line 130: I think this is where we should set the default headers. Ideally, we should provide a configuration option to allow users to override them at config time too, so this would be something like:

```suggestion
        responseHeaders: { ...this.config.defaultResponseHeaders },
```

In this way, we can hardcode the default headers (Or an empty object if there are none) and the user can easily set the default response headers for an environment.

packages/framework-core/src/booster-command-dispatcher.ts
```diff
@@ -41,10 +42,17 @@ export class BoosterCommandDispatcher {
 
     const commandInstance = createInstance(commandClass, commandInput)
 
-    const register = new Register(commandEnvelope.requestID, commandEnvelope.currentUser, commandEnvelope.context)
+    const register = new Register(
+      commandEnvelope.requestID,
+      commandEnvelope.responseHeaders,
```
javiertoledo, line 47: If we remove the `responseHeaders` field from the `CommandEnvelope`, here we should set a reference to the `responseHeaders` object from the context:

```suggestion
      context.responseHeaders,
```

packages/framework-core/src/booster-command-dispatcher.ts
```diff
@@ -41,10 +42,17 @@ export class BoosterCommandDispatcher {
...
+      commandEnvelope.currentUser,
+      commandEnvelope.context
+    )
+
     this.logger.debug('Calling "handle" method on command: ', commandClass)
     const result = await commandClass.handle(commandInstance, register)
     this.logger.debug('Command dispatched with register: ', register)
+    context.responseHeaders = register.responseHeaders
```
javiertoledo, line 55: An advantage of setting in `register.responseHeaders` a reference to `context.responseHeaders` is that we also avoid the need of re-setting it here.

```suggestion
```

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -155,7 +165,13 @@ export class BoosterGraphQLDispatcher {
...
     result.errors = result.errors?.map(toGraphQLErrorWithExtensions)
     this.logger.debug('GraphQL result: ', result)
-    return result
+
+    return {
+      data: result.data,
+      errors: result.errors,
+      extensions: result.extensions,
+      responseHeaders: resolverContext.responseHeaders,
+    }
```
javiertoledo, line 174: As the caller already have a reference to `resolverContext`, there's no need to pass it back in the response, we can leave it as it was before:

```suggestion
    return result
```

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -215,6 +215,7 @@ function toEventSearchRequest(args: EventSearchParameters, context: GraphQLResol
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function toCommandEnvelope(commandName: string, value: any, context: GraphQLResolverContext): CommandEnvelope {
   return {
+    responseHeaders: {},
```
javiertoledo, line 218: If we apply the suggested changes, this is no longer needed

```suggestion
```

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -49,7 +53,12 @@ export class BoosterGraphQLDispatcher {
       case 'CONNECT':
         return this.config.provider.graphQL.handleResult(null, graphQLWebsocketSubprotocolHeaders)
       case 'MESSAGE':
-        return this.config.provider.graphQL.handleResult(await this.handleMessage(envelopeOrError))
+        // eslint-disable-next-line no-case-declarations
```
javiertoledo, line 56: Looking again at the code, I noticed that the `GraphQLResolverContext` object doesn't depend on anything but the envelope. I think that we could initialize it at this point to avoid the changes in return types. I think that something like this is possible:

```typescript
case 'MESSAGE':
  const envelope = this.envelopeOrDie(envelopeOrError) // If we want to access the envelope here, we'd need to handle the errors. We could add a new method to do this.
  const context: GraphQLResolverContext = {
    connectionID: envelope.connectionID,
    responseHeaders: { ...this.config.defaultResponseHeaders },
    requestID: envelope.requestID,
    user: envelope.currentUser,
    // operation, <= This object can be added as we go deeper in the call stack
    pubSub: new NoopReadModelPubSub(),
    storeSubscriptions: true,
    context: envelope.context,
  }
  const response = this.handleMessage(envelope)
  return this.config.provider.graphQL.handleResult(response, context.responseHeaders)
```

javiertoledo (changes requested):
Hey @tpellegrin, this is a great start. I left a few comments with specific ideas and suggestions. This part of the codebase is a bit clunky, and the way it was designed is making this change a bit harder than it should be, but there are some small refactors that you could apply, which should help. Please take a look at the comments and let me know what you think ;-)

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -49,7 +48,8 @@ export class BoosterGraphQLDispatcher {
       case 'CONNECT':
         return this.config.provider.graphQL.handleResult(null, graphQLWebsocketSubprotocolHeaders)
       case 'MESSAGE':
-        return this.config.provider.graphQL.handleResult(await this.handleMessage(envelopeOrError))
+        const result = await this.handleMessage(envelopeOrError)
+        return this.config.provider.graphQL.handleResult(result.graphQLResult, result.responseHeaders)
```
javiertoledo, line 52: I think this is 100% fine, but it seems that you're doing nothing with the response headers inside of the `handleMessage` and the other methods. Would it make sense to create the `responseHeaders` instance here instead?

```suggestion
        const responseHeaders = { ...this.config.defaultResponseHeaders }
        const result = await this.handleMessage(envelopeOrError, responseHeaders)
        return this.config.provider.graphQL.handleResult(result, responseHeaders)
```

Doing this also requires passing this object through all functions, so not sure if this will produce any cleaner code...

packages/framework-types/src/config.ts
```diff
@@ -33,6 +33,7 @@ export class BoosterConfig {
   public rockets?: Array<RocketDescriptor>
   public appName = 'new-booster-app'
   public assets?: Array<string>
+  public defaultResponseHeaders?: Record<string, string>
```
javiertoledo, line 36: Unless you're initializing this somewhere else, I think that we should provide a default value here:

```suggestion
  public defaultResponseHeaders: Record<string, string> = {}
```

packages/framework-types/src/envelope.ts
```diff
@@ -16,6 +16,7 @@ export interface Envelope {
 
 export interface CommandEnvelope extends Envelope {
   typeName: string
+  responseHeaders: Record<string, string>
```
javiertoledo, line 19: Not sure if we still need this

```suggestion
```

packages/framework-core/src/services/graphql/websocket-protocol/graphql-websocket-protocol.ts
```diff
@@ -21,9 +22,7 @@ import {
 import { BoosterTokenVerifier } from '../../../booster-token-verifier'
 
 export interface GraphQLWebsocketHandlerCallbacks {
-  onStartOperation: (
-    envelope: GraphQLRequestEnvelope
-  ) => Promise<AsyncIterableIterator<ExecutionResult> | ExecutionResult>
+  onStartOperation: (envelope: GraphQLRequestEnvelope) => Promise<DispatchResult>
```
javiertoledo, line 25: Are you sure that this interface needs to change?

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -218,6 +218,7 @@ function toCommandEnvelope(commandName: string, value: any, context: GraphQLReso
     requestID: context.requestID,
     currentUser: context.user,
     typeName: commandName,
+    responseHeaders: context.responseHeaders,
```
javiertoledo, line 221: I think that this is no longer needed, as we're keeping a reference in the graphQL dispatcher.

```suggestion
```

javiertoledo (changes requested):
This is looking good, but there might be remainings of the initial change that can be removed, left a few comments pointing out some changes that might not be needed ;-)

.vscode/launch.json
```diff
@@ -139,7 +139,10 @@
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen",
       "protocol": "inspector",
-      "runtimeArgs": ["--preserve-symlinks"]
+      "runtimeArgs": ["--preserve-symlinks"],
+      "env": {
+        "BOOSTER_ENV": "test"
+      }
```
javiertoledo, line 145: Fix to help debugging core tests with VSCode

packages/framework-core/src/booster-command-dispatcher.ts
```diff
@@ -43,6 +44,7 @@ export class BoosterCommandDispatcher {
     let result: unknown
     const register: Register = new Register(
       migratedCommandEnvelope.requestID,
+      context.responseHeaders,
```
javiertoledo, line 47: I added the context as a parameter here to allow building the register directly from the context and avoid having the headers object in the `CommandEnvelope`. The previous solution worked, but a command envelope is meant to be a final object because in some implementations it could be useful to be serializable, so it's better that it doesn't contain any mutable state.

packages/framework-core/src/booster-graphql-dispatcher.ts
```diff
@@ -139,9 +139,9 @@ export class BoosterGraphQLDispatcher {
       switch (operationData.operation) {
         case 'query':
         case 'mutation':
-          return this.handleQueryOrMutation(queryDocument, resolverContext)
+          return await this.handleQueryOrMutation(queryDocument, resolverContext)
         case 'subscription':
-          return this.handleSubscription(queryDocument, resolverContext)
+          return await this.handleSubscription(queryDocument, resolverContext)
```
javiertoledo, line 144: This change is unrelated to the purpose of this PR, but when I was fixing the tests I realized that we had a potential bug here (probably introduced after the recent changes for global error handling). When there's an error in `handleQueryOrMutation` or `handleSubscription` methods, the catch in line 146 cannot see the exception and it is not properly handled. /cc @gonzalogarciajaubert @alvaroloes @NickSeagull

packages/framework-core/test/booster-command-dispatcher.test.ts
```diff
@@ -109,12 +109,56 @@ describe('the `BoosterCommandsDispatcher`', () => {
...
+
+      const context = {
+        responseHeaders: {},
+      }
+
+      await new BoosterCommandDispatcher(config as any, logger).dispatchCommand(commandEnvelope as any, context as any)
+
+      expect(ProperlyHandledCommand.handle).to.have.been.calledWithMatch(commandValue, { responseHeaders: {} })
+      expect(context.responseHeaders).to.deep.equal({ 'Test-Header': 'test' })
+    })
```
javiertoledo, line 159: Added this test that checks that the handler can change the `responseHeaders` in the context.

packages/framework-core/test/booster-graphql-dispatcher.test.ts
```diff
@@ -228,7 +233,56 @@ describe('the `BoosterGraphQLDispatcher`', () => {
...
+          expect(executeFake).to.have.been.calledWithExactly({
+            schema: match.any,
+            document: match.any,
+            contextValue: match(resolverContext),
+            variableValues: match(graphQLVariables),
+            operationName: match.any,
+          })
+          expect(config.provider.graphQL.handleResult).to.have.been.calledWithExactly(graphQLResult, {
+            'Test-Header': 'Test-Value',
+          })
```
javiertoledo, line 285: This test checks that the whole method chain is able to keep the state of the `responseHeaders` object in both directions (while the handler is being called and after the handler have changed the headers)

javiertoledo (approved):
@tpellegrin I added a commit to the PR with a couple of changes to avoid modifying the `CommandEnvelope` object, which is desirable to keep its immutability condition. I also added a couple of unit tests to check the new feature and fixed a potential bug (unrelated to this change)

packages/framework-core/src/booster-graphql-dispatcher.ts
gonzalogarciajaubert: I don't think this change is on main @javiertoledo but only in this PR (from a previous commit). I mean, the removal of the `await` was done in this PR and it's good that you restored it. You can verify it on the "file changes" tab. This line is not there because it's not modified

packages/framework-core/test/booster-graphql-dispatcher.test.ts
```diff
@@ -178,14 +182,14 @@ describe('the `BoosterGraphQLDispatcher`', () => {
           const dispatcher = new BoosterGraphQLDispatcher(config, logger)
           replace(gqlValidator, 'validate', fake.returns([]))
 
+          //await expect(dispatcher.dispatch({})).to.be.rejectedWith(errorRegex)
```
NickSeagull, line 185: Commented code, maybe it should be deleted?

NickSeagull (approved):
Pre-approving, just one small comment

PR #1067, opened 2022-04-22
Promisified public key field in the token verifier config

javiertoledo:
Promisified the `publicKey` field in `TokenVerifierConfig` to allow async ways to load the public key (like reading it from a key vault or a file)


gonzalogarciajaubert (approved):
LGTM

packages/framework-types/src/concepts/projection-metadata.ts
```diff
@@ -1,4 +1,4 @@
-import { AnyClass } from '@boostercloud/framework-types'
+import { AnyClass } from '../typelevel'
```
javiertoledo, line 1: O_o

PR #1070, opened 2022-04-25
Refactored rocket loader in a DRY-er way

javiertoledo:
I found that we had several implementations for the rocket loader. This PR creates a generic one and adapts all provider implementations to use it.


packages/framework-common-helpers/src/rocket-loader.ts
```diff
@@ -0,0 +1,26 @@
+import { RocketDescriptor } from '@boostercloud/framework-types'
+
+export class RocketLoader {
```
gonzalogarciajaubert, line 3: 💯

gonzalogarciajaubert (approved):
LGTM!

PR #1076, opened 2022-05-14
feat(cli): make resource templates publishable

### LINKED ISSUE #1074: Publishable resource templates (stubs)
## Feature Request
From [Discord](https://discord.com/channels/763753198388510780/781960568541806593/973877735028043816)

Implement a command to publish resource [templates](https://github.com/boostercloud/booster/tree/main/packages/cli/src/templates) in the `/stubs` folder to allow the developer to customize the resource template.
Commands that create resource files will need to be updated accordingly.

## Description
New command like `boost stub:publish` that will publish all resource templates to `/stubs` folder. In this folder, the developer will be able to update the resource template to reduce the time it takes to update a newly created template each time.
For example developer might want all new commands to have method and instead of copy pasting this code he can publish stubs and update them accordingly ✨ 

## Possible Solution

- New command like `boost stub:publish`
- Update the methods that create the resources appropriately to find the stub file, and if it does not exist, use the default stub.
- Update test files
- When the default stubs are updated, developers can call `boost stub:publish` again to publish the updated template and see what has changed (to keep track of new updates to the resource template).

## Additional information
- Laravel has `artisan stub:publish` command that publishes all stub files (templates) in `/stubs` folder.

@ClaytonFarr may have more details or other thoughts on this 🐙

Fecony:
## Description
This PR is adding new `boost stub:publish` command that will improve DX by allowing developers to publish and customise resource template files 🗄
image
> Do you see that "look content from updated stub! 🐙" ? 👀  
It will now appear in every new command file!

### Usage
```bash 
$ boost stub:publish 
```
Will create `/stubs` folder in project root and copy each resource [template](https://github.com/boostercloud/booster/tree/main/packages/cli/src/templates) file. If you have published resources before, you will have to confirm overwriting the files, or you can pass the `--force` flag to bypass this.

#### QA
<details>
  <summary>Can I have only one stub for a certain resource?</summary>
  
  Yes! The resource generator will check if you have a custom template or it will use the default template
</details>

<details>
  <summary>How can I keep up with new template updates?</summary>
  
  1. Run `boost stub:publish --force` command
  2. Review changes
  3. done!
</details>

<details>
  <summary>Can I adjust the command template and leave the other resources as they are?</summary>
  
  Yes. You can only have the `command.ts` file in the `/stubs` folder and customize it. The generator will use the default templates for the other resources.
</details>

<details>
  <summary>How can I use the default templates again!?</summary>
  
  Simply delete the `/stubs` folder or a specific resource file.
</details>

## Changes
- New CLI command `stub:publish`
- Tests
  - New code test coverage
  - Fixed old test code formatting
- New `template` method to get default or custom resource template 
- New files
  - `packages/cli/src/services/stub-publisher.ts`
  - `packages/cli/src/commands/stub/publish.ts`
- Removed `packages/cli/src/templates/index.ts` file

## Checks
- [x] Project Builds
- [ ] ~Updated documentation accordingly~

## Additional information
- Closes #1074


packages/cli/src/services/generator.ts
```diff
@@ -13,6 +19,16 @@ export async function generate<TInfo>(target: Target<TInfo>): Promise<void> {
   await fs.outputFile(renderPath, rendered)
 }
 
+export function template(name: string): string {
```
javiertoledo, line 22: This is great as it is, but one nice touch would be to have here a type that restricts the template names only to existing templates so a misuse is detected by the compiler.

One way is using an union type that restricts `string` to only valid values:

```suggestion
export function template(name: 'command' | 'event' | ...): string {
```

Another might be using the actual classes instead of strings:

```suggestion
export function template(name: Command | Event | ...): string {
```

But as I said, the current implementation is good too, we can make this improvement in future PRs too ;-)

javiertoledo (approved):
Great code quality!! Just left a suggestion, but can be left for a future change, thanks for fixing the lint errors in tests too!!

packages/cli/src/services/generator.ts
Fecony: @javiertoledo Thanks for pointing that out! I fixed it by adding `TemplateType`.

Fecony:
Something weird happened to test and it failed 👁👁

gonzalogarciajaubert:
Hey @Fecony , I just re-run the unit tests and they pass. There are some issues we are experiencing when running github actions that cause a socket connections error. We are investigating the issue, but your PR is fine.

PR #1077, opened 2022-05-16
Add entity ids paginated finders

gonzalogarciajaubert:
## Description
Add entity ids paginate finders

## Changes
* Add new Booster.eventsIds method to get a paginated list of latest events for an entity

Example:
`return await Booster.eventsIds(command.entityName, command.limit, command.afterCursor)`

## Checks
- [X] Project Builds

NOTE: Only Azure and Local Providers are supported.

### DIFF
--- a/packages/framework-provider-azure/test/helpers/query-helper.test.ts
+++ b/packages/framework-provider-azure/test/helpers/query-helper.test.ts
@@ -75,7 +75,6 @@ describe('Query helper', () => {
       await search(
         mockCosmosDbClient as any,
         mockConfig,
-        mockLogger,
         mockReadModelName,
         {},
         undefined,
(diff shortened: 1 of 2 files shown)


packages/framework-provider-azure/src/helpers/query-helper.ts
```diff
@@ -18,10 +18,11 @@ export async function search(
   limit?: number | undefined,
   afterCursor?: Record<string, string> | undefined,
   paginatedVersion = false,
-  order?: SortFor<unknown>
+  order?: SortFor<unknown>,
+  projections = '*'
 ): Promise<Array<any> | ReadModelListResult<any>> {
```
javiertoledo, line 23: This method is used to perform any kind of query to any table in the database, so it's kind of coupling all user functions together: If a function requires a change in the `search` logic, then every other client function is at risk of stop working and, as developers, we need to take the extra care of making sure that the changes don't affect other user functions. Not needed for this PR, but I think it would be worth splitting this into separate use-case-specific query builder functions.

javiertoledo (approved):
Great work here! Just left a couple of comments

PR #1083, opened 2022-05-18
Token Verifiers and Authorizers Refactor to allow custom logic

### LINKED ISSUE #1121: Allow users with the same role to access different resources
It would be very useful if 2 users with the same role could access different resources.

Use case: my app has 2 players, Player A belongs to Team 1 and Player B belongs to team 2.
Both players have the role "_Player_"
I want Player A to be able to see all his team mates, that is all players in Team 1, but not the players in Team 2. 
And I want player B to see all players in team 2 but not in Team 1.

tpellegrin:
# Refactor Booster's Token Verifiers to ease implementing custom token verification methods.

Token Verifiers are configured in Booster as plain objects that match the following interface:

```typescript
export type TokenVerifierConfig = {
  issuer: string
  jwksUri?: string
  publicKey?: Promise<string>
  rolesClaim?: string
  extraValidation?: (jwtToken: Record<string, unknown>, rawToken: string) => Promise<void>
}
```

These objects are fed to the config by just adding them to the config.tokenVerifiers array:

```typescript
import { Booster } from '@boostercloud/framework-core'
import { BoosterConfig } from '@boostercloud/framework-types'

  config.appName = 'demoapp'
  config.providerPackage = '@boostercloud/framework-provider-aws'
  config.tokenVerifiers = [
    {
      jwksUri: 'https://demoapp.firebase.com/.well-known/jwks.json',
      issuer: 'https://securetoken.google.com/demoapp',
      rolesClaim: 'firebase:groups',
    }
  ]
})
```

This approach has the limitation that it restricts the way tokens can be validated for a specific application to the default implementation. Over time we have added a couple of improvements and options like being able to load a public key using a promise or allowing users to provide an `extraValidation` function that performs extra validations after the token has been validated by the jwk library.

### New `TokenVerifer` interface and default implementations

This PR is a rework of the token verification logic. It allows users to provide custom implementations as well as use the framework's implementations with sensible defaults. The `TokenVerifierConfig` interface has been replaced by a new `TokenVerifier` interface:

```typescript
interface TokenVerifier {
  /** Returns an UserEnvelope object if the token is valid. Raises an exception if not valid */
  verify(token: string): Promise<UserEnvelope> 
}
```

Booster now provides two `TokenVerifier` implementations with sensible defaults that check a token against a jwksUri or a public key. These implementations can be instantiated and provided to the `tokenVerifiers` config option as follows:

```typescript
import { Booster } from '@boostercloud/framework-core'
import { BoosterConfig } from '@boostercloud/framework-types'

  config.appName = 'demoapp'
  config.providerPackage = '@boostercloud/framework-provider-aws'
  config.tokenVerifiers = [
    new JwskUriTokenVerifier(
      issuer: 'https://securetoken.google.com/demoapp',
      jwksUri: 'https://demoapp.firebase.com/.well-known/jwks.json',
      rolesClaim: 'firebase:groups'
    ),
    new PublicKeyTokenVerifier(
      issuer: 'custom-key-pair',
      publicKeyResolver: getCustomKey(),
      rolesClaim: 'custom:roles'
    )
  ]
})
```

With this design, you can easily create a new class implementing a fully custom `TokenVerifier` implementation from scratch or reuse one of the default verifier classes by extending one of the provided classes.

### Token utilities

You can now use any library to parse and validate the tokens in any way you want, like using the [`jwt` node package](https://www.npmjs.com/package/jwt) in your own way, but Booster also exposes the following functions as building blocks if you want to build a custom implementation but want to reuse Booster's sensible defaults:

```typescript
/**
 * Initializes a jwksRSA client that can be used to get the public key of a JWKS URI using the
 * `getKeyWithClient` function.
 */
export function getJwksClient(jwksUri: string) {
  ...
}

/**
 * Initializes a function that can be used to get the public key from a JWKS URI with the signature
 * required by the `verifyJWT` function. You can create a client using the `getJwksClient` function.
 */
export function getKeyWithClient(
  client: jwksRSA.JwksClient,
  header: jwt.JwtHeader,
  callback: jwt.SigningKeyCallback
): void {
  ...
}

/**
 * Verifies a JWT token using a key or key resolver function and returns a Booster UserEnvelope.
 */
export async function verifyJWT(
  token: string,
  issuer: string,
  key: jwt.Secret | jwt.GetPublicKeyOrSecret,
  rolesClaim?: string
): Promise<UserEnvelope> {
 ...
}
```

### New Authorizer functions

Now you can set an authorizer function in any command's or read model's `authorize` decorator parameter or in an event's `authorizeReadEvent` decorator parameter to perform custom authorization:

```typescript
@Command({
  authorize: async (currentUser) => {
    if (!currentUser.claims['permissions'].includes('Permission-To-Rock')) {
      throw new Error(`User ${currentUser.username} should not be rocking!`)
    }
  }
})
export class PerformIncredibleGuitarSolo {
  ...
}
```

See the documentation changes for more details on how this feature can be used in Read models and entities event streams.

## TODO

- [X] Refactor and custom token verifiers logic
- [X] Token utilities
- [X] Update docs
- [X] Unit tests for the changes and new functionality
- [x] Integration tests with different scenarios

---

Closes #1121


packages/framework-types/src/concepts/token-verifier.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export interface JwskUriTokenVerifier extends ITokenVerifier {
+  readonly jwksUri: string
+}
+
+export interface PublicKeyTokenVerifier extends ITokenVerifier {
+  readonly publicKey: Promise<string>
+}
+
+export type TokenVerifier = JwskUriTokenVerifier | PublicKeyTokenVerifier
```
NickSeagull, line 19: I'd avoid relying only on inheritance to avoid explicit casts/checks by callers (like the `has*` functions defined below).

A tagged union would work well here :)

```suggestion
interface TokenVerifierCommon {
  readonly issuer: string
  readonly rolesClaim: string
  readonly extraValidation?: (jwtToken: Record<string, unknown>, rawToken: string) => Promise<void>
  verify?(token: string): Promise<UserEnvelope>
}

type TokenVerifierInfo =
  | { readonly _tag: 'JwksUriTokenVerifier', readonly jwksUri: string }
  | { readonly _tag: 'PublicKeyTokenVerifier', readonly publicKey: Promise<string> }

export type TokenVerifier = TokenVerifierCommon & TokenVerifierInfo
```

This way, if someone does `if (tokenVerifier._tag === '...')`, not only they will get autocompletion in their IDE for the `_tag`, but also inside of the `if` TS will know the appropriate available info fields

packages/framework-types/src/concepts/token-verifier.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const hasJwskUri = (b: TokenVerifier): b is JwskUriTokenVerifier => {
+  return (b as JwskUriTokenVerifier).jwksUri !== undefined
+}
+
+export const hasPublicKey = (b: TokenVerifier): b is PublicKeyTokenVerifier => {
+  return (b as PublicKeyTokenVerifier).publicKey !== undefined
+}
```
NickSeagull, line 27: If the above is implemented we can remove this code

```suggestion
```

javiertoledo (changes requested):
I think it could be worth restricting with types which `TokenVerifierConfig` options can be used together. Let's give it some thought.

packages/framework-types/src/config.ts
```diff
@@ -225,12 +191,6 @@ export class BoosterConfig {
   }
 }
 
-export const JWT_ENV_VARS = {
```
charlietfe, line 228: @javiertoledo We were keeping this env variables to be backwards compatible with the aws cognito rocket, [take a look at this](https://github.com/boostercloud/rocket-auth-aws-infrastructure/blob/809907db7bc32eb583210832eec60b62a3ced434/src/auth-stack.ts#L398). Anyway, I'm totally agree of removing this but probably some work will be needed in the aws cognito rocket to use the future Booster versions.

packages/framework-core/src/services/token-verifiers/jwsk-token-verifier.ts
```diff
@@ -0,0 +1,35 @@
+import { TokenVerifier, UserEnvelope } from '@boostercloud/framework-types'
+import { getJwksClient, getKeyWithClient, verifyJWT } from '.'
+
+export abstract class JwksTokenVerifier implements TokenVerifier {
```
charlietfe, line 4: 💯

packages/framework-core/src/services/token-verifiers/jwsk-token-verifier.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  public async verify(token: string): Promise<UserEnvelope> {
+    const client = getJwksClient(this.jwksUri)
+    const key = getKeyWithClient.bind(this, client)
+    return verifyJWT(token, this.issuer, key, this.rolesClaim)
+  }
+}
+
+export class PublicKeyTokenVerifier extends JwksTokenVerifier {
```
charlietfe, line 22: 💯

charlietfe (approved):
@javiertoledo Good improvement! Only a minor comment about removing environment variables.

NickSeagull (approved):

packages/application-tester/src/token-helper.ts
```diff
@@ -20,20 +26,18 @@ export class TokenHelper {
       issuer,
       keyid,
     } as jwt.SignOptions
-    if (expiresIn || expiresIn === 0) {
-      options['expiresIn'] = expiresIn
+    if (tokenOptions?.expiresIn || tokenOptions?.expiresIn === 0) {
```
NickSeagull, line 29: Maybe
```suggestion
    if (tokenOptions?.expiresIn !== undefined) {
```

packages/framework-core/src/services/token-verifiers/role-based-token-verifier.ts
```diff
@@ -0,0 +1,36 @@
+import { DecodedToken, TokenVerifier, UserEnvelope } from '@boostercloud/framework-types'
+
+export const DEFAULT_ROLES_CLAIM = 'custom:role'
+
+function rolesFromTokenRole(rolesClaim: unknown): Array<string> {
```
NickSeagull, line 5: I'd indicate the nullability explicitly
```suggestion
function rolesFromTokenRole(rolesClaim: unknown | undefined): Array<string> {
```

packages/framework-core/src/services/token-verifiers/utilities.ts
```diff
@@ -0,0 +1,88 @@
...
+    callback(null, key.getPublicKey())
+  })
+}
+
+/**
+ * Strips the `Bearer` prefix from the Authorization header so it can be fed to a TokenVerifier `verify` method.
+ * @param authorizationHeader The value of the Authorization header
+ * @returns The token part of the Authorization header
+ */
+export const sanitizeAuthorizationHeader = (authorizationHeader: string): string =>
```
NickSeagull, line 49: Maybe this name is better?
```suggestion
export const stripBearerPrefix = (authorizationHeader: string): string =>
```

packages/framework-core/src/services/token-verifiers/utilities.ts
```diff
@@ -0,0 +1,88 @@
...
+        if (err) {
+          return reject(err)
+        }
+        if (!decoded) {
+          return reject(new Error('The token could not be decoded'))
+        }
+        return resolve(decoded as DecodedToken)
+      }
+    )
+  })
```
NickSeagull, line 87: Maybe we can promisify this?
```suggestion
  const jwtVerify = util.promisify(jwt.verify)
  const decoded = jwtVerify(sanitizedToken, key, {
        algorithms: ['RS256'],
        issuer,
        complete: true, // To return headers, payload and other useful token information
      })
  if (!decoded) {
    new Error('The token could not be decoded')
  } 
  return decoded as DecodedToken
  })
```

packages/framework-core/test/booster-command-dispatcher.test.ts
```diff
@@ -51,7 +52,7 @@ describe('the `BoosterCommandsDispatcher`', () => {
       const config = {
         commandHandlers: {
           UnauthorizedCommand: {
-            authorizedRoles: [Thor],
+            authorizer: BoosterAuthorizer.authorizeRoles.bind(null, [Thor]),
```
NickSeagull, line 55: Why keep using `function.bind` instead of currying the callee?

I know this is a test file, but I see the pattern all over the codebase

packages/framework-core/test/booster.test.ts
```diff
@@ -253,4 +254,47 @@ describe('the `Booster` class', () => {
...
+        process.env.BOOSTER_ROLES_CLAIM = 'BOOSTER_ROLES_CLAIM_VALUE'
+      })
+
+      afterEach(() => {
+        delete process.env.BOOSTER_JWT_ISSUER
+        delete process.env.BOOSTER_JWKS_URI
+        delete process.env.BOOSTER_ROLES_CLAIM
+
+        Booster.config.tokenVerifiers = []
+      })
```
NickSeagull, line 272: At some point we should start injecting `process.env` into Booster rather than relying on the variables themselves directly in the codebase, in order to make it more testable

NickSeagull (approved):
Thanks! Left some suggestions 😄

packages/framework-core/test/booster.test.ts
javiertoledo: 👍🏼

packages/framework-core/test/booster-command-dispatcher.test.ts
javiertoledo: What do you mean by "currying the callee"? Do you mean to use nested functions?

NickSeagull: Yep I mean having a `const authorizer = (arg) => BoosterAuthorizer.authorizeRoles([Thor], arg)`.

I'm not sure if we have `fp-ts-std` installed, but also it can be written as:

```ts
curry2(BoosterAuthorizer.authorizeRoles)([Thor])
```

Anyway this is just stylistic, so let's leave this out 😊

PR #1085, opened 2022-05-20
fix(cli): fix publishable resource templates

Fecony:
## Description
This PR is a follow up of PR #1076, which introduces the new `boost stub:publish` command. It should fix a problem that was discovered by @ClaytonFarr.

**Issue:** In CLI v0.28.3 we have template files`.js` and `.d.ts`, which are compiled versions of TS [template](https://github.com/boostercloud/booster/tree/main/packages/cli/src/templates) files.

## Possible solutions
Before fixing the issue, I had two ways to do it: 

1. Read the exported template content from the `template` constant, and then use ts-morph to generate files exporting the content that would have been read from the template file.
```ts
// Possible solution to generate const export would look like:
sourceFile.addVariableStatement({
  isExported: true,
  declarationKind: VariableDeclarationKind.Const,
  declarations: [{
    name: "template",
    initializer: templateContent, // content from template file
  }],
});
```

2. Since we don't use `.ts` template files anywhere, we can remove `export const template = ...`, change the file extension to `.mustache` and then adapt the code to read the contents of the file directly, since Mustache JS only needs a string to render the content.

### Solution
I decided not to complicate things and chose the second option, but changed the file extension to a shorter one, `.stub`.
It is shorter, shows the purpose of the file and does not make life difficult for the developer. We also don't use MustacheJS much, we only use it for rendering template files.

> To test this approach, I compiled the CLI, replaced the `dist` files in the global `node_modules` folder for the booster cli and tried running commands to publish stubs and then tried different ways to generate resources (with and without stubs folder).

## Changes
- Updated template files to be simple stub files
- Tests
  - Updated template files
  - One more unit tests to make sure we don't copy anything but `.stub` files
  - Adapt tests for new logic
- `copyfiles` dependency for `cli` to copy `.stub` files to dist folder

## Checks
- [x] Project Builds
- [ ] ~Updated documentation accordingly~


packages/cli/package.json
```diff
@@ -75,7 +76,7 @@
     "format": "prettier --write --ext '.js,.ts' **/*.ts **/*/*.ts",
     "lint:check": "eslint --ext '.js,.ts' **/*.ts",
     "lint:fix": "eslint --quiet --fix --ext '.js,.ts' **/*.ts",
-    "compile": "tsc -b tsconfig.json",
+    "compile": "tsc -b tsconfig.json && copyfiles -f src/templates/*.stub  dist/templates",
```
NickSeagull, line 79: Maybe split this into two scripts? Like
```suggestion
    "compile": "tsc -b tsconfig.json",
    "copy-templates": "copyfiles -f src/templates/*.stub dist/templates",
```

And then invoke `npm run copy-templates` from where we're invoking `npm run compile`?

Fecony: @NickSeagull We already use copyfiles [here](https://github.com/boostercloud/booster/blob/main/packages/framework-provider-azure-infrastructure/package.json#L54). And I think it is not possible to run two commands because they are called by lerna in every repo when you run `lerna run compile` 🥲

NickSeagull: Ah nevermind, I thought this was code from the user's template!

NickSeagull (approved):
LGTM!

gonzalogarciajaubert (approved):
LGTM

PR #1090, opened 2022-05-25
fix(cli): fix publishable resource templates

Fecony:
> **Note**
> This is a re-open of previous PR #1085 that was merged and reverted 🥲 

## Description
This PR is a follow up of PR #1076, which introduces the new `boost stub:publish` command. It should fix a problem that was discovered by @ClaytonFarr.

**Issue:** In CLI v0.28.3 we have template files`.js` and `.d.ts`, which are compiled versions of TS [template](https://github.com/boostercloud/booster/tree/main/packages/cli/src/templates) files.

## Possible solutions
Before fixing the issue, I had two ways to do it: 

1. Read the exported template content from the `template` constant, and then use ts-morph to generate files exporting the content that would have been read from the template file.
```ts
// Possible solution to generate const export would look like:
sourceFile.addVariableStatement({
  isExported: true,
  declarationKind: VariableDeclarationKind.Const,
  declarations: [{
    name: "template",
    initializer: templateContent, // content from template file
  }],
});
```

2. Since we don't use `.ts` template files anywhere, we can remove `export const template = ...`, change the file extension to `.mustache` and then adapt the code to read the contents of the file directly, since Mustache JS only needs a string to render the content.

### Solution
I decided not to complicate things and chose the second option, but changed the file extension to a shorter one, `.stub`.
It is shorter, shows the purpose of the file and does not make life difficult for the developer. We also don't use MustacheJS much, we only use it for rendering template files.

> To test this approach, I compiled the CLI, replaced the `dist` files in the global `node_modules` folder for the booster cli and tried running commands to publish stubs and then tried different ways to generate resources (with and without stubs folder).

## Changes
- Updated template files to be simple stub files
- Tests
  - Updated template files
  - One more unit tests to make sure we don't copy anything but `.stub` files
  - Adapt tests for new logic
- `copyfiles` dependency for `cli` to copy `.stub` files to dist folder

## Checks
- [x] Project Builds


packages/cli/package.json
```diff
@@ -20,6 +20,7 @@
     "@oclif/plugin-help": "^5.1",
     "chalk": "^2.4.2",
     "child-process-promise": "^2.2.1",
+    "copyfiles": "^2.3.0",
```
gonzalogarciajaubert, line 23: This dependency is only used for the `compile` script. Could you move it to the `devDependencies` section?

packages/cli/test/services/stub-publisher.test.ts
```diff
@@ -17,7 +17,7 @@ const rewire = require('rewire')
 describe('stub publisher', () => {
```
gonzalogarciajaubert, line 17: Could you add the new command to the documentation?

gonzalogarciajaubert:
Thanks @Fecony , could you rebase this branch with master?

gonzalogarciajaubert (changes requested):

packages/cli/test/services/stub-publisher.test.ts
Fecony: Hey @gonzalogarciajaubert I would love to, but there is no CLI commands section 😢 
What do you think would be a good place to mention this command?

Fecony:
> Thanks @Fecony , could you rebase this branch with master?

@gonzalogarciajaubert I messed up the history when rebasing, so I started from 0 and had to force push 😬

packages/cli/package.json
Fecony: Moved to devDependencies ✅

packages/cli/test/services/stub-publisher.test.ts
gonzalogarciajaubert: You are right! We have two options:

1. The features section: http://docs.booster.cloud/chapters/04_features
2. The going deeper section: http://docs.booster.cloud/chapters/05_going-deeper: 

I think this is a Booster extension so the `going deeper` is the best option.

gonzalogarciajaubert:
> > Thanks @Fecony , could you rebase this branch with master?
> 
> @gonzalogarciajaubert I messed up the history when rebasing, so I started from 0 and had to force push 😬

No problem. Good job!

gonzalogarciajaubert (approved):
LGTM. Documentation would be great but it's approved as it is

Fecony:
@gonzalogarciajaubert I have added "Customizing CLI resource templates" section to "going deeper". 
But I am not sure how good/bad it is 😅

ClaytonFarr:
@Fecony thanks again for all the work you’ve put into this, and @gonzalogarciajaubert and @NickSeagull thanks for the quick reviews and merges. 

For whatever it’s worth as an external set of eyes, the documentation updates read very well to me.

PR #1094, opened 2022-06-01
Migrate snapshot before reduce it and persist it if needed

gonzalogarciajaubert:
## Description
Migrate snapshot before reduce it and persist it if needed

## Checks
- [X] Project Builds

### DIFF
--- a/packages/framework-core/src/services/event-store.ts
+++ b/packages/framework-core/src/services/event-store.ts
@@ -17,28 +17,31 @@ export class EventStore {
   private config: BoosterConfig
   private provider: ProviderLibrary
   private logger: Logger
+  private migrator: Migrator
 
   public constructor(config: BoosterConfig, logger: Logger) {
     this.config = config
     this.provider = config.provider
     this.logger = logger
+    this.migrator = new Migrator(this.config, this.logger)
   }
 
   public async fetchEntitySnapshot(entityName: string, entityID: UUID): Promise<EventEnvelope | null> {
     this.logger.debug(`[EventStore#fetchEntitySnapshot] Fetching snapshot for entity ${entityName} with ID ${entityID}`)
     const latestSnapshotEnvelope = await this.loadLatestSnapshot(entityName, entityID)
+    const migrateSnapshotEnvelope = await this.migrateSnapshot(latestSnapshotEnvelope)
 
     // eslint-disable-next-line @typescript-eslint/no-extra-parens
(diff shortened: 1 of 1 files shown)


javiertoledo (approved):
LGTM. Just left a minor change proposal

packages/framework-core/src/services/event-store.ts
```diff
@@ -172,4 +170,15 @@ export class EventStore {
...
+  private async migrateSnapshot(eventEnvelope: EventEnvelope | null): Promise<EventEnvelope | null> {
+    if (eventEnvelope) {
+      const migratedEventEnvelope = await this.migrator.migrate(eventEnvelope)
+      if (eventEnvelope.version !== migratedEventEnvelope.version) {
+        await this.storeSnapshot(migratedEventEnvelope)
+      }
+      return migratedEventEnvelope
+    }
+    return null
+  }
```
NickSeagull, line 183: Let's avoid using `null`, it is considered bad practice. Use `undefined` instead

```suggestion
  private async migrateSnapshot(eventEnvelope?: EventEnvelope): Promise<EventEnvelope | undefined> {
    if (eventEnvelope) {
      const migratedEventEnvelope = await this.migrator.migrate(eventEnvelope)
      if (eventEnvelope.version !== migratedEventEnvelope.version) {
        await this.storeSnapshot(migratedEventEnvelope)
      }
      return migratedEventEnvelope
    }
    return undefined
  }
```

packages/framework-core/src/services/event-store.ts
```diff
@@ -117,16 +121,10 @@ export class EventStore {
...
       const eventInstance = createInstance(eventMetadata.class, migratedEventEnvelope.value)
       const entityMetadata = this.config.entities[migratedEventEnvelope.entityTypeName]
-      let migratedLatestSnapshot: EventEnvelope | null = null
-      if (latestSnapshot) {
-        migratedLatestSnapshot = await new Migrator(this.config, this.logger).migrate(latestSnapshot)
-      }
-      const snapshotInstance = migratedLatestSnapshot
-        ? createInstance(entityMetadata.class, migratedLatestSnapshot.value)
-        : null
+      const snapshotInstance = latestSnapshot ? createInstance(entityMetadata.class, latestSnapshot.value) : null
```
NickSeagull, line 127:
```suggestion
      const snapshotInstance = latestSnapshot ? createInstance(entityMetadata.class, latestSnapshot.value) : undefined
```

NickSeagull (changes requested):

PR #1108, opened 2022-06-20
Make reducers and projections more resilient

NickSeagull:
## Description

Fixes a bug that makes the entities end up in a weird state because of a reducer that throws. Also, if a projection threw, it would make the read model just fail, now we catch the error, but leave it in the previous state.

## Changes 

* try/catch around user code calling
* Added tests

## Checks
- [X] Project Builds


NickSeagull:
This PR is getting out of scope, gonna split it in smaller chunks

PR #1109, opened 2022-06-21
fix: add UUID and Date types to GraphQL sort generator

thomas-advantitge:
Similar to the filter generation, UUID and Date need to be filtered because they have typeGroup `Class` and we don't want `getClassMetadata` to be executed.

### DIFF
--- a/packages/framework-core/src/services/graphql/query-helpers/graphql-query-sort-builder.ts
+++ b/packages/framework-core/src/services/graphql/query-helpers/graphql-query-sort-builder.ts
@@ -25,11 +25,12 @@ export class GraphqlQuerySortBuilder {
 
   private generateSortFor(prop: PropertyMetadata): GraphQLInputObjectType | GraphQLEnumType {
     let sortByName = `${prop.typeInfo.name}PropertySortBy`
-    sortByName = sortByName.charAt(0).toUpperCase() + sortByName.substr(1).replace(/\[]/g, '')
+    sortByName = sortByName.charAt(0).toUpperCase() + sortByName.substring(1).replace(/\[]/g, '')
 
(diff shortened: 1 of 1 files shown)


packages/framework-core/src/services/graphql/query-helpers/graphql-query-sort-builder.ts
```diff
@@ -25,11 +25,12 @@ export class GraphqlQuerySortBuilder {
 
   private generateSortFor(prop: PropertyMetadata): GraphQLInputObjectType | GraphQLEnumType {
     let sortByName = `${prop.typeInfo.name}PropertySortBy`
-    sortByName = sortByName.charAt(0).toUpperCase() + sortByName.substr(1).replace(/\[]/g, '')
+    sortByName = sortByName.charAt(0).toUpperCase() + sortByName.substring(1).replace(/\[]/g, '')
```
thomas-advantitge, line 28: (unrelated improvement: `substr` is deprecated)

packages/framework-core/src/services/graphql/query-helpers/graphql-query-sort-builder.ts
```diff
@@ -25,11 +25,12 @@ export class GraphqlQuerySortBuilder {
...
 
     if (this.generatedSortByByTypeName[sortByName]) return this.generatedSortByByTypeName[sortByName]
     if (!prop.typeInfo.type || typeof prop.typeInfo.type === 'object') return this.orderType
     if (prop.typeInfo.typeGroup === 'Array') return this.orderType
+    if (prop.typeInfo.name === 'UUID' || prop.typeInfo.name === 'Date') return this.orderType
```
gonzalogarciajaubert, line 33: Good catch

PR #1111, opened 2022-06-21
feat: token verifier improvements

thomas-advantitge:
* Replaces `jsonwebtoken` and `jwks-rsa` dependencies with `jose` because [jose](https://www.npmjs.com/package/jose) has zero dependencies.
* Adds an option `decryptionKey` (next to `publicKey`) which can be used to decrypt JWE tokens. This can be useful when using libraries like [NextAuth.js](https://next-auth.js.org/configuration/options#override-jwt-encode-and-decode-methods).
* Adds stricter typing to the `tokenVerifierConfig` so only one option can be configured at a time.

Note: GitHub Actions build seem to fail because of an unrelated issue of a missing dependency (`@aws-cdk/core`). Probably due to dependency changes in this PR the package-lock.json was updated, exposing the missing dependency.
This missing dependency is also added in this PR: https://github.com/boostercloud/booster/pull/1114/files#diff-548dc534cd80fb7795092faaf28b3d13e17964e46292327b031b68b90c3967edR34


javiertoledo:
Interesting. I've been working on a refactor of the token verifiers in #1083 to allow users to define their own token verifier classes. Our idea was to provide a couple of default implementations, but this change will allow users to decode tokens in any way they want with the libraries of their choice. We choose `jsonwebtoken` for being kind of a reference implementation, but it's interesting that `jose` has less dependencies. Thinking out loud, instead of replacing our current implementation with this one, it could make sense to move it to a rocket, so we make the Booster core "library agnostic". I'll think about it 🤔

javiertoledo:
Closing this PR as I've moved the changes to a new PR (#1147) in which we're also updating the code to work with the new `TokenVerifier` class-based structures.

PR #1114, opened 2022-06-21
feat: yarn workspaces

thomas-advantitge:
This pull request is a suggestion for using yarn instead of npm and using yarn workspaces instead of Lerna.
We've been using this for months in our Booster fork without problems because we had some issues with dependencies and the monorepo structure in the current npm + lerna setup.

This implementation
* Adds missing dependencies (yarn raises fair warnings about these)
* Uses yarn workspaces which has similar features as Lerna
* Is able to use "resolutions", one of the powerful features of yarn to override dependencies
* Also improves "local linking" Booster inside other projects (we sometimes change booster-core code which we want to test in another project before needing to publish a new version)

This PR would still need documentation updates, mainly want to check if the Booster team is interested in shifting to yarn (already saw a spike migrating to yarn)

Usage: just run `yarn` to install all dependencies, run `yarn build` to compile the whole project.

image


javiertoledo:
Hey @thomas-advantitge, thanks for sharing this, we'll definitely consider it. Lerna is clearly not the best solution, and there are multiple voices in the team calling for a switch to "something else" (not a clear winner though), but we haven't found the bandwidth to do so, as it would require us to rewrite the CI/CD scripts. I've been tempted to work on dropping lerna in favor of [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces), which are similar to yarn's, but when I remember about these scripts, the fear of breaking something makes me think it twice. We'll think about it. We're also seeing the need to eventually rewrite the CI/CD to allow keeping multiple versions of the framework alive in separate branches, so that rewrite could be a good time to switch...

NickSeagull:
Hey @thomas-advantitge , I've been playing lately with [RushJS](https://rushjs.io/), which supports npm, yarn, and pnpm under the hoods (personally, I've been using pnpm because its faster), do you have any opinion on that?

I agree that lerna is a pain in the *** 😄

thomas-advantitge:
Hi @javiertoledo, @NickSeagull,

We've also been evaluating the potential solutions a few months ago before deciding to go with yarn workspaces. 
Rush seems like a good tool. We mainly didn't go for it because it adds more complexity than needed (at least for our case).

Some arguments:
* When using npm or yarn workspaces, you just run `npm install` or `yarn` and the package manager takes care of it. With Lerna and Rush you have to be aware to not run npm or yarn commands directly. 
* Less config is needed when using npm or yarn, as you just add the `workspaces` field to the package.json and you're ready to create workspaces.
* We eventually chose yarn above npm because we had dependency issues linking the booster repo to our own project locally, e.g. when altering Booster code and testing it in our project. Also, publishing packages from our fork introduced dependency issues. Like publishing "@advantitge/booster-framework-core", awaiting for example a PR which has not yet been merged in the Booster repo. Yarn `resolutions` helped us multiple times to override or update dependencies.

I definitely would understand is this change is a bit too drastic or the Booster team would want to go in another direction like Rush. Last year we just tried the migration because we were stuck with the Lerna setup, and it "just worked". Since then we've kept the branch up to date with `main` without issues so I thought it was worth sharing!

NickSeagull:
Thanks for sharing!

Now that you mention it, a project that I'm following closely (Effect-TS) is using Yarn workspaces as well, and their argument to use Yarn is exactly the same as yours, the `resolutions` field.

NickSeagull:
Hey, we just switched to MS Rush, which uses pnpm underneath. We've been comparing it to Yarn, and we've found that there were times that it behave better (e.g. being faster). Imma close this one, but feel free to leave your thoughts 😄

PR #1116, opened 2022-06-22
Core event dispatcher refactor and bug fixes

NickSeagull:
This PR reworks the core event dispatcher to:

1. Bug fix: Due to internal implementation details of the databases used for event storage, the Booster event store receives a page of events that can potentially contain events belonging to different entities. If for some reason an exception is triggered in an entity reducer or a read model projection before all events in the page are visited, it could potentially halt the process and leave some events unattended. This change groups events by entity and processes each in a separate thread to gracefully manage errors in an entity without affecting others.
2. Refactored the event store implementation: We had two different functions processing the events, one was used for the main event dispatcher, and the other was used by the `Booster.events` function that fetches the current state of an entity. Now both methods use the same code.
3. Optimization: As part of the deduplication, the `fetchEntitySnapshot` method now caches the generated snapshots by default to avoid reprocessing when the same entity is addressed often.

This PR also includes the changes from #1331


packages/framework-core/src/services/event-store.ts
```diff
@@ -44,40 +49,7 @@ export class EventStore {
...
-      return latestSnapshotEnvelope
-    }
-
-    await this.storeSnapshot(newEntitySnapshot)
-
-    return newEntitySnapshot
-  }
-
-  private async storeSnapshot(snapshot: EventEnvelope): Promise<void> {
+  public async storeSnapshot(snapshot: EventEnvelope): Promise<void> {
```
javiertoledo, line 52: I think that we should have a unique high-level method with a signature like the following:

```typescript
public async fetchEntitySnapshot(entityName: string, entityID: UUID, timestamp?: string): Promise<EventEnvelope | null> {
  /* 
  1. If a timestamp is provided, we get the snapshot at the specified time (enabling the long awaited time-traveling feature 😎 ). If not, you get the latest known state.
  2. It looks for the latest snapshot.
    2.1. If not found, it looks for the events since the beginning of time
    2.2. If no events found, returns null
  3. If snapshot is found, it looks for events after that snapshot
    3.1. If no events, the snapshot is returned right away without caching it
    3.2. If there are events, it reduces all pending events to generate a new snapshot and stores the newly calculated snapshot before returning it.
  */
}
```

Notice that for this to work, it's important to guarantee that the snapshots and the events related to the same entity are being stored in the same shard of the database, otherwise, this algorithm would cause consistency issues at scale.

Also, I think that we should ignore the events that we already have in the event dispatcher, and use them only as a signal that a specific entity needs to be refreshed. So, I'd remove the `pendingEnvelopes` parameter. This makes the algorithm more resilient to event gaps if for some reason a page of events is not properly processed, and in case of errors, failing events are going to be retried until a bug is solved. Making this extra I/O makes event processing slower, but this process is run *after* event ingestion and I think that we gain significantly more resilience.

Caching snapshots every time they're calculated optimizes consecutive reads. With the current algorithm, we have seen how when event handlers need to load the current state of an entity using `Booster.entity`, the snapshot calculation runs several times. It's also safe to store on read, because when the sharding strategy guarantees ordering, snapshot calculation is idempotent: storing two snapshots that are calculated in parallel would have no overall effect, as both calculations will always produce the same exact result. Also, given that the ID of the snapshot is tied to the last event read, both calculations will end up having the same ID, making the second one of them overwrite the first one with the same value, ending with a single record in the database anyway.

I know that having operations that look like read but are also updating are not much functional, but in this case I don't see caching as a side effect, but as a way to optimize the overall algorithm. The algorithm would work exactly the same way if we didn't cache the results, it would just be slower.

javiertoledo (changes requested):
Let's discuss the proposed ideas

packages/framework-core/src/booster-event-dispatcher.ts
```diff
@@ -60,12 +60,12 @@ export class BoosterEventDispatcher {
     readModelStore: ReadModelStore
   ): Promise<void> {
     const logger = getLogger(config, 'BoosterEventDispatcher#snapshotAndUpdateReadModels')
-    const entitySnapshot = await eventStore.calculateAndStoreEntitySnapshot(entityName, entityID, envelopes)
+    const entitySnapshot = await eventStore.fetchEntitySnapshot(entityName, entityID, envelopes)
```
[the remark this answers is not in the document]
NickSeagull: This is done in #1120 :)

packages/framework-core/src/services/event-store.ts
NickSeagull: Sounds good, I'd merge the 3 PRs first and then refactor into this algorithm you're describing in order to avoid merging conflicts in the other PRs given that this is a major restructure

NickSeagull:
One thing that's missing related to this change is marking a Read Model as "not-updated", presumably in the errors array of GraphQL

NickSeagull:
Not really, will pick it up asap @jesusAlvSoto

packages/framework-core/src/services/event-store.ts
```diff
@@ -17,90 +17,72 @@ const originOfTime = new Date(0).toISOString() // Unix epoch
...
-    const logger = getLogger(this.config, 'EventStore#calculateAndStoreEntitySnapshot')
-    logger.debug('Processing events: ', pendingEnvelopes)
-    logger.debug(`Fetching snapshot for entity ${entityName} with ID ${entityID}`)
-    const latestSnapshotEnvelope = await this.loadLatestSnapshot(entityName, entityID)
-    const snapshotsPerNewEntityId: Record<string, EventEnvelope> = {}
-    if (latestSnapshotEnvelope) {
-      snapshotsPerNewEntityId[latestSnapshotEnvelope.entityID.toString()] = latestSnapshotEnvelope
-    }
+      if (newEntitySnapshot?.entityID !== entityID) {
+        logger.debug(
```
gonzalogarciajaubert, line 51: If the newEntitySnapshot is undefined then this condition will be true but it doesn't means that the entity was migrated, right? I think this is only true if (newEntitySnapshot && newEntitySnapshot.entityID !== entityID)

packages/framework-core/src/services/event-store.ts
```diff
@@ -17,90 +17,72 @@ const originOfTime = new Date(0).toISOString() // Unix epoch
...
+          `Migrated entity ${entityName} with previous ID ${entityID} to ${newEntitySnapshot?.typeName} with the new ID ${newEntitySnapshot?.entityID}`,
+          newEntitySnapshot
+        )
+      } else {
+        logger.debug(`Reduced new snapshot for entity ${entityName} with ID ${entityID}: `, newEntitySnapshot)
```
gonzalogarciajaubert, line 56: What happens if newEntitySnapshot is undefined? Could we say that it was migrated?

packages/framework-core/src/services/event-store.ts
```diff
@@ -17,90 +17,72 @@ const originOfTime = new Date(0).toISOString() // Unix epoch
...
 
-    logger.debug(
-      `[EventStore#calculateAndStoreEntitySnapshot] Reduced new snapshots for entity ${entityName} with ID ${entityID}: `,
-      snapshotsPerNewEntityId
-    )
+        return newEntitySnapshot
+      }
 
-    const snapshotsForNewEntityIds = Object.values(snapshotsPerNewEntityId)
+      await this.storeSnapshot(newEntitySnapshot)
```
gonzalogarciajaubert, line 65: Correct me if I'm wrong, but if we persist here then we can have concurrency issues. See the following diagram:

booster_concurrency_3 001

Do you thinks this use case is possible? If so, will Booster lost the event number 3 in this case?

javiertoledo: This is fun, I've always assumed that snapshots were stored using the last event `createdAt` value as their sort key, but reviewing the code it seems that we're storing them as they come with the current timestamp. I think that the current behavior is a mistake. It's okay to store the time of actual creation for reference, but in my opinion, the sortKey should be exactly the same as the latest processed event's.

So yeah, with the current behavior what you're describing is fully right, but I think we should change the way snapshots are stored to use the last event's sortKey as their sortKey. This way, the `loadLatestSnapshot` method, which sorts the snapshots by this field and takes the latest, will always get the snapshot that includes the latest event, no matter in which order it was inserted in the database.

Another thing that could happen with the current implementation is that when you're storing a snapshot that already exists in the database (because two threads calculated the same one after reading the same list of events), the keys don't match because each of them will have their own timestamp, they'll end up duplicated but with different identifiers. If you try to persist an already existing snapshot, it should be ignored.

gonzalogarciajaubert: Makes sense. I just realized that `createdAt` is not only the current timestamp when the entity is built on `core` but it is overridden for each Provider before persisting the entity with a new `current timestamp`.

About your last comment:
```
Another thing that could happen with the current implementation is that when you're storing a snapshot that already exists in the database (because two threads calculated the same one after reading the same list of events), the keys don't match because each of them will have their own timestamp, they'll end up duplicated but with different identifiers. If you try to persist an already existing snapshot, it should be ignored.
```

For Azure, there is only a partition key defined. AFAIK there is not any unique constraint in the container. For Azure, if you want uniqueness, you need to define it explicitly.

javiertoledo: Gotcha, I'm working on fixing this here: https://github.com/boostercloud/booster/pull/1331

packages/framework-core/src/booster-event-dispatcher.ts
```diff
@@ -55,27 +48,17 @@ export class BoosterEventDispatcher {
...
   ): Promise<void> {
     const logger = getLogger(config, 'BoosterEventDispatcher#snapshotAndUpdateReadModels')
-    const snapshotsForIndependentEntities = await eventStore.calculateAndStoreEntitySnapshot(
-      entityName,
-      entityID,
-      envelopes
-    )
-    if (snapshotsForIndependentEntities.length === 0) {
-      logger.debug('No new snapshots generated, skipping read models projection')
+    const entitySnapshot = await eventStore.fetchEntitySnapshot(entityName, entityID)
```
gonzalogarciajaubert, line 55: Should we rename this method as we are not snapshoting here? What do you think about `updateReadModel`?

packages/framework-core/src/services/event-store.ts
```diff
@@ -17,90 +17,72 @@ const originOfTime = new Date(0).toISOString() // Unix epoch
...
+   * Will fetch the latest snapshot for an entity by applying a reduction
+   * since the time of creation of the last snapshot, UNLESS `pendingEnvelopes`
+   * are passed, in that case, it won't load the events, but rather it will
+   * proceed with the creation of the new snapshot by considering `pendingEnvelopes`
+   * as the latest events.
+   *
+   * Also, in order to make next calls faster, this method stores the calculated
+   * snapshot.
+   */
   public async fetchEntitySnapshot(entityName: string, entityID: UUID): Promise<EventEnvelope | null> {
```
gonzalogarciajaubert, line 30: Should we rename this method to make it clear that we are going to create snapshots here?

packages/framework-core/src/services/event-store.ts
```diff
@@ -17,90 +17,72 @@ const originOfTime = new Date(0).toISOString() // Unix epoch
 export class EventStore {
   public constructor(readonly config: BoosterConfig) {}
 
+  /**
+   * Will fetch the latest snapshot for an entity by applying a reduction
+   * since the time of creation of the last snapshot, UNLESS `pendingEnvelopes`
```
gonzalogarciajaubert, line 22: I think we don't have pendingEvenlopes here right?

packages/framework-provider-kubernetes-infrastructure/tsconfig.json
```diff
@@ -1,7 +1,6 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "esModuleInterop": true,
```
NickSeagull, line 4: Another package that loses ESModules support 😂 

Just joking, this one is deprecated so it doesnt matter

NickSeagull (commented):
Hah, I cannot approve this PR as I was the original author. Anyway:

Mouse_Highlight_Overlay

packages/framework-core/src/services/event-store.ts
javiertoledo: Good catch :-D

packages/framework-core/src/services/event-store.ts
javiertoledo: Ugh, let's better create an issue for this or we will never merge this change 😅

PR #1145, opened 2022-08-01
feat: Update deps

NickSeagull:
## Description

Updated many dependencies and also Lerna to v5 in order to make stuff build properly.

Tagging as `feat` so it triggers a minor version release.

## Checks
- [x] Project Builds


packages/cli/package.json
```diff
@@ -12,8 +12,8 @@
   },
   "bugs": "https://github.com/boostercloud/booster/issues",
   "dependencies": {
-    "@boostercloud/framework-core": "^0.30.4",
-    "@boostercloud/framework-types": "^0.30.4",
+    "@boostercloud/framework-core": "*",
+    "@boostercloud/framework-types": "*",
```
javiertoledo, line 16: Shouldn't this match the CLI's version?

packages/cli/src/services/project-checker.ts
```diff
@@ -111,6 +111,9 @@ class LowerCliVersionError extends Error {
 }
 
 async function compareVersionsAndDisplayMessages(cliVersion: string, projectVersion: string): Promise<void> {
+  // If were running tests, the version is *. Also, its a good escape hatch in case the user uses other versioning
+  if (cliVersion === '*' || projectVersion === '*') return
```
javiertoledo, line 115: I'd expand the comment to explain how this works and why, I'm not sure if I understand it.

packages/cli/test/commands/build.test.ts
```diff
@@ -9,9 +9,9 @@ import { IConfig } from '@oclif/config'
 describe('build', () => {
   describe('Build class', () => {
     beforeEach(() => {
-      replace(configService, 'compileProject', fake.resolves({}))
-      replace(projectChecker, 'checkCurrentDirIsABoosterProject', fake.resolves({}))
-      replace(projectChecker, 'checkCurrentDirBoosterVersion', fake.resolves({}))
+      replace(configService, 'compileProject', fake())
+      replace(projectChecker, 'checkCurrentDirIsABoosterProject', fake())
+      replace(projectChecker, 'checkCurrentDirBoosterVersion', fake())
```
javiertoledo, line 14: I guess this change is OK-ish if the tests still pass, but notice that the logic is not the same, the old version is resolving a promise and returning an empty object, while the fake with no params does nothing AFAIK

packages/framework-core/test/decorators/command.test.ts
```diff
@@ -32,7 +32,9 @@ describe('the `Command` decorator', () => {
       expect(commandMetadata).to.be.an('object')
       expect(commandMetadata.class).to.equal(PostComment)
       expect(commandMetadata.properties[0].name).to.equal('comment')
+      console.log('WAT')
       expect(commandMetadata.properties[0].typeInfo.name).to.equal('string')
+      console.log('WAT')
```
javiertoledo, line 37: These seem to be debugging leftovers, I guess you can remove them now that the PR is ready for merge

```suggestion
      expect(commandMetadata.properties[0].typeInfo.name).to.equal('string')
```

packages/framework-provider-aws/package.json
```diff
@@ -20,8 +20,8 @@
     "url": "git+https://github.com/boostercloud/booster.git"
   },
   "dependencies": {
-    "@boostercloud/framework-common-helpers": "^0.30.4",
-    "@boostercloud/framework-types": "^0.30.4"
+    "@boostercloud/framework-common-helpers": "*",
+    "@boostercloud/framework-types": "*"
```
javiertoledo, line 24: I'm not sure how this could work in the long term 🤔 If every package only requires the presence of the others, but it doesn't care which version, it could happen that npm eventually installs, let's say Booster CLI v 1.20.4, but it's okay with framework types 0.24.0 because there are no version restrictions. If we want to allow different packages to have different versions, I think we should set at least a minimum version here to guarantee that the package is at least using a compatible dependency. Still, I understand why allowing different packages to have different versioning schemas, as we're publishing new versions that are essentially the same package as before when a new version has no changes, but keeping all core packages in sync also helps to guarantee that a specific bundle works well together. What do you think?

packages/metadata-booster/src/type-cache.ts
```diff
@@ -0,0 +1,81 @@
...
+  `${name}${isNullable ? '?' : ''}`
+
+type TypeMetadataWithTypeName = TypeMetadata & {
+  typeName: string
+}
+
+const flipName = (typeInfo: TypeMetadataWithTypeName): TypeMetadataWithTypeName => ({
+  ...typeInfo,
+  name: typeInfo.typeName,
+})
```
javiertoledo, line 81: Very cool addition, I'm glad we moved the `metadata-booster` package to the core 🚀

javiertoledo (changes requested):
You changed a few files huh? 😂

Great changes, I left a few comments. The only thing I see that we should discuss is about restricting versions between booster core dependencies. Would it make sense to require at least a minimum version that is guaranteed to work well with each package instead of an asterisk? Unless there's something I'm missing, this setup seems a bit fragile to me, as it would allow, for example, a future version of the CLI, happily work with today's version and fail in runtime because some interface has changed. What do you think?

NickSeagull:
@javiertoledo regarding the `*` version concern: this is actually what the newest version of Lerna advices: https://lerna.js.org/docs/getting-started#bootstrapping-projects

This indicates Lerna that the project uses the dependencies of the folder, and not some other dependency from npm or that is installed globally, so when it is published, those `*` will be replaced by the actual version of the library that is currently in the repo.

packages/cli/test/commands/build.test.ts
NickSeagull: Actually this was compiling for no reason, because those methods return `void` while this resolved with an empty object. These tests passed because we weren't checking the result 😅

packages/framework-core/test/decorators/command.test.ts
NickSeagull: Thanks for catching those! You can see my despair in these lines lol

packages/framework-provider-aws/package.json
NickSeagull: Answered as a PR comment :)

packages/metadata-booster/src/type-cache.ts
NickSeagull: Thanks, although I'd like to come back at this and rewrite this properly as its a bit fragile

javiertoledo:
Oh gotcha! Didn't know about that lerna feature. All my other comments were nuances, so I'll approve the changes 👌

javiertoledo (approved):

PR #1058, opened 2022-04-13
Add Scheduled Commands for Local Provider

gonzalogarciajaubert:
## Description
Add Schedule commands for Local Provider

## Checks
- [X] Project Builds
- [X] Project passes tests and checks
- [ ] Updated documentation accordingly

PR #1169, opened 2022-09-03
fix: implement unit tests for scheduled command in local provider

### LINKED ISSUE #1066: Implement scheduled command's unit tests for local provider
## Description

PR #1058 (Scheduled commands for local provider) is missing the unit tests so we need to implement them.

rodsoaresTD:
## Description

Implement scheduled command's unit tests for local provider. Scheduled command originally implemented in #1058

## Checks
- [x] Project Builds

## Additional information

- fixes #1066


javiertoledo:
Hey @rodsoaresTD, this is great thanks!

javiertoledo (approved):
Hey @rodsoaresTD, this is great thanks!

packages/framework-provider-local-infrastructure/test/scheduler.test.ts
```diff
@@ -0,0 +1,70 @@
...
+import { configureScheduler } from '../src/scheduler'
+import { expect } from '../test/expect'
+import * as scheduler from 'node-schedule'
+
+describe('scheduler', () => {
+  let mockConfig: any
+
+  const userProject = {
+    boosterTriggerScheduledCommand: () => {
+      console.log('triggering scheduled command...')
```
NickSeagull, line 13: I'd avoid having any console.log in the unit tests. I know there are some already, but I believe unit tests shouldn't log unless specified, and given that there is no logger module for the unit tests yet I'd skip it 😄

packages/framework-provider-local/test/library/scheduled-adapter.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    mockConfig.logger = {
+      info: fake(),
+      warn: fake(),
+      error: fake(),
+      debug: loggerDebugStub,
+    }
+  })
+
+  describe('rawScheduledInputToEnvelope', () => {
+    ;[{ typeName: '' }, { typeName: undefined }].forEach(function (arg) {
```
NickSeagull, line 26: Is there any way to make this more readable? 

I really believe that the unit tests are a lot of times the entry points for possible contributors, or even users that are trying to figure out how does the framework work under the hood.

If we keep the code as readable as possible and keep the barrier low here, it can help a lot of people :)

NickSeagull (commented):
Looking good! Left some comments

Is there a way that we can test this feature better? Maybe some more complex cases?

package-lock.json
```diff
@@ -2692,39 +2692,31 @@
 			}
 		},
 		"node_modules/@boostercloud/framework-common-helpers": {
```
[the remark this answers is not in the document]
NickSeagull: Yeah, let's keep these changes

packages/framework-provider-local-infrastructure/test/scheduler.test.ts
rodsoaresTD: Sure thing! I just wanted to have a "do-nothing" function as the `boosterTriggerScheduledCommand`. I can try instead something like `boosterTriggerScheduledCommand: () => {}`, would that be ok?

packages/framework-provider-local/test/library/scheduled-adapter.test.ts
rodsoaresTD: Just to confirm, you're referring specifically to this line only, right?
`;[{ typeName: '' }, { typeName: undefined }].forEach(function (arg) {`

I understand your point, it can be a bit unwieldy for a first-time reader. In that case, I see 2 options.

Option 1 - decompose the expression:
```
const emptyScheduledCommandEnvelope = { typeName: '' }
const undefinedScheduledCommandEnvelope = { typeName: undefined }

const testCases = [emptyScheduledCommandEnvelope, undefinedScheduledCommandEnvelope]
testCases.forEach(function (arg) {
(...)
```

Option 2 - decompose the test itself into 2:
```
it('should throw an error when typeName is empty', async () => {
        const emptyScheduledCommandEnvelope = { typeName: '' }
        await expect(rawScheduledInputToEnvelope(mockConfig, emptyScheduledCommandEnvelope)).to.eventually.be.rejectedWith(
          new Error(
            `typeName is not defined or empty, scheduled command envelope should have the structure {typeName: string }, but you gave ${JSON.stringify(
              arg
            )}`
          ).message
        )
      })

it('should throw an error when typeName is undefined', async () => {
        const undefinedScheduledCommandEnvelope = { typeName: undefined }
        await expect(rawScheduledInputToEnvelope(mockConfig, emptyScheduledCommandEnvelope)).to.eventually.be.rejectedWith(
          new Error(
            `typeName is not defined or empty, scheduled command envelope should have the structure {typeName: string }, but you gave ${JSON.stringify(
              arg
            )}`
          ).message
        )
      })
```

I would personally go with option 1 because I see it as 2 very similar forms of invalid input, but what would you prefer?

packages/framework-provider-local-infrastructure/test/scheduler.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  describe('configureScheduler', () => {
+    beforeEach(() => {
+      replace(scheduler, 'scheduleJob', fake())
+    })
+
+    afterEach(() => {
+      restore()
+    })
+
+    it('should schedule all commands with the corresponding crons', async () => {
```
rodsoaresTD, line 26: @NickSeagull, regarding more complex cases, I can think of this one for this test file:

```
it('should only schedule commands where `scheduledOn` is defined'

(...)
```

This would test the filtering feature of this line specifically:
https://github.com/boostercloud/booster/pull/1058/files#diff-0871e6dff27016ddf8f5c7753a016ebcf1717fa0bfcd7bad37bd6c06de4029b3R13


I can also think of testing multiple cron expressions of higher complexity, such as the examples seen here:
https://www.baeldung.com/cron-expressions
https://docs.oracle.com/cd/E12058_01/doc/doc.1014/e12030/cron_expressions.htm

Did you have other kinds of tests in mind for this file?

packages/framework-provider-local/test/library/scheduled-adapter.test.ts
```diff
@@ -0,0 +1,44 @@
...
+          new Error(
+            `typeName is not defined or empty, scheduled command envelope should have the structure {typeName: string }, but you gave ${JSON.stringify(
+              arg
+            )}`
+          ).message
+        )
+      })
+    })
+
+    it('should return a ScheduledCommandEnvelope with the same typeName', async () => {
```
rodsoaresTD, line 38: Regarding this test file, I can't really think of additional test cases, but I can identify an improvement for this particular test case: ensuring the returned object is a `ScheduledCommandEnvelope`, i.e. that it includes a UUID.

Did you have other test cases in mind for this file, @NickSeagull?

packages/framework-provider-local-infrastructure/test/scheduler.test.ts
NickSeagull: Sure!

packages/framework-provider-local/test/library/scheduled-adapter.test.ts
NickSeagull: Although I usually prefer writing highly abstract code that removes duplication, **for tests** I believe it is much better if every test can be read as a single case that makes clear the point of it.

So for this case, I'd go with option 2

packages/framework-provider-local-infrastructure/test/scheduler.test.ts
NickSeagull: Awesome, that first suggestion sounds great! 🙏 

If you have the time and energy, maybe trying [fast-check](https://github.com/dubzzz/fast-check) to generate some cron expressions could be a good idea. No need to go overboard on this one, as fast-check is a pretty deep library :)

packages/framework-provider-local/test/library/scheduled-adapter.test.ts
NickSeagull: Nah, I think for this one it is good, no need to test schema here :)

packages/framework-provider-local-infrastructure/src/scheduler.ts
```diff
@@ -8,9 +8,14 @@ interface ScheduledCommandInfo {
...
   const triggerScheduleCommand = userProject['boosterTriggerScheduledCommand']
+
+  const isEmpty = (obj: {}) => {
+    return Object.keys(obj).length === 0
+  }
+
   Object.keys(config.scheduledCommandHandlers)
     .map((scheduledCommandName) => buildScheduledCommandInfo(config, scheduledCommandName))
-    .filter((scheduledCommandInfo) => scheduledCommandInfo.metadata.scheduledOn)
+    .filter((scheduledCommandInfo) => scheduledCommandInfo.metadata.scheduledOn && !isEmpty(scheduledCommandInfo.metadata.scheduledOn))
```
rodsoaresTD, line 18: I believe the existing condition in this predicate filters out elements where `scheduledCommandInfo.metadata.scheduledOn` is "[false-ish](https://developer.mozilla.org/en-US/docs/Glossary/Truthy)", to avoid calling `scheduler.scheduleJob` for cron expressions that are undefined, null, etc:

This 2nd condition will filter out `scheduledOn` that are empty (`{}`), to avoid calling the scheduler on the resulting undefined expression.

packages/framework-provider-local-infrastructure/test/scheduler.test.ts
```diff
@@ -0,0 +1,145 @@
...
+            scheduledOn: null,
+          },
+        },
+      }
+
+      configureScheduler(mockConfig as unknown as BoosterConfig, userProject)
+
+      expect(scheduler.scheduleJob).to.have.been.calledWithMatch('cmd1', '5 4 * * *')
+      expect(scheduler.scheduleJob).to.have.been.calledWithMatch('cmd2', '* 4 3 * *')
+      // expect(scheduler.scheduleJob).to.have.been.calledWithMatch('cmdWithEmptyScheduledOn', {})
```
rodsoaresTD, line 72: This assertion was passing before the change here:
https://github.com/boostercloud/booster/pull/1169/files#diff-0871e6dff27016ddf8f5c7753a016ebcf1717fa0bfcd7bad37bd6c06de4029b3R18

If that change is correct, then this commented out assertion should be removed; if the change is wrong, this assertion should be re-enabled and its opposite assertion (`expect(scheduler.scheduleJob).to.not.have.been.calledWithMatch('cmdWithEmptyScheduledOn', {})`) should be removed instead.

javiertoledo:
Closing this PR as the commits have been moved to https://github.com/boostercloud/booster/pull/1362

PR #1178, opened 2022-09-13
Add headers to Azure and Local Api responses

gonzalojaubert:
## Description
Add headers to the requestSuccess method for Azure and Local

## Changes
* Fix new project templates

## Checks
- [X] Project Builds


alvaroloes (approved):

packages/cli/src/templates/project/package-json.ts
```diff
@@ -26,9 +26,9 @@ export const template = `{
     "@types/mocha": "8.2.2",
     "nyc": "15.1.0",
     "prettier":  "2.3.0",
-    "typescript": "4.5.4",
+    "typescript": "4.7.4",
     "ts-node": "9.1.1",
-    "@types/node": "15.0.2",
+    "@types/node": "18.0.6",
```
NickSeagull, line 31: Shouldnt this be 16?

packages/framework-provider-azure/src/library/api-adapter.ts
```diff
@@ -13,11 +13,15 @@ export interface ContextResponse {
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export async function requestSucceeded(body?: any): Promise<ContextResponse> {
+export async function requestSucceeded(
+  body?: any,
```
NickSeagull, line 17: This should be `unknown` not any

packages/cli/src/templates/project/package-json.ts
gonzalojaubert: Not sure about it. ./package.json is `"@types/node": "16.11.7"` but in the integration-folder `package.json` there is not `@types/node`

packages/framework-provider-azure/src/library/api-adapter.ts
gonzalojaubert: I only added `headers?: Record<string, number | string | ReadonlyArray<string>>` not changed the body type. Do you think it should be changed?

adrian-lorenzo: I would try to change it if it is just a 1 liner. If suddenly, this could chain a lot of changes more, so I would keep the PR small in that case and just create tickets / solve it next time.

adrian-lorenzo (approved):
Great!

packages/cli/src/templates/project/package-json.ts
gonzalojaubert: Done

packages/framework-provider-azure/src/library/api-adapter.ts
gonzalojaubert: Done

javiertoledo:
Moved to integration branch, see #1250

PR #1182, opened 2022-09-21
Add port to mountStack for local rockets

gonzalojaubert:
## Description
Add port to mountStack for local rockets.

## Checks
- [X] Project Builds

### DIFF
--- a/packages/framework-provider-local-infrastructure/src/index.ts
+++ b/packages/framework-provider-local-infrastructure/src/index.ts
@@ -48,7 +48,7 @@ export const Infrastructure = (rocketDescriptors?: RocketDescriptor[]): Provider
       router.use('/graphql', new GraphQLController(graphQLService).router)
       if (rockets && rockets.length > 0) {
         rockets.forEach((rocket) => {
-          rocket.mountStack(config, router, port)
+          rocket.mountStack(config, router, { port })
         })
       }
       expressServer.use(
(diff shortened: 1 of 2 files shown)


packages/framework-provider-local-infrastructure/src/index.ts
```diff
@@ -48,7 +48,7 @@ export const Infrastructure = (rocketDescriptors?: RocketDescriptor[]): Provider
       router.use('/graphql', new GraphQLController(graphQLService).router)
       if (rockets && rockets.length > 0) {
         rockets.forEach((rocket) => {
-          rocket.mountStack(config, router)
+          rocket.mountStack(config, router, port)
```
alvaroloes, line 51: It could be possible that we need to pass more specific parameters in the future. In order to avoid any possible breaking changes, I would pass a configuration record as the third parameter, where right now we only pass the `port` field. Something like:
```suggestion
          rocket.mountStack(config, router, { port })
```

adrian-lorenzo: I agree. We are missing a configuration record here! 

We would have the same problem now with the `start` method:

```typescript
 start: async (config: BoosterConfig, port: number): Promise<void> => {
```

Expanding the `start` functionality already implies a potential breaking change

gonzalojaubert: Done

alvaroloes (approved):

javiertoledo (approved):
This is ok, but I think that the rockets system is calling for a redesign 😅

PR #1184, opened 2022-09-27
fix(core): includes filter logic

NickSeagull:
## Description

Fixed logic of includes filter for subscriptions

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
+++ b/packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
@@ -113,11 +113,7 @@ function includes(readModelPropValue: any, element: any): boolean {
   if (Array.isArray(readModelPropValue)) {
     // Check that array includes an object
     if (typeof element === 'object') {
-      return readModelPropValue.some((item) => {
-        return Object.keys(element).every((key) => {
-          return item[key] === element[key]
-        })
-      })
+      return readModelPropValue.some((item) => Object.keys(element).every((key) => item[key] === element[key]))
     }
     // If not, do a regular includes
     return readModelPropValue.includes(element)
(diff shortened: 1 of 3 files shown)


packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
```diff
@@ -113,7 +113,9 @@ function beginWith(readModelPropValue: any, element: string): boolean {
...
 
 function includes(readModelPropValue: any, element: any): boolean {
-  if (!Array.isArray(readModelPropValue)) return false
-  if (readModelPropValue.includes(element)) return true
-  return readModelPropValue.some((prop: any) => Object.keys(prop).some((key) => prop[key] === element[key]))
+  if (Array.isArray(readModelPropValue)) {
+    return readModelPropValue.includes(element)
+  } else {
+    return readModelPropValue.some((prop: any) => Object.keys(prop).some((key) => prop[key] === element[key]))
+  }
```
javiertoledo, line 120: The rubyist inside of me refurbishes this code in this way:

```suggestion
  return Array.isArray(readModelPropValue) && readModelPropValue.includes(element)
    || readModelPropValue.some((prop: any) => Object.keys(prop).some((key) => prop[key] === element[key])
```

But your version is definitely easier to understand for the majority 😅

javiertoledo (approved):
LGTM

packages/framework-core/src/services/pub-sub/read-model-pub-sub.ts
NickSeagull: I'd go with your version 100%, BUT, here we have a lot of `any` types and that's what led to the bug we had.

My plan is to refactor this file for 1.0.0 and remove all the anys so its easier to write expressive code

PR #1187, opened 2022-09-29
[skip ci] Added a helper stack cleaner tool to cleanup stale stacks 

NickSeagull:
## Description
 
We had some stale stacks in our AWS account so I added a script that we can reuse to cleanup those in the future. Also, it is easily extendable to Azure as well.

Code is a bit of a mess, but given its importance (just a side tool) I think it is good to have it here

## Checks
- [x] Project Builds


NickSeagull:
Merging because it doesn't do anything to the framework code at all

PR #1208, opened 2022-10-24
Remove npm dependency

NickSeagull:


packages/cli/src/services/file-system/index.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface FileSystemService {
+  readonly readDirectoryContents: (directoryPath: string) => Effect<unknown, FileSystemError, ReadonlyArray<string>>
+}
+
+export const FileSystemService = tag<FileSystemService>()
+
+/**
+ * Helper SDK to be able to run service methods outside of the layers
+ */
+export const fileSystemInternals = deriveLifted(FileSystemService)(
```
NickSeagull, line 20: This SDK style should be removed as it introduces confusion IMO

packages/framework-integration-tests/integration/helper/sleep.ts
```diff
@@ -6,7 +6,7 @@ export async function waitForIt<TResult>(
   tryFunction: () => Promise<TResult>,
   checkResult: (result: TResult) => boolean | string,
   trialDelayMs = 1000,
-  timeoutMs = 60000
+  timeoutMs = 600000
```
NickSeagull, line 9: I incremented all the timeouts because Azure was wonky, maybe I can turn it back, but it doesnt hurt for now

packages/framework-types/src/effect/index.ts
```diff
@@ -0,0 +1,33 @@
+import { pipe } from '@effect-ts/core'
```
NickSeagull, line 1: This is the module that I have to document thoroughly

tools/eslint-config/index.js
```diff
@@ -38,14 +38,7 @@ module.exports = {
...
     '@typescript-eslint/no-empty-function': 0,
-    '@typescript-eslint/explicit-function-return-type': [
-      'warn',
-      {
-        allowExpressions: true,
-        allowTypedFunctionExpressions: true,
-        allowHigherOrderFunctions: true,
-      },
-    ],
+    '@typescript-eslint/explicit-function-return-type': 0,
```
NickSeagull, line 41: This rule was not being useful at all, as it makes the code that uses `effect-ts` much more contrived

packages/cli/src/services/config-service.ts
```diff
@@ -1,18 +1,27 @@
...
 export async function createDeploymentSandbox(): Promise<string> {
   const config = await compileProjectAndLoadConfig(process.cwd())
   const sandboxRelativePath = createSandboxProject(DEPLOYMENT_SANDBOX, config.assets)
-  await installProductionDependencies(sandboxRelativePath)
+  const effect = gen(function* ($) {
+    const { setProjectRoot, installProductionDependencies } = yield* $(PackageManagerService)
+    yield* $(setProjectRoot(sandboxRelativePath))
+    yield* $(installProductionDependencies())
+  })
+  await unsafeRunEffect(effect, {
```
NickSeagull, line 21: The idea is that in the future we're not using `unsafeRunEffect` all over the place, but rather have a single call at the top level.

packages/cli/src/services/config-service.ts
```diff
@@ -1,18 +1,24 @@
...
+import { gen, runWithLayer } from '@boostercloud/framework-types/src/effect'
+import { LivePackageManager } from './package-manager/live.impl'
 
 export const DEPLOYMENT_SANDBOX = '.deploy'
 
 export async function createDeploymentSandbox(): Promise<string> {
   const config = await compileProjectAndLoadConfig(process.cwd())
   const sandboxRelativePath = createSandboxProject(DEPLOYMENT_SANDBOX, config.assets)
-  await installProductionDependencies(sandboxRelativePath)
+  // await installProductionDependencies(sandboxRelativePath)
```
BorjaZarco: Since this line is commented, could you remove it at all from the code?

packages/cli/src/common/errors.ts
```diff
@@ -10,3 +11,7 @@ export function wrapExecError(e: Error, prefix: string): Error {
   const { stdout, stderr } = e as any
   return new Error(Brand.dangerize(prefix) + '\n' + stdout + stderr)
 }
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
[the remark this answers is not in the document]
BorjaZarco: Could we add a TODO here so we remember to remove this?

BorjaZarco (approved):
LGTM! 👍  I'm missing lots of context, though

NickSeagull:
Integration tests passed in the previous commit and this one was about applying minor suggestions, so I'm merging it already

PR #1212, opened 2022-10-27
Improve DX on project creation

### LINKED ISSUE #1196: Remove most of the parameters from new:project command and other developer experience improvements
## Description

After these months of creating a bunch of projects using Booster, our team feels that the project creation project could be shorter and faster

The **only parameter that makes sense to request is the provider**, the rest feel completely unnecessary for any greenfield project. Our team usually skips most of the parameters when we're creating an application in Booster. I don't believe they are providing value to the user right away, so I think we should remove them and make the process faster. 

The process of creating a Booster project would be **reduced from eight steps to just two**: 
1. Submitting the `new:project` command
2. Selecting the provider

Also, I would also suggest **moving the tags from the providers to the beginning of the line** and also **using coloured text** to make it easier to read for users, since the name of the package is not crucial for their knowledge.

We could **even remove the name of the package**. I don't believe it is providing much value.

Fecony:
## Description
As mentioned by @adrian-lorenzo  this changes should Improve Developer Experience by reducing steps needed to create new project and by moving provider names to the beginning of the line so it's easier to read.

Screenshot 2022-10-27 at 22 49 54

## Changes
- Removed all unnecessary steps that can be configured later (author, license, etc)
- Move Provider names before package names

## Checks
- [ ] Project Builds

## Additional information
Fixes #1196

### DIFF
--- a/packages/cli/src/commands/new/project.ts
+++ b/packages/cli/src/commands/new/project.ts
@@ -143,37 +143,32 @@ export const parseConfig = async (
     })
   }
 
-  const description = await prompter.defaultOrPrompt(
-    flags.description,
-    'What\'s your project description? (default: "")'
-  )
-  const versionPrompt = await prompter.defaultOrPrompt(flags.version, "What's the first version? (default: 0.1.0)")
-  const version = versionPrompt || '0.1.0'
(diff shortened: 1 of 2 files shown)


NickSeagull:
I'm not really sure about this one to be honest, I like the initial wizard...

PR #1231, opened 2022-11-14
Add RegisterHandler.flush method

gonzalojaubert:
## Description
RegisterHandler exported the `handler` method. This change rename the function to `flush` and keep `handle` for compatibility purpose.

Add integration test to check that a call to `flush` method allow to emit the events on the register that were pending.

## Changes
* Add `flush` method
* Refactor `handle` method
* Add integration tests

## Checks
- [X] Project Builds


adrian-lorenzo (approved):

damagode (approved):
LGTM

adrian-lorenzo (approved):
Great!

packages/framework-core/src/booster-register-handler.ts
```diff
@@ -21,16 +23,31 @@ const boosterEventsTypesNames: Array<string> = [
 
 export class RegisterHandler {
   public static async handle(config: BoosterConfig, register: Register): Promise<void> {
-    if (register.eventList.length == 0) {
+    return this.flush(config, register.eventList, register.requestID, register.currentUser)
+  }
+
+  public static async flush(
```
alvaroloes, line 29: I don't fully understand the need of this function, maybe I'm missing something. Let me explain:
- Before there was the `RegisterHandler.handle` method, which basically flushed all the events.
- Why we need another method in this class to do the same? Can't we just make `register.flush` to call the `RegisterHandler.handle` method (through the flusher parameter) and, thus, do not make any change here?

packages/framework-types/src/concepts/register.ts
```diff
@@ -40,4 +44,9 @@ export class Register {
     this.eventList.push(...events)
     return this
   }
+
+  public async flush(config: BoosterConfig): Promise<void> {
```
alvaroloes, line 48: We can make this function not to need the `config` and have a better UX by doing any of the following things:
1. Make the constructor of this class receive the config and store as instance member. Then pass it to the flusher here.
2. Make the `flusher`  to already have the config stored in the function closure.

Maybe point 1 is simpler

packages/framework-types/src/concepts/register.ts
```diff
@@ -1,6 +1,9 @@
 import { EventInterface } from './event'
 import { UserEnvelope, ContextEnvelope } from '../envelope'
 import { UUID } from './uuid'
+import { BoosterConfig } from '../config'
+
+export type FlusherFunction = (config: BoosterConfig, eventList: Array<EventInterface>, requestID: UUID, currentUser?: UserEnvelope) => Promise<void>
```
alvaroloes, line 6: We can make this function to have the same signature of `RegisterHandler.handle` and avoid the need of the `RegisterHandler.flush` method 😃

packages/framework-types/src/concepts/register.ts
gonzalojaubert: I see you point, but there are some considerations:

1. For the option 1. The Register is instantiated several times in different places and not all of them has a config object. This will force us to pass the `config` object for all the dependency tree until we reach the `new Register` method or to refactor a lot of classes.  In my opinion the `flush` call is not something that we expect to be used so much, so adding a config object is not a problem.
2. For the option 2. Not sure if I understood you, `handle` is a static method. What do you mean with `stored in the closure` ??

charlietfe (approved):
lgtm

alvaroloes (approved):

adrian-lorenzo (approved):

NickSeagull (approved):

PR #1235, opened 2022-11-15
fix(ci): Remove nextBump to force asking for changes

NickSeagull:
This PR makes `rush change --verify` to work properly and `rush change` to prompt for version changes.


NickSeagull:
Merging as its one line

PR #1239, opened 2022-11-18
Add ReducerUnknownEvents annotation

gonzalojaubert:
## Description
Add ReducerUnknownEvents annotation to handle any unknown event received

## Changes
* Add ReducerUnknownEvents annotation
* Don't fail if an event is not reduced
* Add a new BoosterUnknownEntity 

## Checks
- [X] Project Builds


packages/framework-core/src/booster-event-dispatcher.ts
```diff
@@ -89,7 +92,18 @@ export class BoosterEventDispatcher {
         logger.debug(`No event-handlers found for event ${eventEnvelope.typeName}. Skipping...`)
         continue
       }
-      const eventClass = config.events[eventEnvelope.typeName]
+      const eventClass = BoosterEventDispatcher.eventMetadata(eventEnvelope, config)
+      if (!eventClass) {
+        throw new NotFoundError(
```
charlietfe, line 97: @gonzalojaubert  Does it make sense to use the global dispatcher here to track this error?

packages/framework-core/src/services/event-store.ts
```diff
@@ -206,4 +227,24 @@ export class EventStore {
...
     }
   }
+
+  private eventMetadata(eventEnvelope: EventEnvelope): EventMetadata | UnknownEventMetadata | undefined {
+    const event = this.config.events[eventEnvelope.typeName]
+    if (event) return event
+    return this.config.unknownEvent
+  }
+
+  private isUnknownEventMetadata(
```
charlietfe, line 237: I think we have the same method declared before in `booster-event-dispatcher.ts`. If is the same method maybe we should move it to a common place.

packages/framework-core/src/booster-register-handler.ts
```diff
@@ -62,15 +58,25 @@ export class RegisterHandler {
     return boosterEventsTypesNames.includes(eventTypeName) ? BOOSTER_SUPER_KIND : DOMAIN_SUPER_KIND
   }
 
-  private static getEntityTypeName(
+  private static reducerClassName(
```
charlietfe, line 61: A better naming could be `getReducerClassNameForEntity`,just leave it or take it.

charlietfe (approved):
lgtm, just some minor comments.

PR #1247, opened 2022-11-25
Query annotation

gonzalojaubert:
## Description
Add `@Query` annotation

## Changes

* New annotation
* New GlobalErrorHandler
* Update GraphQL

## Checks
- [X] Project Builds


packages/framework-core/src/booster-authorizer.ts
```diff
@@ -15,6 +26,20 @@ export class BoosterAuthorizer {
     }
     return BoosterAuthorizer.denyAccess()
   }
+
+  public static build(
```
charlietfe, line 30: 💯

packages/framework-types/src/concepts/authorizers.ts
```diff
@@ -1,8 +1,10 @@
-import { EventSearchRequest, ReadModelRequestEnvelope, UserEnvelope, CommandEnvelope } from '../envelope'
+import { EventSearchRequest, ReadModelRequestEnvelope, UserEnvelope, CommandEnvelope, QueryEnvelope } from "../envelope";
```
charlietfe, line 1: remove `;` and replace double quotes by single quotes.

charlietfe (approved):
lgtm

packages/framework-types/src/concepts/authorizers.ts
gonzalojaubert: It looks like my code is the one without the `;` and with single quotes. Isn't it?

docs/chapters/03_booster-architecture.md
```diff
@@ -795,3 +795,132 @@ Booster GraphQL API also provides support for real-time updates using subscripti
...
+## The query handler function
+
+Each query class must have a method called `handle`. This function is the command handler, and it will be called by the framework every time one instance of this query is submitted. Inside the handler you can run validations, return errors and query entities to make decisions.
+
+### Returning a value
+
+By default, the command handler function is generated with `Promise<void>` as a return type,
+and in consequence, it will return `true` when called through the GraphQL.
+
+> [!NOTE] `true` is returned because GraphQL lacks of a `void` type, and it represents that the execution was successful.
```
alvaroloes, line 877: Copy paste issue. I'm not sure if this section is needed because a query without a return value doesn't make sense. Maybe saying something along those lines would be ok

packages/cli/src/templates/query.stub
```diff
@@ -0,0 +1,18 @@
...
+})
+export class {{{ name }}} {
+  public constructor(
+    {{#fields}}
+    readonly {{{name}}}: {{{type}}},
+    {{/fields}}
+  ) {}
+
+  public static async handle(query: {{{ name }}} , register: Register): Promise<void> {
+    register.events( /* YOUR EVENT HERE */)
```
alvaroloes, line 16: The body should be other thing. Maybe just a comment saying "your query code here"

packages/framework-integration-tests/integration/provider-unaware/end-to-end/queries.integration.ts
```diff
@@ -0,0 +1,146 @@
...
+            CartWithRole
+          }
+        `,
+      })
+
+      expect(result).not.to.be.null
+      expect(result?.data?.CartWithRole).to.be.true
+    })
+  })
+})
```
alvaroloes, line 146: Great tests!

packages/framework-types/src/concepts/query.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface QueryInterface<TQuery = unknown, THandleResult = unknown> extends Class<TQuery> {
+  handle(query: TQuery): Promise<THandleResult>
+}
+
+export interface QueryMetadata<TCommand = unknown> {
+  readonly class: QueryInterface<TCommand>
+  readonly properties: Array<PropertyMetadata>
+  readonly methods: Array<PropertyMetadata>
+  readonly authorizer: CommandAuthorizer
+}
```
alvaroloes, line 14: As I said in a comment above, maybe we can leverage that a command and query are exaclty the same thing and reuse the types. Let me know what you think.

alvaroloes: BTW, you are using the CommandAuthorizer type, and it should be the QueryAuthorizer (again, if they are exactly the same, we could use the same type in both command and queries)

alvaroloes (commented):
Again, outstanding work @gonzalojaubert

packages/framework-types/src/concepts/authorizers.ts
gonzalojaubert: oh sorry! let me fix it

packages/cli/src/templates/query.stub
```diff
@@ -0,0 +1,18 @@
...
+  authorize: // Specify authorized roles here. Use 'all' to authorize anyone
+})
+export class {{{ name }}} {
+  public constructor(
+    {{#fields}}
+    readonly {{{name}}}: {{{type}}},
+    {{/fields}}
+  ) {}
+
+  public static async handle(query: {{{ name }}} , register: Register): Promise<void> {
```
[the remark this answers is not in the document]
gonzalojaubert: You are right, the handle method doesn't have the register parameter

packages/cli/src/templates/query.stub
gonzalojaubert: Yep, copy/paste issue

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -212,11 +228,11 @@ function toEventSearchRequest(args: EventSearchParameters, context: GraphQLResol
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-function toCommandEnvelope(commandName: string, value: any, context: GraphQLResolverContext): CommandEnvelope {
+function toEnvelope(typeName: string, value: any, context: GraphQLResolverContext): CommandEnvelope | QueryEnvelope {
```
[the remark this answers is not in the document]
gonzalojaubert: Not sure about it. Having one type for two uses with `Or` or `And` is a smell for me. The only thing the method is doing is a `as`. Thoughts?

packages/framework-core/src/services/graphql/query-helpers/graphql-generate-handled-fields.ts
```diff
@@ -0,0 +1,64 @@
+import { GraphQLFieldConfigMap, GraphQLNonNull } from 'graphql'
+import { ResolverBuilder, TargetTypeMetadata, TargetTypesMap } from '../common'
+import { TypeMetadata } from '@boostercloud/metadata-booster'
+import { GraphQLTypeInformer } from '../graphql-type-informer'
+
+export class GraphqlGenerateHandledFields {
```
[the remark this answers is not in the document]
gonzalojaubert: good point

packages/framework-types/src/concepts/query.ts
gonzalojaubert: You got me :) I don't have a strong opinion on this but for sure we should follow the same approach. Perhaps having a good name for both cases will help me.

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -119,14 +125,24 @@ export class GraphQLGenerator {
...
     }
   }
 
+  public static queriesResolverBuilder(
+    queryClass: AnyClass
+  ): GraphQLFieldResolver<unknown, GraphQLResolverContext, { input: unknown }> {
+    return async (parent, args, context) => {
+      const queryEnvelope = toEnvelope(queryClass.name, args.input, context) as QueryEnvelope
+      const result = await this.queriesDispatcher.dispatchQuery(queryEnvelope, context)
+      return result ?? true
```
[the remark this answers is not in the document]
gonzalojaubert: Right, it doesn't make sense

alvaroloes (approved):

PR #1253, opened 2022-12-01
Adopt a standard documentation tool (Docusaurus)

BorjaZarco:
## Description

Migrate the current documentation towards a standard documentation tool. We are going to adopt [docusuarus](https://docusaurus.io/) as that standard tool, given its popularity among others.

This is a partial migration. The content of the docs needs to be improved in order to provide a better experience. Plus, the blog and demo sections need to be migrated to this new tool too. 

Until we fully migrate from the current docs system to docusaurus, we need to support both systems. That's why it was necessary to add a new folder called `website` to have the new system set up. 

## Changes

- Added a new folder `website` containing the new documentation system
- Migrate the docs to the new system

## Checks
- [X] Project Builds


adrian-lorenzo (approved):
Awesome! :D

NickSeagull (approved):
LGTM

PR #1263, opened 2022-12-06
[Issue #1196] Change default behaviour of new:project

hannahbrooks:
## Description
This PR updates the`new:project` command default behaviour. Instead of passing the `--default` flag, the command will use all defaults unless otherwise specified by the user in the initial command. The user must also now pass the `--providerPackageName` flag if not using the newly created `--interactive` flag.

## Changes

- Adds new `--interactive` flag
- Changes behaviour to automatically use defaults instead of using `--default` flag
- Update tests accordingly.

## Checks
- [x] Project Builds

[Github Issue #1196](https://github.com/boostercloud/booster/issues/1196)

Currently, the test coverage has dropped, which is not ideal


packages/cli/src/commands/new/project.ts
```diff
@@ -126,54 +126,58 @@ export const parseConfig = async (
...
+      homepage,
+      license,
+      repository,
+      boosterVersion,
+      installDependencies: flags.installDependencies || false,
+      initializeGit: flags.initializeGit || false,
+      interactive: true,
+    }) 
+  } else {
+    const providerPackageName = await getProviderPackageName(prompter, flags.providerPackageName)
```
javiertoledo, line 164: I think that in non-interactive mode we should fail and show an error message requesting the user to manually set the provider using a new `--provider` flag or instructing them to use the `--interactive` mode, something like:

```sh
$ boost new:project
ERROR: You must set a provider runtime package using the `--provider` flag or use the interactive mode with the `--interactive` flag.

Examples:

    boost new:project --provider @boostercloud/framework-provider-aws
    
  or

    boost new:project --interactive # This will show an interactive prompt that will guide you through the setup of your project.
    
<Additionally, print here the full --help message>
```

packages/cli/src/commands/new/project.ts
```diff
@@ -82,8 +82,8 @@ const run = async (flags: Partial<ProjectInitializerConfig>, boosterVersion: str
   Script.init(`boost ${Brand.energize('new')} 🚧`, parseConfig(new Prompter(), flags, boosterVersion))
     .step('Creating project root', generateRootDirectory)
     .step('Generating config files', generateConfigFiles)
-    .optionalStep(Boolean(flags.skipInstall), 'Installing dependencies', installDependencies)
-    .optionalStep(Boolean(flags.skipGit), 'Initializing git repository', initializeGit)
+    .optionalStep(Boolean(!flags.installDependencies), 'Installing dependencies', installDependencies)
+    .optionalStep(Boolean(!flags.initializeGit), 'Initializing git repository', initializeGit)
```
javiertoledo, line 86: Now that I see this change I wonder if it would make more sense for someone using the default behavior to install the packages and install git for them...

As a new user, what would you expect?

packages/cli/src/commands/new/project.ts
```diff
@@ -126,54 +126,58 @@ export const parseConfig = async (
...
+      initializeGit: flags.initializeGit || false,
+      interactive: true,
+    }) 
+  } else {
+    const providerPackageName = await getProviderPackageName(prompter, flags.providerPackageName)
+
     return Promise.resolve({
       projectName: flags.projectName as string,
-      providerPackageName: '@boostercloud/framework-provider-aws',
+      providerPackageName: providerPackageName,
```
javiertoledo, line 168: I think that it would also make sense to add a `--package-manager` flag (and the corresponding interactive step) to set up the package manager, as we now support `npm`, `yarn` and `pnpm`, but this can be done as a separate issue.

packages/cli/src/commands/new/project.ts
hannahbrooks: okay gotcha!

packages/cli/src/commands/new/project.ts
hannahbrooks: this makes sense

packages/cli/src/commands/new/project.ts
hannahbrooks: i agree: i think being able to start writing code right away is the better way to go. you would eventually have to do those steps yourself anyway, so might as well do them during install

hannahbrooks: i have made these changes, let me know what you think!

packages/cli/src/commands/new/project.ts
hannahbrooks: i added a small error message. how should I improve it? is there another way to send a message other than `throw`?

packages/cli/test/services/project-initializer.test.ts
```diff
@@ -35,9 +35,6 @@ describe('project initializer', (): void => {
     repository: '',
     providerPackageName: defaultProvider,
     boosterVersion: '0.5.1',
-    default: true,
-    skipInstall: false,
-    skipGit: false,
```
javiertoledo, line 40: It looks like these changes were left from previous commits.

```suggestion
    skipInstall: false,
    skipGit: false,
    interactive: false,
```

javiertoledo: It could make sense to properly tag the assigned variable type to make the compiler detect type mismatches.

```typescript
const defaultProjectInitializerConfig: ProjectInitializerConfig = {
  ...
}
```

Instead of casting the type with the `as` at the end of the assignment.

packages/framework-integration-tests/integration/provider-unaware/cli/cli.project.integration.ts
```diff
@@ -252,9 +252,6 @@ describe('Project', () => {
           `-p "${PROVIDER}"`,
           `-r "${REPO_URL}"`,
           `-v "${VERSION}"`,
-          // We skip dependencies and git installation to make this test faster
-          '--skipInstall',
-          '--skipGit',
```
javiertoledo, line 257: I think that it still makes sense to keep this too

packages/framework-integration-tests/integration/provider-unaware/cli/cli.project.integration.ts
```diff
@@ -271,9 +268,6 @@ describe('Project', () => {
           `--providerPackageName "${PROVIDER}"`,
           `--repository "${REPO_URL}"`,
           `--version "${VERSION}"`,
-          // We skip dependencies and git installation to make this test faster
-          '--skipInstall',
-          '--skipGit',
```
javiertoledo, line 276: Same here

packages/framework-integration-tests/integration/provider-unaware/cli/cli.project.integration.ts
```diff
@@ -338,9 +332,8 @@ describe('Project', () => {
           license: LICENSE,
           repository: REPO_URL,
           provider: 'default', // Just "hit enter" to choose the default one
-        }
-        // We skip dependencies and git installation to make this test faster
-        const output = await execNewProject(projectName, ['--skipInstall', '--skipGit'], promptAnswers)
+        }  
+        const output = await execNewProject(projectName, [], promptAnswers)
```
javiertoledo, line 336: Same here

packages/framework-integration-tests/integration/provider-unaware/cli/cli.project.integration.ts
```diff
@@ -356,8 +349,7 @@ describe('Project', () => {
           repository: REPO_URL,
           provider: PROVIDER,
         }
-        // We skip dependencies and git installation to make this test faster
-        const output = await execNewProject(projectName, ['--skipInstall', '--skipGit'], promptAnswers)
+        const output = await execNewProject(projectName, [], promptAnswers)
```
javiertoledo, line 352: Same here

packages/framework-integration-tests/integration/provider-unaware/cli/cli.project.integration.ts
```diff
@@ -376,9 +368,6 @@ describe('Project', () => {
         const flags = [
           `--providerPackageName "${PROVIDER}"`,
           `--repository "${REPO_URL}"`,
-          // We skip dependencies and git installation to make this test faster
-          '--skipInstall',
-          '--skipGit',
```
javiertoledo, line 381: another one 😅

packages/framework-integration-tests/integration/provider-unaware/cli/cli.project.integration.ts
```diff
@@ -410,9 +399,6 @@ describe('Project', () => {
           '--providerPackageName "invalid-provider"',
           `--repository "${REPO_URL}"`,
           `--version "${VERSION}"`,
-          // We skip dependencies and git installation to make this test faster
-          '--skipInstall',
-          '--skipGit',
```
javiertoledo, line 415: More here

hannahbrooks:
documentation also needs to be updated. should that be done in this commit?

javiertoledo:
Yes, updating the docs as part of this change would be ideal

hannahbrooks:
okay, if dev is looking okay, i can get on documentation updates tomorrow hopefully!

docs/chapters/02_getting-started.md
```diff
@@ -373,11 +373,10 @@ After choosing your provider, you will see your project generated!:
 ℹ Project generated!
 ```
 
-> [!TIP] If you prefer to create the project with default parameters, you can run the command as `boost new:project booster-blog --default`. The default
+> [!TIP] If you prefer to create the project with default parameters, you can run the command as `boost new:project booster-blog --providerPackageName <package-name>`. The package name of the provider is the only required parameter in project creatio. The default
```
javiertoledo, line 376:
```suggestion
> [!TIP] If you prefer to create the project with default parameters, you can run the command `boost new:project booster-blog --providerPackageName <package-name>`. The package name of the provider is the only required parameter and can be any `npm` package that implements Booster's provider interfaces. The framework provides the following: 

* `@boostercloud/framework-provider-aws`: Runs your app in AWS Lambda and stores the events, entity snapshots, and read models in DynamoDB.
* `@boostercloud/framework-provider-azure`: Runs your app in Azure Functions and stores the events, entity snapshots, and read models in CosmosDB.
* `@boostercloud/framework-provider-kubernetes` (Experimental): This package is a proof of concept that runs a Booster application on a Kubernetes cluster and Redis as the database.

The default
```

javiertoledo (approved):
Looks great, thanks for the great work! I just added a suggestion for the documentation. I'll start running the checks in preparation for the merge.

docs/chapters/02_getting-started.md
NickSeagull: I think that the DynamoDB, Cosmos, Lambda, etc... is too low-level for the getting started section. I'd remove that, specially with the K8s provider, given it is deprecated now 😅

packages/cli/src/commands/new/project.ts
```diff
@@ -58,6 +58,10 @@ export default class Project extends Command {
       description: 'skip git initialization',
       default: false,
     }),
+    interactive: flags.boolean({
+      description: 'choose options rather than defaults',
```
NickSeagull, line 62: Maybe...?
```suggestion
      description: 'run an interactive wizard to configure your project',
```

packages/cli/src/commands/new/project.ts
```diff
@@ -126,54 +132,54 @@ export const parseConfig = async (
...
+      author,
+      homepage,
+      license,
+      repository,
+      boosterVersion,
+      skipInstall: flags.skipInstall || false,
+      skipGit: flags.skipGit || false,
+      interactive: true,
+    })
+  } else {
```
NickSeagull, line 169: Given that the function is async we can actually just return the object as is. Could you change the return in the `else` too please? 🙏

NickSeagull (changes requested):
Looking good so far, left a couple of suggestions!

packages/cli/src/commands/new/project.ts
hannahbrooks: yes this makes more sense. my original description was quite literal haha

docs/chapters/02_getting-started.md
```diff
@@ -373,14 +373,11 @@ After choosing your provider, you will see your project generated!:
...
+> [!TIP] If you prefer to create the project with default parameters, you can run the command as `boost new:project booster-blog --providerPackageName <package-name>`. The package name of the provider is the only required parameter and can be any `npm` package that implements Booster's provider interfaces. The framework provides the following: 
 >
-> - Project name: The one provided when running the command, in this case "booster-blog"
-> - Provider: AWS
-> - Description, author, homepage and repository: ""
-> - License: MIT
-> - Version: 0.1.0
+> - `@boostercloud/framework-provider-aws`: Runs your app in AWS Lambda and stores the events and entity snapshots.
+> - `@boostercloud/framework-provider-azure`: Runs your app in Azure Functions and stores the events and entity snapshots.
+> - `@boostercloud/framework-provider-kubernetes` (Experimental): This package is a proof of concept that runs a Booster application on a Kubernetes cluster.
```
javiertoledo, line 380: If we don't want to mention the specific technologies, this might be easier for a beginner:

```suggestion
> - `@boostercloud/framework-provider-aws`: Runs your app in AWS.
> - `@boostercloud/framework-provider-azure`: Runs your app in Azure.
> - `@boostercloud/framework-provider-kubernetes` (Experimental): Proof of concept that shows how to run your app on a Kubernetes cluster.
```

What do you think @hannahbrooks @NickSeagull ?

javiertoledo: That's explained in more depth in other sections of the docs, so I think there is no need to repeat it here. I initially suggested it as part of the explanation of what's stored in the database, but as we're not mentioning it anymore, I'm not sure that this extra bit provides valuable information. What do you think?

hannahbrooks: no i agree, i think it's probably best to keep the getting started section as simple as possible!

PR #1271, opened 2022-12-12
Integration branch for PR #1182

javiertoledo:
Integration branch for PR #1182


javiertoledo:
I don't understand why the changeset check is failing in this PR. It works flawlessly locally, there are no error messages and I see no reason for it to fail... Any idea @NickSeagull? 🤔

NickSeagull:
From what I'm seeing, it could be that the `rush change` is not being able to pull the branch to do the changeset verification 🤔 

Maybe its doing the `actions/checkout` without the provided github token?

PR #1273, opened 2022-12-13
fix(cli): Make error messages more readable

NickSeagull:
## Description

After 1.0, the CLI error messages became a bit more unreadable, because they were being stringified. This PR fixes that

## Changes

Reworked error bubbling as well as printing

## Checks
- [x] Project Builds

### DIFF
--- a/packages/cli/test/services/package-manager/rush.test.ts
+++ b/packages/cli/test/services/package-manager/rush.test.ts
@@ -1,5 +1,5 @@
 import { fake } from 'sinon'
-import { gen, Layer, unsafeRunEffect } from '@boostercloud/framework-types/dist/effect'
+import { Effect, gen, Layer, mapError, pipe, unsafeRunEffect } from '@boostercloud/framework-types/dist/effect'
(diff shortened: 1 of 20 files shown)


packages/cli/test/services/package-manager/rush.test.ts
```diff
@@ -10,6 +10,12 @@ import { PackageManagerService } from '../../../src/services/package-manager'
 const TestFileSystem = makeTestFileSystem()
 const TestProcess = makeTestProcess()
 
+const mapEffError = <R, A>(effect: Effect<R, { error: Error }, A>) =>
+  pipe(
+    effect,
+    mapError((e) => e.error)
+  )
```
BorjaZarco, line 17: This is being used in many places, should we declare it on a separate file and export it?

NickSeagull: Probably, but given its for tests right now and I didn't really wanted to make it explicit I decided to leave it like this, thanks!

PR #1310, opened 2023-01-20
Implemented a fallback to check the existence of a 'compile' command in the project package.json file for backwards compatibility

### LINKED ISSUE #1307: Implement backwards compatibility for projects generated with versions under v1.0.0
# Bug Report

## Current Behavior

Old projects (before v1.0.0) were using a package.json script called `compile` but the new version of the project uses the `build` script. The `compile` script is wired in these old projects' CI/CD processes, so switching to the new `build` script is a pain that is better to solve.

## Expected behavior

Old projects should keep working without having to manually update the CI/CD processes to use the new script name.

## Possible Solution

A fallback mechanism that tries the `compile` script first and then runs the new `build` script if the first one fails.

javiertoledo:
## Description

Now When Booster CLI commands need to build the user project, Booster will check if there's a "compile" script defined in the user's package.json file. If it exists, it will run it instead of the default "build" script.

This is a fallback to provide backwards compatibility with projects created with previous versions of Booster that used this "compile" script by default instead.

This PR closes #1307

## Checks
- [x] Project Builds
~Updated documentation accordingly~

### DIFF
--- a/packages/cli/src/services/package-manager/common.ts
+++ b/packages/cli/src/services/package-manager/common.ts
@@ -17,7 +17,7 @@ const ensureProjectDir = (processService: ProcessService, projectDirRef: Ref.Ref
   })
 
 /**
- * Checks that the script exists in the package.json file
+ * Checks if a script exists in the package.json file
  */
 const checkScriptExists = (processService: ProcessService, fileSystemService: FileSystemService, scriptName: string) =>
   gen(function* ($) {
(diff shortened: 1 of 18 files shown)


packages/cli/src/services/package-manager/common.ts
```diff
@@ -15,18 +16,52 @@ const ensureProjectDir = (processService: ProcessService, projectDirRef: Ref.Ref
...
+ */
+export const makeScopedBuild = (command: string, projectDirRef: Ref.Ref<string>) =>
+  gen(function* ($) {
+    const processService = yield* $(ProcessService)
+    const fileSystemService = yield* $(FileSystemService)
+    const scriptExists = yield* $(checkScriptExists(processService, fileSystemService, 'compile'))
+    if (scriptExists) {
+      return effectfulRun(processService, command, projectDirRef).bind(null, 'compile')
+    }
+    return effectfulRun(processService, command, projectDirRef).bind(null, 'build')
```
NickSeagull, line 64: Generally I try to avoid `if` statements in effectful code, not because of anything special, but because I believe everything in the same level of indentation makes it easier to read, as the syntax is already a bit noisy, why not:

```suggestion
    const script = scriptExists ? 'compile' : 'build'
    return effectfulRun(processService, command, projectDirRef).bind(null, script)
```

Also, effect tends to favor curried functions. In fact, most of it's API is like that, I'd say that its better to curry functions rather than using `bind`

packages/cli/src/services/package-manager/common.ts
```diff
@@ -15,18 +16,52 @@ const ensureProjectDir = (processService: ProcessService, projectDirRef: Ref.Ref
...
+  gen(function* ($) {
+    const { cwd } = processService
+    const { readFileContents } = fileSystemService
+    const pwd = yield* $(cwd())
+    const packageJson = yield* $(readFileContents(`${pwd}/package.json`))
+    const packageJsonContents = JSON.parse(packageJson)
+    if (!packageJsonContents.scripts || !packageJsonContents.scripts[scriptName]) {
+      return false
+    }
+    return true
```
NickSeagull, line 32: Why not 
```suggestion
    return !packageJsonContents.scripts || !packageJsonContents.scripts[scriptName]
```

NickSeagull (commented):
Lovely!

packages/cli/src/services/package-manager/common.ts
javiertoledo: I tried to remove the `bind`, but that made the use of this function less clear in other places and tests, so better leaving it there for now 😅

charlietfe (approved):
Looks good!

PR #1318, opened 2023-01-25
Fixed VSCode launch.json commands to work with the Rush monorepo structure

javiertoledo:
I have updated the launch.json scripts to use the new rush commands from VSCode to be able to use the integrated debugger


NickSeagull (approved):
❤️

Issue #1254, opened 2022-12-01
Commands start failing after an exception is raised from an entity reducer


javiertoledo:
# Bug Report

Originally posted by @damienvieira [in Discord](https://discord.com/channels/763753198388510780/1047870803863142450/1047870803863142450)

## Current Behavior

I was doing some stress testing on Booster project, and I think I found a bug which causes an inconsistent state in an object entity. When this happens I can't perform any commands on that object.

More details:
Hundreds of questions are created in a short amount of time by calling a command CreateQuestion from an iOS simulator. 
This command is only creating a new UUID and assign it to this new event called  TopicQuestionCreated.

```typescript
const questionID = UUID.generate()
const createdAt = Date.now()

const eventQuestionCreated = new TopicQuestionCreated(command.topicID, questionID, createdAt)
register.events(eventQuestionCreated)
```

Then when the event  TopicQuestionCreated is reduced, it checks if a question ID already belongs to an array of questionsIDs for that topic, and if it does it raises an exception in the reducer.
You can check bellow the reducer for the event and where the exception is being raised.
 throw QuestionablyErrors.questionAlreadyExists

```typescript
  @Reduces(TopicQuestionCreated)
  public static reduceTopicQuestionCreated(event: TopicQuestionCreated, currentTopic?: Topic): Topic {
    if (!currentTopic) {
      throw QuestionablyErrors.topicNotFound
    }

    let questionIDs = currentTopic.questionIDs

    if (questionIDs.find((id) => id === event.questionID)) {
      throw QuestionablyErrors.questionAlreadyExists
    }

    questionIDs.push(event.questionID)
    
    return {
      ...currentTopic,
      updatedAt: event.updatedAt,
      questionIDs: questionIDs
    }
  }
```

So after around 300 questions created, I can see that error being raised, and from that point I can not perform any command on that topic object. If I replace that line of code where it throws the error with return currentTopic, everything works correctly, which I wouldn't expect to behave this way.

You can see the screenshot with the error logs.
![BoosterError](https://user-images.githubusercontent.com/175096/205092031-db9ea4df-f2d4-48a5-91a3-ca0f083af5b1.png)

PR #1331, opened 2023-02-07
Separated event from snapshot storage and started using the value `snapshottedEventCreatedAt` in snapshots to initialize the database sort key.

javiertoledo:
## Description

TL;DR. This change makes it possible to build the correct state of an entity starting from an arbitrary snapshot in the database, making the process of snapshot generation idempotent, and allowing snapshot writes from competing threads without inducing risks of missing events.

In the current codebase, events and entity snapshots are ordered in the database by a timestamp calculated at the time of persisting them. This makes Booster apps especially sensible to insertion order, making it harder to properly cache snapshots at read time, and forcing us to limit snapshot creation to the events dispatcher. While this approach has shown great resiliency, it comes at the cost of performance. When a certain number of commands and event handlers need to access the most recent version of an entity to validate further actions, and there's no recent snapshot, all of them have to recalculate the last state from the event source, making the event dispatcher process slower.

This behavior has been changed in #1116, where @gonzalogarciajaubert pointed out about a potential issue caused by the way we were setting up the snapshots sort keys. This PR fixes that potential issue with a rework on how events and snapshots are stored and ordered in the database.

> Notice that this change has been worked on top of the changes in #1116, so they're related and must be merged in order.

## Checks
- [x] Project Builds
~- [ ] Updated documentation accordingly~

## Additional checklist (Scenarios to add to the automated tests)

- [x] An exception while reducing events from a specific entity does no longer affect the reduction of other entities (#1254)
- [x] Event handlers are processed in parallel with read model updates
- [x] An error reducing an entity halts any future reduction for the failing entity (until someone fixes it)
- [x] An error thrown in an event handler does not stop other event handlers
- [x] When a snapshot already exists in the database, trying to rewrite it fails silently (writing it again is not needed)
- [x] Once a failing entity is fixed, the process restarts where it was left no matter how much time passed
- [x] Event handlers no longer wait for snapshots and read models to be calculated, they are processed in parallel, and a failure in one of the threads does not affect the other
- [ ] This change increases performance enough to augment the current load test limits

### DIFF
--- a/packages/framework-core/src/booster-event-dispatcher.ts
+++ b/packages/framework-core/src/booster-event-dispatcher.ts
@@ -36,6 +36,10 @@ export class BoosterEventDispatcher {
     }
   }
 
+  /**
+   * Builds a function that will be called once for each entity from the `RawEventsParser.streamPerEntityEvents` method
+   * after the page of events is grouped by entity.
+   */
   private static eventProcessor(eventStore: EventStore, readModelStore: ReadModelStore): EventsStreamingCallback {
     return async (entityName, entityID, eventEnvelopes, config) => {
       await Promise.allSettled([
@@ -59,7 +63,12 @@ export class BoosterEventDispatcher {
     readModelStore: ReadModelStore
   ): Promise<void> {
     const logger = getLogger(config, 'BoosterEventDispatcher#snapshotAndUpdateReadModels')
-    const entitySnapshot = await eventStore.fetchEntitySnapshot(entityName, entityID)
+    let entitySnapshot = undefined
+    try {
+      entitySnapshot = await eventStore.fetchEntitySnapshot(entityName, entityID)
+    } catch (e) {
+      logger.error('Error while fetching or reducing entity snapshot:', e)
+    }
     if (!entitySnapshot) {
       logger.debug('No new snapshot generated, skipping read models projection')
       return
--- a/packages/framework-core/src/booster-event-dispatcher.ts
+++ b/packages/framework-core/src/booster-event-dispatcher.ts
@@ -38,9 +38,16 @@ export class BoosterEventDispatcher {
 
   private static eventProcessor(eventStore: EventStore, readModelStore: ReadModelStore): EventsStreamingCallback {
     return async (entityName, entityID, eventEnvelopes, config) => {
-      // TODO: Separate the snapshot creation/read-model generation from the event handling into two independent processes.
-      await BoosterEventDispatcher.snapshotAndUpdateReadModels(config, entityName, entityID, eventStore, readModelStore)
-      await BoosterEventDispatcher.dispatchEntityEventsToEventHandlers(eventEnvelopes, config)
+      await Promise.allSettled([
+        await BoosterEventDispatcher.snapshotAndUpdateReadModels(
+          config,
+          entityName,
+          entityID,
+          eventStore,
+          readModelStore
+        ),
+        await BoosterEventDispatcher.dispatchEntityEventsToEventHandlers(eventEnvelopes, config),
+      ])
     }
   }
 
--- a/packages/framework-provider-aws/src/library/events-adapter.ts
+++ b/packages/framework-provider-aws/src/library/events-adapter.ts
@@ -3,12 +3,14 @@ import { DynamoDBRecord, DynamoDBStreamEvent } from 'aws-lambda'
 import {
   BoosterConfig,
   EventEnvelope,
+  EntitySnapshotEnvelope,
   OptimisticConcurrencyUnexpectedVersionError,
   UUID,
+  NonPersistedEventEnvelope,
 } from '@boostercloud/framework-types'
 import { DynamoDB } from 'aws-sdk'
 import { eventsStoreAttributes } from '../constants'
-import { partitionKeyForEvent, partitionKeyForIndexByEntity } from './keys-helper'
+import { partitionKeyForEntitySnapshot, partitionKeyForEvent, partitionKeyForIndexByEntity } from './keys-helper'
 import { Converter } from 'aws-sdk/clients/dynamodb'
 import { getLogger, retryIfError } from '@boostercloud/framework-common-helpers'
 
@@ -57,7 +59,7 @@ export async function readEntityLatestSnapshot(
   config: BoosterConfig,
   entityTypeName: string,
   entityID: UUID
-): Promise<EventEnvelope | null> {
+): Promise<EntitySnapshotEnvelope | undefined> {
   const logger = getLogger(config, 'EventsAdapter#readEntityLatestSnapshot')
   const result = await dynamoDB
     .query({
@@ -65,7 +67,7 @@ export async function readEntityLatestSnapshot(
       ConsistentRead: true,
       KeyConditionExpression: `${eventsStoreAttributes.partitionKey} = :partitionKey`,
       ExpressionAttributeValues: {
-        ':partitionKey': partitionKeyForEvent(entityTypeName, entityID, 'snapshot'),
+        ':partitionKey': partitionKeyForEntitySnapshot(entityTypeName, entityID),
       },
       ScanIndexForward: false, // Descending order (newer timestamps first),
       Limit: 1,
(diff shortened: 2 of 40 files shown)


packages/framework-types/src/provider.ts
```diff
@@ -35,16 +37,22 @@ export interface ProviderEventsLibrary {
...
     limit: number,
     afterCursor: Record<string, string> | undefined,
     entityTypeName: string
   ): Promise<PaginatedEntitiesIdsResult>
-  /** Streams an event to the corresponding event handler */
-  store(eventEnvelopes: Array<EventEnvelope>, config: BoosterConfig): Promise<void>
+  /** Stores a list of events in the events store */
+  store(eventEnvelopes: Array<NonPersistedEventEnvelope>, config: BoosterConfig): Promise<void>
+  /** Stores an entity snapshot in the event store */
+  storeSnapshot(snapshotEnvelope: EntitySnapshotEnvelope, config: BoosterConfig): Promise<void>
```
javiertoledo, line 55: This PR introduces a dedicated method in the `ProviderLibrary` interface to store snapshots.

packages/framework-types/src/envelope.ts
```diff
@@ -26,18 +26,28 @@ export interface ScheduledCommandEnvelope extends Envelope {
...
+}
+export interface EventEnvelope extends NonPersistedEventEnvelope {
+  persistedAt: string
+}
+
+export interface EntitySnapshotEnvelope extends EventStoreEntryEnvelope {
+  kind: 'snapshot'
+  snapshottedEventCreatedAt: string
+  snapshottedEventPersistedAt: string
+}
```
javiertoledo, line 50: There's a new type semantics for event store entries:
* `NonPersistedEventEnvelope`: An event that has not been stored yet
* `EventEnvelope`: An event that has been stored (it has a `persistedAt` timestamp set)
* `EntitySnapshotEnvelope`: A dedicated snapshot envelope that forces the presence of `snapshottedEventCreatedAt` and `snapshottedEventPersistedAt`. These fields reference the latest event processed for the current snapshot.

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -101,28 +103,78 @@ export async function storeEvents(
...
+        TableName: config.resourceNames.eventsStore,
+        ConditionExpression: `${eventsStoreAttributes.partitionKey} <> :partitionKey AND ${eventsStoreAttributes.sortKey} <> :sortKey`,
+        ExpressionAttributeValues: {
+          ':partitionKey': partitionKey,
+          ':sortKey': sortKey,
+        },
+        Item: {
+          ...snapshotEnvelope,
+          [eventsStoreAttributes.partitionKey]: partitionKey,
+          [eventsStoreAttributes.sortKey]: sortKey,
```
javiertoledo, line 136: One detail here: While we're introducing a new `snapshottedEventPersistedAt` field and using it as the sort key value, the `eventsStoreAttributes.sortKey` constant is still `createdAt`. This means that at the time of writing, the value stored will override whatever value we set in `createdAt` and replace It with the value set in `snapshottedEventPersistedAt`. I think that this is important to keep backward compatibility, but I'm totally open to other thoughts on this.

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -101,28 +103,78 @@ export async function storeEvents(
...
         ExpressionAttributeValues: {
           ':partitionKey': partitionKey,
-          ':sortKey': sortKey,
+          ':sortKey': persistableEvent.persistedAt,
         },
         Item: {
           ...eventEnvelope,
           [eventsStoreAttributes.partitionKey]: partitionKey,
-          [eventsStoreAttributes.sortKey]: sortKey,
+          [eventsStoreAttributes.sortKey]: persistableEvent.persistedAt,
```
javiertoledo, line 177: Same here: Doing this we're overriding the `createdAt` that comes in the `eventEnvelope`by the calculated `persistedAt` value (see line 164 in this same file)

packages/framework-provider-azure/src/library/events-adapter.ts
```diff
@@ -72,35 +78,69 @@ export async function readEntityLatestSnapshot(
...
-        ...eventEnvelope,
+        ...persistableEvent,
         [eventsStoreAttributes.partitionKey]: partitionKeyForEvent(
           eventEnvelope.entityTypeName,
-          eventEnvelope.entityID,
-          eventEnvelope.kind
+          eventEnvelope.entityID
         ),
-        [eventsStoreAttributes.sortKey]: new Date().toISOString(),
+        [eventsStoreAttributes.sortKey]: persistableEvent.persistedAt,
```
javiertoledo, line 111: Same as in AWS, we're keeping the same sort key in the database to ease upgrades and backward compatibility

packages/framework-core/src/booster-event-dispatcher.ts
```diff
@@ -38,9 +38,16 @@ export class BoosterEventDispatcher {
...
+      await Promise.allSettled([
+        await BoosterEventDispatcher.snapshotAndUpdateReadModels(
+          config,
+          entityName,
+          entityID,
+          eventStore,
+          readModelStore
+        ),
+        await BoosterEventDispatcher.dispatchEntityEventsToEventHandlers(eventEnvelopes, config),
+      ])
```
javiertoledo, line 50: With the changes introduced in this PR, and given that at this point all events are 100% persisted and settled in the database, it does no longer matter the order in which we calculate the read models and process event handlers, so we can now process them in parallel.

packages/framework-provider-aws/src/library/events-adapter.ts
```diff
@@ -101,28 +103,78 @@ export async function storeEvents(
...
-  eventEnvelope: EventEnvelope
+  eventEnvelope: NonPersistedEventEnvelope
 ): Promise<void> {
   try {
-    const partitionKey = partitionKeyForEvent(eventEnvelope.entityTypeName, eventEnvelope.entityID, eventEnvelope.kind)
+    const partitionKey = partitionKeyForEvent(eventEnvelope.entityTypeName, eventEnvelope.entityID)
     // Generate a new timestamp as sorting key on every try until we can persist the event.
     // This way we guarantee ordering even with events that are stored in the same millisecond
-    const sortKey = new Date().toISOString()
+    const persistableEvent: EventEnvelope = {
```
gonzalogarciajaubert, line 162: I think this object is never used

packages/framework-provider-aws/src/library/events-adapter.ts
gonzalogarciajaubert: Not sure if we will need it or not, but we are not going to persist a timestamp with the snapshot creation, are we?

packages/framework-provider-azure/src/library/events-adapter.ts
```diff
@@ -72,35 +78,69 @@ export async function readEntityLatestSnapshot(
...
+  await cosmosDb
+    .database(config.resourceNames.applicationStack)
+    .container(config.resourceNames.eventsStore)
+    .items.create({
+      ...snapshotEnvelope,
+      [eventsStoreAttributes.partitionKey]: partitionKey,
+      [eventsStoreAttributes.sortKey]: sortKey,
+    })
+
+  logger.debug('[EventsAdapter#storeSnapshot] Snapshot stored')
```
gonzalogarciajaubert, line 145: Could we improve this log?

```suggestion
  logger.debug('Snapshot stored', snapshotEnvelope)
```

packages/framework-provider-aws/src/library/events-adapter.ts
javiertoledo: I rolled back the introduction of the `persistedAt` field to use the `createdAt` field again, but I'm still updating the `createdAt` to match the one from the entity. As we're now generating snapshots from multiple threads, this is the only way to know if a specific snapshot has been generated before, but it could make sense to also keep an extra timestamp with the snapshot persist time. I'll add that field here.

javiertoledo:
Merging this branch into #1116 to start working on **the merge™**

PR #1334, opened 2023-02-13
async rocket dispatcher

gonzalojaubert:
Fix missing async on rocket dispatcher

### DIFF
--- a/packages/framework-core/src/booster-rocket-dispatcher.ts
+++ b/packages/framework-core/src/booster-rocket-dispatcher.ts
@@ -3,7 +3,7 @@ import { BoosterConfig, rocketFunctionIDEnvVar } from '@boostercloud/framework-t
 export class BoosterRocketDispatcher {
   constructor(readonly config: BoosterConfig) {}
 
-  public dispatch(request: unknown): Promise<unknown> {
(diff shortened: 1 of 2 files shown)


NickSeagull (approved):
If this works, LGTM. How does this work tho? 😅

charlietfe (approved):

packages/framework-core/src/booster-rocket-dispatcher.ts
```diff
@@ -3,7 +3,7 @@ import { BoosterConfig, rocketFunctionIDEnvVar } from '@boostercloud/framework-t
 export class BoosterRocketDispatcher {
   constructor(readonly config: BoosterConfig) {}
 
-  public dispatch(request: unknown): Promise<unknown> {
+  public async dispatch(request: unknown): Promise<unknown> {
```
alvaroloes, line 6: 🤔 If I have seen everything correctly, this change is only adding the async keyword, but it is now needed, as the code inside the function doesn't have any `await`. 

You are already returning a promise so, unless you want to use await inside the function, you don't need to add async. 

Was this because of some bug?

gonzalojaubert: You are right, it's not needed

PR #1347, opened 2023-02-18
Azure no longer described as experimental

javiertoledo:
## Description

It's been a while since many people are using the Azure implementation with no issues, so this PR puts it in the place it deserved, as a fully supported provider.

## Changes

Removed all the misleading notes claiming that Azure support is experimental


NickSeagull (approved):
Approved, note however that it won't release a version with this because you specified a `none` version bump

javiertoledo:
Oh, good catch, it might be worth releasing it so the CLI no longer display the experimental tag during project creation. I'll regenerate the changes file to do a patch bump

PR #1348, opened 2023-02-21
Update Rockets documentation

damienvieira:
## Changes
Added a more detailed documentation regarding existing rockets.

## Checks
- [x] Project Builds


jfsagasti:
I understand the point of having inline docs for better UX. But in this case that they are a duplication of the rocket READMEs, I think the downsides could be higher than the benefits of having a list of links. I would briefly explain what the rocket is for and then the link to its page where it is well explained and how to be used. Otherwise, we are just duplicating a lot of info with a high chance of being outdated (and we need twice the work to keep it in sync, 2 PRs for every change). The only benefit now is to not opening the Github page of the rocket (but you will probably need anyway). As a developer, I don't think it brings much more value to be honest, we are really used to Github. Rockets are also an advanced use case and not part of the basics.

On the other hand, if we are adding AI and we want it to answer rocket questions, having all the info here could be easier (although we can train, it with the readmes too, without adding duplication here anyway).

What are your thoughts @adrian-lorenzo @javiertoledo?

damienvieira:
In my opinion, we should remove the code and examples from the Readme file on each rocket. Instead, we should provide a brief introduction that explains what the rocket is about and include a link to our Booster documentation, where the content is properly formatted and easy to read. 

This guarantees that we have a single centralised location for the documentation of the rocket, which is what most libraries usually do on GitHub.

jfsagasti:
@damienvieira, that would be an AMAZING improvement! And the way to go, too, in my opinion :)

adrian-lorenzo (approved):

javiertoledo:
I think that it could make sense to maintain the documentation pages of the rockets in their respective repositories. I think that we could import these pages from the docusaurus app as part of the documentation build process. That way we have to maintain a single source of truth that is closer to the rocket, but we can render the documentation in the place where the users look for it. The rockets are not well standardized at the moment because they've all been developed in a bit hack-ish way by different people and we've been solving problems as they appeared, so maybe we should invest a bit of time on defining a project template for rockets where we document how to create a rocket, how to document it, how to design it for multicloud environments, etc. I think that could improve the way we manage and explain them by a lot!

javiertoledo (approved):

jfsagasti (approved):

PR #1350, opened 2023-02-22
Fix docs sidebar so it shows notification docs

NickSeagull:


website/sidebars.js
```diff
@@ -26,6 +26,7 @@ const sidebars = {
         'architecture/event-handler',
         'architecture/entity',
         'architecture/read-model',
+        'architecture/notifications',
```
NickSeagull, line 29: this is the change

PR #1358, opened 2023-02-24
Add clarification about Data Migrations and updated example

jfsagasti:


BorjaZarco (approved):
LGTM! 👍

website/docs/10_going-deeper/data-migrations.md
```diff
@@ -8,7 +8,7 @@ Migrations are a mechanism for updating or transforming the schemas of events an
 
 * **Schema migrations** are used to incrementally upgrade an event or entity from a past version to the next. They are applied lazily, meaning that they are performed on-the-fly whenever an event or entity is loaded. This allows you to make changes to your data model without having to manually update all existing artifacts, and makes it possible to apply changes without running lenghty migration processes.
 
-* **Data migrations**, on the other hand, behave as background processes that can actively change the existing values in the database for existing entities and read models. They are particularly useful for data migrations that cannot be performed automatically with schema migrations, or for updating existing read models after a schema change.
+* **Data migrations**, on the other hand, behave as background processes that can actively change the existing values in the database for existing entities and read models. They are particularly useful for data migrations that cannot be performed automatically with schema migrations, or for updating existing read models after a schema change. **Note that Data migrations are only available in the Azure provider at the moment.**
```
javiertoledo, line 11: Let's double check that this is true before merging this change, from what I know this feature was implemented in a vendor-agnostic way 🤔

javiertoledo (changes requested):

website/docs/10_going-deeper/data-migrations.md
javiertoledo: According to the conversations we've been having about this, I'm suggesting a couple of edits:

```suggestion
* **Data migrations**, on the other hand, behave as background processes that can actively change the existing values in the database for existing entities and read models. They are particularly useful for data migrations that cannot be performed automatically with schema migrations, or for updating existing read models after a schema change.
```

website/docs/10_going-deeper/data-migrations.md
```diff
@@ -106,7 +106,9 @@ Data migrations are not run automatically, you need to invoke the `BoosterDataMi
...
 
-Inside your `@DataMigration` classes, you can use the `Booster.migrateEntity` method to update the data for a specific entity. This method takes the old entity name, the old entity ID, and the new entity data as arguments. It will also generate an internal `BoosterEntityMigrated` event before performing the migration.
+Inside your `@DataMigration` classes, you can use the `BoosterDataMigrations.migrateEntity` method to update the data for a specific entity. This method takes the old entity name, the old entity ID, and the new entity data as arguments. It will also generate an internal `BoosterEntityMigrated` event before performing the migration.
+
+**Note that Data migrations are only available in the Azure provider at the moment.**
 
 Here is an example of how you might use the `@DataMigration` decorator and the `Booster.migrateEntity` method to update the quantity of the first item in a cart:
```
javiertoledo, line 113:
```suggestion
Here is an example of how you might use the `@DataMigration` decorator and the `Booster.migrateEntity` method to update the quantity of the first item in a cart (**Notice that at the time of writing this document, the method `Booster.entitiesIDs` used in the following example is only available in the Azure provider, so you may need to approach the migration differently in AWS.**):
```

javiertoledo (changes requested):
@jfsagasti I've added a couple of suggestions according to our earlier conversations.

javiertoledo (approved):

PR #1377, opened 2023-03-15
Add regex and iRegex filters to Azure and Local Provider

gonzalojaubert:
## Description
Add `regex` and `iRegex` filters to Azure and Local Provider

## Changes
* `regex` allow to query using regular expressions
* `iRegex` allow to query using case insensitive regular expressions 

## Checks
- [X] Project Builds

### DIFF
--- a/packages/framework-core/src/services/graphql/query-helpers/graphql-query-filter-arguments-builder.ts
+++ b/packages/framework-core/src/services/graphql/query-helpers/graphql-query-filter-arguments-builder.ts
@@ -140,6 +140,8 @@ export class GraphqlQueryFilterArgumentsBuilder {
         in: { type: GraphQLList(GraphQLString) },
         beginsWith: { type: GraphQLString },
         contains: { type: GraphQLString },
+        regex: { type: GraphQLString },
+        iRegex: { type: GraphQLString },
         isDefined: { type: GraphQLBoolean },
       }
     // use `name`, `typeGroup` === 'Class'
--- a/packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
+++ b/packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
@@ -795,7 +795,7 @@ describe('Read models end-to-end tests', () => {
         expect(cartData[0].payment.id).to.be.eq(mockPaymentId)
       })
 
-      it('should retrieve ´a list of carts when filter by contains with UUID fields', async () => {
+      it('should retrieve a list of carts when filter by contains with UUID fields', async () => {
         const partialMockCartId = mockCartId.slice(1, -1)
         const filter = {
           and: [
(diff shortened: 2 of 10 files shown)


packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
```diff
@@ -795,6 +795,119 @@ describe('Read models end-to-end tests', () => {
...
+            {
+              shippingAddress: {
+                firstName: {
+                  regex: `^${mockAddress.firstName.at(0)}.*`,
+                },
+              },
+            },
+          ],
+        }
+        const queryResult = await waitForIt(
```
charlietfe, line 819: Super nit-picky, why not reuse in the tests a function to get the `ListCartReadModels` passing only the different filters `regex` or `iRegex`?

packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
```diff
@@ -795,6 +795,119 @@ describe('Read models end-to-end tests', () => {
...
+            {
+              shippingAddress: {
+                firstName: {
+                  iRegex: `^${inverseFirstLetter}.*`,
+                },
+              },
+            },
+          ],
+        }
+        const queryResult = await waitForIt(
```
charlietfe, line 877: Take a look at the above comment about reuse.

charlietfe (approved):
LGTM, nit-picky comments

packages/framework-core/src/services/graphql/query-helpers/graphql-query-filter-arguments-builder.ts
```diff
@@ -140,6 +140,8 @@ export class GraphqlQueryFilterArgumentsBuilder {
         in: { type: GraphQLList(GraphQLString) },
         beginsWith: { type: GraphQLString },
         contains: { type: GraphQLString },
+        regex: { type: GraphQLString },
+        iRegex: { type: GraphQLString },
```
NickSeagull, line 144: what does `iRegex` mean?

NickSeagull (changes requested):

packages/framework-core/src/services/graphql/query-helpers/graphql-query-filter-arguments-builder.ts
gonzalojaubert: it's like `ILIKE` in SQL. In this case is case insensitive regex

NickSeagull (approved):

packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-models.integration.ts
alvaroloes: I agree here. It would make super easy to test this more thoroughly by just changing the regex expression

alvaroloes (approved):
Just one minor comment. No blocker

PR #1389, opened 2023-03-27
Add Tracer

gonzalojaubert:
## Description
Add tracer decorator for observability

## Checks
- [X] Project Builds


charlietfe (approved):

packages/framework-types/src/instrumentation/trace-types.ts
```diff
@@ -0,0 +1,55 @@
+import { UUID } from '../concepts'
+import { BoosterConfig } from '../config'
+
+export enum TraceTypes {
+  START,
+  END,
+}
+
+export enum TraceActionTypes {
+  CUSTOM,
```
alvaroloes, line 10: Suggestion: It would be great if the user could use any string in the `@Trace` decorator to be able to define their own actions. This way, they can enable the trace for specific methods of their project. Right now, you only have the ability to enable or disable traces for internal actions. In your project it is all or nothing.

packages/framework-core/src/instrumentation/decorator/trace.ts
```diff
@@ -0,0 +1,58 @@
...
+function buildParameters(
+  target: unknown,
+  member: string,
+  args: unknown[],
+  description: string | undefined,
+  descriptor: PropertyDescriptor
+) {
+  const config = Booster.config
+  let internal = undefined
+  if (config && config.traceConfiguration.includeInternal) {
```
alvaroloes, line 36: Couldn't `traceConfiguration` be undefined?

packages/framework-core/src/instrumentation/decorator/trace.ts
```diff
@@ -0,0 +1,58 @@
...
+import { isTracerConfigured, notifyTrace } from '../trace-notifier'
+import { Booster } from '../../booster'
+
+export function Trace(actionType: TraceActionTypes = TraceActionTypes.CUSTOM, description?: string) {
+  return (target: unknown, member: string, descriptor: PropertyDescriptor): PropertyDescriptor => {
+    const originalMethod = descriptor.value
+    descriptor.value = async function (...args: Array<unknown>) {
+      const tracerConfigured = isTracerConfigured(actionType)
+      if (!tracerConfigured) {
+        return await originalMethod.apply(this, args)
```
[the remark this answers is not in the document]
alvaroloes: Could you please a test to verify this? If the wrong this is used, it could break a lot of code

website/docs/10_going-deep/instrumentation.md
```diff
@@ -0,0 +1,138 @@
+# Booster instrumentation
+
+## Trace Decorator
+The Trace Decorator is a **Booster** functionality that facilitates the reception of notifications whenever significant events occur in Booster's core, such as event dispatching or migration execution.
+
+### Usage
+To configure a custom tracer, you need to define an object with two methods: onStart and onEnd. The onStart method is called before the traced method is invoked, and the onEnd method is called after the method completes. Both methods receive a TraceInfo object, which contains information about the traced method and its arguments.
```
alvaroloes, line 7: You eventually called the type `TraceParameters`, instead of TraceInfo (which, BTW, I like more)

website/docs/10_going-deep/instrumentation.md
```diff
@@ -0,0 +1,138 @@
...
+import { MyTracer } from './my-tracer'
+
+const myTracer = new MyTracer()
+const config: BoosterConfig = {
+// ...other configuration options...
+  trace: {
+    enableTraceNotification: true,
+    verbose: false,
+    onStart: await myTracer.onStart,
+    onEnd: await myTracer.onStart,
```
alvaroloes, line 42: These awaits are wrong here

website/docs/10_going-deep/instrumentation.md
```diff
@@ -0,0 +1,138 @@
...
+```typescript
+import { BoosterConfig } from '@boostercloud/framework-types'
+import { MyTracer } from './my-tracer'
+
+const myTracer = new MyTracer()
+const config: BoosterConfig = {
+// ...other configuration options...
+  trace: {
+    enableTraceNotification: true,
+    verbose: false,
```
alvaroloes, line 40: Verbose? I think this is not in the configuration

website/docs/10_going-deep/instrumentation.md
```diff
@@ -0,0 +1,138 @@
...
+    onStart: await myTracer.onStart,
+    onEnd: await myTracer.onStart,
+  }
+}
+```
+
+In the configuration above, we've enabled trace notifications and specified our onStart and onEnd as the methods to use. Verbose disable will reduce the amount of information generated excluding the internal parameter in the response.
```
alvaroloes, line 47: Same here, we are talking about "Verbose". Maybe you wanted to say `includeInternal`? Also you say `excluding the internal parameter in the response`, What is the "internal parameter"? Also, I guess you wanted to say in the `trace parameters` not `in the response`, correct?

Anyway, I don't think there is a need to exclude information in the trace parameters. It can always be there, it doesn't hurt

website/docs/10_going-deep/instrumentation.md
```diff
@@ -0,0 +1,138 @@
...
+    verbose: false,
+    onStart: await myTracer.onStart,
+    onEnd: await myTracer.onStart,
+  }
+}
+```
+
+In the configuration above, we've enabled trace notifications and specified our onStart and onEnd as the methods to use. Verbose disable will reduce the amount of information generated excluding the internal parameter in the response. 
+
+By default, all traceable actions are disabled, but you can enable all of them or selectively enable only specific actions using an array of TraceActionTypes.
```
alvaroloes, line 49: Suggestion. Feel free to take it or leave it
```suggestion
Setting `enableTraceNotification: true` would enable the trace for all actions. You can either disable them by setting it to `false` or selectively enable only specific actions using an array of TraceActionTypes.
```

website/docs/10_going-deep/instrumentation.md
```diff
@@ -0,0 +1,138 @@
...
+  elapsedInvocationMillis?: number
+  internal: {
+    target: unknown
+    descriptor: PropertyDescriptor
+  }
+  description?: string
+}
+```
+
+className and methodName identify the function that is being traced.
```
alvaroloes, line 116:
```suggestion
`className` and `methodName` identify the function that is being traced.
```

website/docs/10_going-deep/instrumentation.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Adding the Trace Decorator to Your Own Methods
+In addition to using the Trace Decorator to receive notifications when events occur in Booster's core, you can also use it to trace your own methods. To add the Trace Decorator to your own methods, simply add @Trace() before your method declaration.
```
alvaroloes, line 119: If we eventually allow the user to use their own action names, this would need to change

packages/framework-core/src/instrumentation/decorator/trace.ts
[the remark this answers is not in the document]
gonzalojaubert: In this context, `this` is the descriptor so it's the expected one. I' added a test to verify it

packages/framework-core/src/instrumentation/decorator/trace.ts
```diff
@@ -0,0 +1,58 @@
...
+  descriptor: PropertyDescriptor
+) {
+  const config = Booster.config
+  let internal = undefined
+  if (config && config.traceConfiguration.includeInternal) {
+    internal = {
+      target: target,
+      descriptor: descriptor,
+    }
+  }
```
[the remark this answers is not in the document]
gonzalojaubert: Because it's a lot of information and not useful in a lot of cases. I explain it better in the following comments

packages/framework-core/src/read-model-schema-migrator.ts
```diff
@@ -3,12 +3,15 @@ import {
...
   ReadModelInterface,
+  TraceActionTypes,
 } from '@boostercloud/framework-types'
 import { getLogger } from '@boostercloud/framework-common-helpers'
+import { Trace } from './instrumentation'
 
 export class ReadModelSchemaMigrator {
   public constructor(private config: BoosterConfig) {}
 
+  @Trace(TraceActionTypes.READ_MODEL_SCHEMA_MIGRATOR_RUN)
```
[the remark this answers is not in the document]
gonzalojaubert: Good catch

packages/framework-core/src/instrumentation/trace-notifier.ts
```diff
@@ -0,0 +1,31 @@
...
+  type: TraceTypes,
+  actionType: TraceActionTypes,
+  parameters: TraceParameters
+): Promise<void> {
+  const config = Booster.config
+  const handler = type === TraceTypes.START ? config.traceConfiguration.onStart : config.traceConfiguration.onEnd
+  return handler(config, actionType, parameters)
+}
+
+export function isTracerConfigured(actionType: TraceActionTypes): boolean {
```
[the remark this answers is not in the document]
gonzalojaubert: Sure

packages/framework-core/src/instrumentation/decorator/trace.ts
gonzalojaubert: No, the `traceConfiguration` is set in the framework-types config class and it doesn't accepts undefined.

packages/framework-types/src/instrumentation/trace-types.ts
```diff
@@ -0,0 +1,55 @@
...
+  elapsedInvocationMillis?: number
+  internal?: {
+    target: unknown
+    descriptor: PropertyDescriptor
+  }
+  description?: string
+}
+
+export interface traceHandler {
+  (config: BoosterConfig, actionType: TraceActionTypes, traceParameters: TraceParameters): void
```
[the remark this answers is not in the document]
gonzalojaubert: Right!

website/docs/10_going-deep/instrumentation.md
gonzalojaubert: ok :)

website/docs/10_going-deep/instrumentation.md
gonzalojaubert: ouch!

website/docs/10_going-deep/instrumentation.md
gonzalojaubert: it is `includeInternal`. I renamed it and forgot the docs

website/docs/10_going-deep/instrumentation.md
gonzalojaubert: Yes, it's `includeInternal`. The `internal` parameter is in the documentation:

```
  internal: {
    target: unknown
    descriptor: PropertyDescriptor
  }
```

Regarding this parameter I think it's very useful as the methods (onStart and onEnd) are receiving an object with all the information. If you log the object (to console, file, etc...) and don’t exclude that property you may run into problems because of the size of the object. If the destination is a file you may end up having space problems for example. So it is very useful to give that option to the user.

website/docs/10_going-deep/instrumentation.md
gonzalojaubert: love it, thanks!

website/docs/10_going-deep/instrumentation.md
```diff
@@ -0,0 +1,138 @@
...
+import { MyTracer } from './my-tracer'
+
+const myTracer = new MyTracer()
+const config: BoosterConfig = {
+// ...other configuration options...
+  trace: {
+    enableTraceNotification: [TraceActionTypes.DISPATCH_EVENT, TraceActionTypes.MIGRATION_RUN],
+    verbose: false,
+    onStart: await myTracer.onStart,
+    onEnd: await myTracer.onStart,
```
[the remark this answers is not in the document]
gonzalojaubert: yeah

packages/framework-core/src/instrumentation/decorator/trace.ts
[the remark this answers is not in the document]
alvaroloes: You are correct! What complexity there is around the "this" in Javascript 🙄

alvaroloes (approved):

PR #1394, opened 2023-04-11
Improve global error handler

gonzalojaubert:
## Description
<TODO>

## Changes

* Add ScheduleCommandHandler metadata
* Add EventHandler metadata
* Add Reducer metadata
* Add Projection metadata
* Add EventHandler envelope
* Add Reducer envelope
* Add Projection envelope

## Checks
- [X] Project Builds


packages/framework-core/src/booster-global-error-dispatcher.ts
```diff
@@ -78,20 +82,31 @@ export class BoosterGlobalErrorDispatcher {
...
   private async handleEventHandlerError(error: GlobalErrorContainer): Promise<Error | undefined> {
     if (!this.errorHandler || !this.errorHandler.onDispatchEventHandlerError) throw error.originalError
     const currentError = error as EventHandlerGlobalError
-    return await this.errorHandler.onDispatchEventHandlerError(currentError.originalError, currentError.eventInstance)
+    return await this.errorHandler.onDispatchEventHandlerError(
+      currentError.originalError,
+      currentError.eventEnvelope,
+      currentError.eventHandlerMetadata,
+      currentError.eventInstance
+    )
```
davidverdu, line 100: I wonder if we could just pass `currentError: Error` as the only argument for this function, since all params are now being taken from the same object. On the other hand, this could make tests or other parts of the code more difficult, so we need to think about it

packages/framework-core/src/booster-global-error-dispatcher.ts
```diff
@@ -78,20 +82,31 @@ export class BoosterGlobalErrorDispatcher {
...
-  private async handleReducerError(error: GlobalErrorContainer): Promise<Error | undefined> {
+  private async handleReducerError(error: GlobalErrorContainer): Promise<Error> {
     if (!this.errorHandler || !this.errorHandler.onReducerError) throw error.originalError
     const currentError = error as ReducerGlobalError
     return await this.errorHandler.onReducerError(
       currentError.originalError,
+      currentError.eventEnvelope,
+      currentError.reducerMetadata,
       currentError.eventInstance,
       currentError.snapshotInstance
```
davidverdu, line 111: I wonder if we could just pass `currentError: ReducerGlobalError` as the only argument for this function, since all params are now being taken from the same object. On the other hand, this could make tests or other parts of the code more difficult, so we need to think about it

packages/framework-integration-tests/src/common/app-error-handler.ts
```diff
@@ -22,15 +29,21 @@ import {
...
 export class AppErrorHandler {
-  public static async onCommandHandlerError(error: Error, command: CommandEnvelope): Promise<Error | undefined> {
-    if (command.value.cartId === commandHandlerErrorIgnoredCartId) {
+  public static async onCommandHandlerError(
+    error: Error,
+    commandEnvelope: CommandEnvelope,
+    commandMetadata: CommandMetadata
+  ): Promise<Error | undefined> {
+    console.log(commandEnvelope)
+    console.log(commandMetadata)
```
davidverdu, line 38: not sure if should remove these console.log

davidverdu, line 38: not sure if we should remove these console.log

packages/framework-integration-tests/src/common/app-error-handler.ts
```diff
@@ -42,41 +55,65 @@ export class AppErrorHandler {
...
-  public static async onScheduledCommandHandlerError(error: Error): Promise<Error | undefined> {
+  public static async onScheduledCommandHandlerError(
+    error: Error,
+    scheduledCommandEnvelope: ScheduledCommandEnvelope,
+    scheduledCommandMetadata: ScheduledCommandMetadata
+  ): Promise<Error | undefined> {
     console.log('onScheduledCommandHandlerError')
-    return error
+    console.log(scheduledCommandEnvelope)
+    console.log(scheduledCommandMetadata)
```
davidverdu, line 65: not sure if we should remove these console.log

packages/framework-integration-tests/src/common/app-error-handler.ts
```diff
@@ -42,41 +55,65 @@ export class AppErrorHandler {
...
     error: Error,
+    eventEnvelope: EventEnvelope | NotificationInterface,
+    eventHandlerMetadata: unknown,
     eventInstance: EventInterface
   ): Promise<Error | undefined> {
-    const getEntityID = eventInstance?.entityID ?? (() => '')
-    if (getEntityID() === dispatchEventErrorCartId) {
-      return new Error(error.message + '-onDispatchEventHandlerError')
+    console.log(eventEnvelope)
+    console.log(eventHandlerMetadata)
```
davidverdu, line 78: not sure if we should remove these console.log

packages/framework-integration-tests/src/common/app-error-handler.ts
```diff
@@ -42,41 +55,65 @@ export class AppErrorHandler {
...
     eventInstance: EventInterface,
     snapshotInstance: EntityInterface | null
-  ): Promise<Error | undefined> {
-    const getEntityID = eventInstance?.entityID ?? (() => '')
-    if (getEntityID() === reducerErrorCartId) {
-      return new Error(error.message + '-onReducerError')
+  ): Promise<Error> {
+    console.log(eventEnvelope)
+    console.log(reducerMetadata)
+    console.log(snapshotInstance)
```
davidverdu, line 97: not sure if we should remove these console.log

packages/framework-integration-tests/src/common/app-error-handler.ts
```diff
@@ -42,41 +55,65 @@ export class AppErrorHandler {
...
   public static async onProjectionError(
     error: Error,
+    entityEnvelope: EntitySnapshotEnvelope,
+    projectionMetadata: ProjectionMetadata<EntityInterface, ReadModelInterface>,
     entity: EntityInterface,
     readModel: ReadModelInterface | undefined
   ): Promise<Error | undefined> {
+    console.log(entityEnvelope)
+    console.log(projectionMetadata)
+    console.log(readModel)
```
davidverdu, line 114: not sure if we should remove these console.log

packages/framework-types/src/errors/event-handler-global-error.ts
```diff
@@ -1,8 +1,14 @@
 import { GlobalErrorContainer } from './global-error-container'
-import { EventInterface } from '../concepts'
+import { EventInterface, NotificationInterface } from '../concepts'
+import { EventEnvelope } from '../envelope'
 
 export class EventHandlerGlobalError extends GlobalErrorContainer {
-  constructor(readonly eventInstance: EventInterface, originalError: Error) {
+  constructor(
+    readonly eventEnvelope: EventEnvelope | NotificationInterface,
```
davidverdu, line 7: why `eventEnvelope` can be either `EventEnvelope` or `NotificationInterface` and not one single type?

packages/framework-integration-tests/src/common/app-error-handler.ts
davidverdu: never mind, this will only be seen on integration tests execution

davidverdu (approved):
LGTM. I left some non-blocking comments

packages/framework-core/src/booster-global-error-dispatcher.ts
MarcAstr0: I think the point of the `@GlobalErrorHandler` decorator and the `GlobalErrorHandlerInterface` is to expand the current implementation of its methods instead of introducing new error types.

packages/framework-core/src/booster-global-error-dispatcher.ts
MarcAstr0: See [this comment](https://github.com/boostercloud/booster/pull/1394#discussion_r1706051878).

packages/framework-types/src/errors/event-handler-global-error.ts
MarcAstr0: I think @gonzalojaubert could provide a better answer. Still, I believe it's due to distinguishing between a regular event (consumed via the Cosmos change feed processor) and notifications when using the event stream functionality.

PR #1395, opened 2023-04-17
Add NonExposed parameter decorator

gonzalojaubert:
## Description
Add a decorator to remove a parameter from GraphQL Schema

## Checks
- [X] Project Builds


packages/framework-integration-tests/src/commands/change-cart-item.ts
```diff
@@ -18,7 +18,12 @@ import {
   before: [ChangeCartItem.beforeFn, ChangeCartItem.beforeFnV2],
 })
 export class ChangeCartItem {
-  public constructor(readonly cartId: UUID, readonly productId: UUID, readonly quantity: number) {}
+  public constructor(
+    readonly cartId: UUID,
+    readonly productId: UUID,
+    readonly quantity: number,
+    @ignore readonly test: number
+  ) {}
```
alvaroloes, line 26: Would the decorator also work if we define the properties at the class level, instead of in the constructor (something that is common)? I mean if we do this:
```typescript
export class ChangeCartItem {
    readonly cartId: UUID
    readonly productId: UUID
    readonly quantity: number
    @ignore readonly test: number

  // ...
```

I guess it won't work because you are using the field index in the constructor to know which field to ignore, but any of these fields are in a constructor. 
It probably works if you use the name of the field

packages/framework-core/src/services/graphql/common.ts
```diff
@@ -87,3 +87,10 @@ export const buildGraphqlSimpleEnumFor = (enumName: string, values: Array<string
     }, {} as GraphQLEnumValueConfigMap),
   })
 }
+
+export function nonExcludedFields(
+  fields: Array<PropertyMetadata>,
+  excludeProps?: Array<string>
+): Array<PropertyMetadata> {
+  return excludeProps ? fields.filter((field: PropertyMetadata) => !excludeProps.includes(field.name)) : [...fields]
```
alvaroloes, line 95: Nitpick: We can just return the same array instead of making a copy.
```suggestion
  return excludeProps ? fields.filter((field: PropertyMetadata) => !excludeProps.includes(field.name)) : fields
```

packages/framework-core/src/decorators/non-exposed.ts
```diff
@@ -0,0 +1,29 @@
...
+function getFieldName(
+  methodName: string | undefined,
+  target: AnyClass | InstanceType<AnyClass>,
+  parameterIndex: number | undefined
+): string {
+  if (methodName) {
+    return methodName
+  }
+  const argumentNames = getFunctionArguments(target)
+  return argumentNames[parameterIndex!]
```
alvaroloes, line 28: I think using `!` will throw if the value is undefined. I know that, supposedly, if `methodName` is undefined, then `parameterIndex` must not... but we can't be sure. My suggestion is to add a explicit throw here with a message that tells exactly that "we could not get field name information in {this class}, {this method name}, etc.". 

That way, we would know instantly what the error is about, instead of needing to debug something like `undefined can be used to index` or, worse, fail silently and having an array of "non exposed fields" all filled with undefined 😂

alvaroloes (approved):
Approved. I just left two code improvements

charlietfe (approved):

PR #1398, opened 2023-04-19
Add last update and last projected entity to read models

gonzalojaubert:
## Description
Add last update and last projected entity to read models

### DIFF
--- a/packages/framework-core/src/services/read-model-store.ts
+++ b/packages/framework-core/src/services/read-model-store.ts
@@ -132,6 +132,7 @@ export class ReadModelStore {
       ...migratedReadModel?.boosterMetadata,
       version: currentReadModelVersion + 1,
       schemaVersion: schemaVersion,
+      lastUpdateAt: new Date().toISOString(),
     }
     logger.debug(
       `[ReadModelStore#project] Storing new version of read model ${readModelName} with ID ${readModelID}:`,
--- a/packages/framework-core/src/services/read-model-store.ts
+++ b/packages/framework-core/src/services/read-model-store.ts
@@ -56,7 +56,8 @@ export class ReadModelStore {
                 projectionMetadata,
                 readModelName,
                 readModelID,
-                sequenceKey
+                sequenceKey,
+                entitySnapshotEnvelope.entityTypeName
               ),
             OptimisticConcurrencyUnexpectedVersionError,
             logger
@@ -94,7 +95,8 @@ export class ReadModelStore {
     projectionMetadata: ProjectionMetadata<EntityInterface>,
     readModelName: string,
     readModelID: UUID,
-    sequenceKey?: SequenceKey
+    sequenceKey?: SequenceKey,
+    entityTypeName?: string
   ): Promise<unknown> {
     const logger = getLogger(this.config, 'ReadModelStore#applyProjectionToReadModel')
     const readModel = await this.fetchReadModel(readModelName, readModelID, sequenceKey)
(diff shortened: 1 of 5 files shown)


packages/framework-core/src/services/read-model-store.ts
```diff
@@ -94,7 +95,8 @@ export class ReadModelStore {
     projectionMetadata: ProjectionMetadata<EntityInterface>,
     readModelName: string,
     readModelID: UUID,
-    sequenceKey?: SequenceKey
+    sequenceKey?: SequenceKey,
+    entityTypeName?: string
```
charlietfe, line 99: Super minor: Would it make sense to name this param as `lastProjectedEntityName` or something similar?

charlietfe (approved):
LGMT, minor comment leave it or take it.

packages/framework-core/src/services/read-model-store.ts
```diff
@@ -132,6 +134,8 @@ export class ReadModelStore {
       ...migratedReadModel?.boosterMetadata,
       version: currentReadModelVersion + 1,
       schemaVersion: schemaVersion,
+      lastUpdateAt: new Date().toISOString(),
+      lastProjectedEntity: entityTypeName,
```
alvaroloes, line 138: I've been thinking about this, and I think it would be ideal to leverage the fact that you are modifying this to add more information about the entity that can be super useful to debug problems. My suggestion is to change `lastProjectedEntity` by `lastProjectionInfo` and change it to an object with the following params
```typescript
lastProjectionInfo: {
  entityName: "..." // The name of the entity
  entityID: "..." // The ID of the entity
  entityUpdatedAt: entitySnapshotEnvelope.createdAt // Super useful, so we can know the value of the entity used in the last projection.
  projectionMethod: `${projectionMetadata.class.name}.${projectionMetadata.methodName}`
}
```

This way, we can know exactly which projection was last executed, for which entity, and its value (through the "entityUpdatedAt" parameter).

packages/framework-core/src/services/read-model-store.ts
```diff
@@ -94,7 +95,8 @@ export class ReadModelStore {
     projectionMetadata: ProjectionMetadata<EntityInterface>,
     readModelName: string,
     readModelID: UUID,
-    sequenceKey?: SequenceKey
+    sequenceKey?: SequenceKey,
+    entityTypeName?: string
   ): Promise<unknown> {
```
alvaroloes, line 100: In oder to have the information I proposed in the other comment, you would need to pass here the `entitySnapshotEnvelope`, instead of only the entityTypeName

packages/framework-core/src/services/read-model-store.ts
gonzalojaubert: Makes sense

packages/framework-core/src/services/read-model-store.ts
```diff
@@ -132,6 +134,13 @@ export class ReadModelStore {
       ...migratedReadModel?.boosterMetadata,
       version: currentReadModelVersion + 1,
       schemaVersion: schemaVersion,
+      lastUpdateAt: new Date().toISOString(),
+      lastProjectionInfo: {
+        entityId: entity.id,
+        entityName: lastProjectedEntity?.entityTypeName,
+        entityUpdateAt: lastProjectedEntity?.createdAt,
```
alvaroloes, line 141: Minor typo:
```suggestion
        entityUpdatedAt: lastProjectedEntity?.createdAt,
```

charlietfe (approved):

PR #1402, opened 2023-04-28
Create host.json file if not exist

gonzalojaubert:


JYCabello:
If I understand correctly, there's a change in the integration tests that takes this change into account, but neither a test verifying that the zip defaults nor one verifying that the provided one is used.

Is it feasible to unit test it?

charlietfe (approved):
LGTM!

alvaroloes (approved):

PR #1403, opened 2023-05-02
Update packages 02

JYCabello:
## Description
Updated GraphQL to the latest version of the library as part of an ongoing effort to update all dependencies and the runtime.

## Checks
- [x] Project Builds


packages/framework-core/src/services/graphql/common.ts
```diff
@@ -41,22 +42,26 @@ export interface GraphQLResolverContext {
   context?: ContextEnvelope
 }
 
-export const DateScalar = new GraphQLScalarType({
+function parseDate(inputValue: unknown): Date {
+  // TODO: Don't let this pass review.
```
javiertoledo, line 46: Not sure if this should be removed, pointing it out just in case 😜

javiertoledo (approved):
Great work here, much needed upgrade!

PR #1404, opened 2023-05-09
Add global event handler

gonzalojaubert:
## Description
Add a global event handler

## Checks
- [X] Project Builds


charlietfe (approved):
LGTM

JYCabello (approved):

packages/framework-core/src/booster-event-dispatcher.ts
```diff
@@ -87,26 +88,49 @@ export class BoosterEventDispatcher {
...
-          } catch (e) {
-            const globalErrorDispatcher = new BoosterGlobalErrorDispatcher(config)
-            const error = await globalErrorDispatcher.dispatch(new EventHandlerGlobalError(eventInstance, e))
-            if (error) throw error
-          }
-          return RegisterHandler.handle(config, register)
-        })
-      )
+      const eventInstance = this.getEventInstance(config, eventEnvelope)
+      if (eventHandlers && eventHandlers.length > 0) {
```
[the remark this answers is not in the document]
gonzalojaubert: Good idea

MarcAstr0:
About commit f9b730e: this commit fixes an issue introduced in [v2.11](https://github.com/boostercloud/booster/releases/tag/v2.11.0) (related to PR #1481).

The issue in question is changing Cosmos DB SQL queries so they use square brackets instead of dot notation for properties. i.e., a SQL query such as `SELECT c.id FROM c` is changed to `SELECT c["id"] FROM c`.

This prevents that Cosmos throws syntax errors when there are properties in the read model that are also reserved Cosmos DB keywords (e.g., `value`). This issue was found when running [this integration test](https://github.com/boostercloud/booster/pull/1404/files#diff-b3aeb7529e03c8fc7ef74c3cc840fd604e06fba14c4c650601ee9a0d1d3175b8R807).

packages/framework-core/src/decorators/global-event-handler.ts
```diff
@@ -0,0 +1,24 @@
...
+  eventHandlerClass: TEventHandler
+): void {
+  Booster.configureCurrentEnv((config: BoosterConfig): void => {
+    const registeredEventHandlers = config.eventHandlers[GLOBAL_EVENT_HANDLER] || []
+    if (registeredEventHandlers.length > 0) {
+      throw new Error(`A global event handler is already registered.
+        If you think that this is an error, try performing a clean build.`)
+    }
+    registeredEventHandlers.push(eventHandlerClass)
+    config.eventHandlers[GLOBAL_EVENT_HANDLER] = registeredEventHandlers // todo
```
alvaroloes, line 22: This code prevents having more than one global event handler in the application, and I think we shouldn't do that. It can be very useful to have different global error handler for different purposes (for example, one to send logs to some third party, another one to do some other tracking, etc.).

By allowing that, then this method can be deleted and reuse the existing one in `event-handler.ts`. See next comment.

packages/framework-core/src/decorators/global-event-handler.ts
```diff
@@ -0,0 +1,24 @@
+import { EventHandlerInterface, BoosterConfig } from '@boostercloud/framework-types'
+import { Booster } from '../booster'
+
+export const GLOBAL_EVENT_HANDLER = 'GLOBAL_EVENT_HANDLER'
+
+export function GlobalEventHandler(): <TEventHandler extends EventHandlerInterface>(
+  eventHandlerClass: TEventHandler
+) => void {
+  return (eventHandlerClass) => registerGlobalEventHandler(eventHandlerClass)
+}
```
alvaroloes, line 10: Two things here: 
1. This decorator doesn't receive anything as parameter, so it doesn't have to be a function that return the decorator. It can be the decorator itself.
2. Taking into account the previous comment, we can delete the function `registerGlobalEventHandler` and call the one that is in `event-handler.ts`. This way we avoid repeating the logic.

This is the suggested code with both changes applied:

```suggestion
export function GlobalEventHandler<TEventHandler extends EventHandlerInterface>(
  eventHandlerClass: TEventHandler
): void {
  // We would need to do the function `registerEventHandler` exported
  registerEventHandler(GLOBAL_EVENT_HANDLER, eventHandlerClass) 
}
```

packages/framework-core/src/decorators/global-event-handler.ts
```diff
@@ -0,0 +1,24 @@
+import { EventHandlerInterface, BoosterConfig } from '@boostercloud/framework-types'
+import { Booster } from '../booster'
+
+export const GLOBAL_EVENT_HANDLER = 'GLOBAL_EVENT_HANDLER'
```
alvaroloes, line 4: To minimize the chances of collision with a user defined event handler, I would prefix this with `BOOSTER`: 'BOOSTER_GLOBAL_EVENT_HANDLER'

Also, and this is being nitpicking, I would add an S to both the var name and value, because we are storing multiple event handlers

packages/framework-integration-tests/src/event-handlers/global-handler.ts
```diff
@@ -0,0 +1,13 @@
+import { EventInterface, NotificationInterface, Register } from '@boostercloud/framework-types'
+import { GlobalEventHandler } from '@boostercloud/framework-core'
+import { LogEventReceived } from '../events/log-event-received'
+import { LogEventReceivedTest } from '../events/log-event-received-test'
+
+@GlobalEventHandler()
```
alvaroloes, line 6: We would not need the () after the changes described above

website/docs/03_architecture/04_event-handler.mdx
```diff
@@ -100,3 +100,18 @@ export class HandleAvailability {
...
 }
 ```
+
+## Creating a global event handler
+
+**Booster** includes a `Global event handler`. This feature allows you to react to any event that occurs within the system.
+By annotating a class with the @GlobalEventHandler decorator, the handle method within that class will be automatically called for any event that is generated
+
+```typescript
+@GlobalEventHandler()
```
alvaroloes, line 110: Here we can remove the extra ()

MarcAstr0:
@alvaroloes your suggestions have been applied in commit 4b970ce91b45f61a52da045a7e46f5b06cc96f8c

packages/framework-core/src/decorators/global-event-handler.ts
```diff
@@ -0,0 +1,11 @@
+import { EventHandlerInterface } from '@boostercloud/framework-types'
+import { registerEventHandler } from './event-handler'
+
+export const BOOSTER_GLOBAL_EVENT_HANDLERS = 'BOOSTER_GLOBAL_EVENT_HANDLERS'
+
+export function GlobalEventHandler<TEventHandler extends EventHandlerInterface>(
+  eventHandlerClass: TEventHandler
+): void {
+  // We would need to do the function `registerEventHandler` exported
```
alvaroloes, line 9: No need to include this comment

alvaroloes (approved):
Approved! Check the unneeded comment

MarcAstr0:
Closes PR #1535

PR #1410, opened 2023-05-24
Return readModel instance when querying by id

hendrik-lemon:
## Description
When querying a ReadModel list, the ReadModels are returned as instances. When a ReadModel is queried by id, this is not the case. As a consequence, getters do not work when querying by id. 

## Changes
When querying a ReadModel by id, an instance is not returned so the getters work.

## Checks
- [ ] Project Builds


JYCabello (approved):

JYCabello:
@hendrik-lemon Run `rush change` in the repo and define the change. Make it a fix.

PR #1412, opened 2023-05-27
Disable subscriptions

gonzalojaubert:
Add config parameter to disable subscriptions artifacts to be created

### DIFF
--- a/packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/permissions.ts
+++ b/packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/permissions.ts
@@ -16,10 +16,13 @@ export const setupPermissions = (
   websocketAPI?: CfnApi,
   scheduledCommandStack?: ScheduledCommandStackMembers
 ): void => {
+  if (config.enableSubscriptions && !websocketAPI) {
+    throw new Error('WebsocketAPI undefined with enableSubscriptions enabled')
+  }
   const { graphQLLambda, subscriptionsStore, subscriptionNotifier, connectionsStore } = graphQLStack
   const { eventsLambda, eventsStore } = eventsStack
   const scheduledLambda = scheduledCommandStack?.scheduledLambda
-  if (config.enableSubscriptions && websocketAPI) {
+  if (websocketAPI) {
     const websocketManageConnectionsPolicy = createPolicyStatement(
       [
         Fn.join(':', [
--- a/packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/permissions.ts
+++ b/packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/permissions.ts
@@ -6,43 +6,51 @@ import { GraphQLStackMembers } from './graphql-stack'
 import { ScheduledCommandStackMembers } from './scheduled-commands-stack'
 import { EventsStackMembers } from './events-stack'
 import { Function } from '@aws-cdk/aws-lambda'
+import { BoosterConfig } from '@boostercloud/framework-types'
 
 export const setupPermissions = (
+  config: BoosterConfig,
   graphQLStack: GraphQLStackMembers,
   eventsStack: EventsStackMembers,
   readModelTables: Array<Table>,
-  websocketAPI: CfnApi,
+  websocketAPI?: CfnApi,
   scheduledCommandStack?: ScheduledCommandStackMembers
 ): void => {
   const { graphQLLambda, subscriptionsStore, subscriptionNotifier, connectionsStore } = graphQLStack
   const { eventsLambda, eventsStore } = eventsStack
   const scheduledLambda = scheduledCommandStack?.scheduledLambda
-  const websocketManageConnectionsPolicy = createPolicyStatement(
-    [
-      Fn.join(':', [
-        'arn',
-        Fn.ref('AWS::Partition'),
-        'execute-api',
-        Fn.ref('AWS::Region'),
-        Fn.ref('AWS::AccountId'),
-        `${websocketAPI.ref}/*`,
-      ]),
-    ],
-    ['execute-api:ManageConnections']
-  )
+  if (config.enableSubscriptions && websocketAPI) {
+    const websocketManageConnectionsPolicy = createPolicyStatement(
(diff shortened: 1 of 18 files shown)


packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/permissions.ts
```diff
@@ -6,43 +6,51 @@ import { GraphQLStackMembers } from './graphql-stack'
...
+          [connectionsStore.tableArn],
+          ['dynamodb:GetItem', 'dynamodb:PutItem', 'dynamodb:DeleteItem']
+        )
+      )
+      graphQLLambda.addToRolePolicy(websocketManageConnectionsPolicy)
+      // Subscription notifier lambda permissions
+      subscriptionNotifier.addToRolePolicy(createPolicyStatement([subscriptionsStore.tableArn], ['dynamodb:Query*']))
+      subscriptionNotifier.addToRolePolicy(websocketManageConnectionsPolicy)
+    }
+  }
```
alvaroloes, line 49: I think we can simplify the `ifs` by not asking for `config.enableSubscriptions` again, as, if it is false, all the other members related to subscriptions are undefined (the reason is to avoid duplicity in the logic to know whether the subscriptions are enabled or not). Something like:
```suggestion
  if (websocketAPI) {
    const websocketManageConnectionsPolicy = createPolicyStatement(
      [
        Fn.join(':', [
          'arn',
          Fn.ref('AWS::Partition'),
          'execute-api',
          Fn.ref('AWS::Region'),
          Fn.ref('AWS::AccountId'),
          `${websocketAPI.ref}/*`,
        ]),
      ],
      ['execute-api:ManageConnections']
    )
  }
  if (subscriptionsStore && connectionsStore && subscriptionNotifier) {
    grantFullAccessToSubscriptionsStore(subscriptionsStore, graphQLLambda)
    graphQLLambda.addToRolePolicy(
      createPolicyStatement(
        [connectionsStore.tableArn],
        ['dynamodb:GetItem', 'dynamodb:PutItem', 'dynamodb:DeleteItem']
      )
    )
    graphQLLambda.addToRolePolicy(websocketManageConnectionsPolicy)
    // Subscription notifier lambda permissions
    subscriptionNotifier.addToRolePolicy(createPolicyStatement([subscriptionsStore.tableArn], ['dynamodb:Query*']))
    subscriptionNotifier.addToRolePolicy(websocketManageConnectionsPolicy)
  }
}
```

packages/framework-provider-azure-infrastructure/src/infrastructure/synth/application-synth.ts
```diff
@@ -151,34 +154,38 @@ export class ApplicationSynth {
...
+          azurermProvider,
+          this.terraformStackResource,
+          resourceGroupResource,
+          webPubSubResource,
+          this.appPrefix,
+          functionAppResource,
+          functionAppDataResource,
+          hubName
+        )
+      }
```
alvaroloes, line 179: Same here. So the question is: Could `this.config.enableSubscriptions` be true, but `webPubSubResource` be undefined? I guess it could not, so I would write the code in a way that it is impossible to happen. What I mean is to just ask for `webPubSubResource` in the main if, and not for `this.config.enableSubscriptions`. 

My preference would be to ask always for the `this.config.enableSubscriptions`, but as the wePubSubResource can be undefined, then we would get a compilation error

alvaroloes: Or! Maybe this is a better idea for this comment and the previous one: Always ask for `this.config.enableSubscriptions`, and then throw an error if the required resources are undefined, like `webPubSubResource` in this case. 

This is closer to the truth as if we are a the situation where `this.config.enableSubscriptions` is true but `webPubSubResource` is undefined, then there was some error, and it is better to show it to the user instead of ignoring it

packages/framework-provider-aws-infrastructure/src/infrastructure/stacks/permissions.ts
alvaroloes: See next comment for what I believe is a better approach

packages/framework-provider-azure-infrastructure/src/infrastructure/synth/terraform-function-app.ts
```diff
@@ -27,7 +27,7 @@ export class TerraformFunctionApp {
       resourceGroupName: resourceGroupResource.name,
       servicePlanId: servicePlanResource.id,
       appSettings: {
-        WebPubSubConnectionString: webPubsubResource.primaryConnectionString,
+        WebPubSubConnectionString: webPubsubResource?.primaryConnectionString || '',
```
alvaroloes, line 30: Shouldn't we just leave this as it was so that `WebPubSubConnectionString` is undefined  (equivalent to not providing it) instead of the empty string?

website/docs/06_graphql.md
```diff
@@ -324,6 +324,11 @@ mutation {
 Remember that, in case you want to subscribe to a read model that is restricted to a specific set of roles, you must send the **access token** retrieved upon sign-in. Check ["Authorizing operations"](#authorizing-operations) to know how to do this.
 :::
 
+:::note
+You can disable the creation of the artifacts needed to manage subscriptions by using `config.enableSubscriptions=false`.
```
alvaroloes, line 328: Not sure whether the word `artifact` conveys the desired meaning here (I associate it to packages or dependencies, maybe it is only me)

My suggestion is to change it to something like this:
```suggestion
You can disable the creation of all the infrastructure and functionality needed to manage subscriptions by setting `config.enableSubscriptions=false` in your `Booster.config` block.
```

packages/framework-provider-azure-infrastructure/src/infrastructure/synth/terraform-function-app.ts
gonzalojaubert: WebPubSubConnectionString doesn't accept undefined

alvaroloes (approved):

charlietfe (approved):

PR #1414, opened 2023-06-06
Add Local Provider optimistic concurrency support

gonzalojaubert:
Add Local Provider optimistic concurrency support

### DIFF
--- a/packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-model-concurrency.integration.ts
+++ b/packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-model-concurrency.integration.ts
@@ -1,0 +1,99 @@
+import { ApolloClient, gql, NormalizedCacheObject } from '@apollo/client'
+import { random } from 'faker'
+import { expect } from '../../helper/expect'
+import { applicationUnderTest } from './setup'
+import 'mocha'
+import { ReadModelInterface, UUID } from '@boostercloud/framework-types'
+import { waitForIt } from '../../helper/sleep'
+
+describe('Concurrency end-to-end tests', () => {
+  let client: ApolloClient<NormalizedCacheObject>
+
+  before(async () => {
+    client = applicationUnderTest.graphql.client()
+  })
+
+  context('ReadModels', () => {
+    describe('With one projection', () => {
+      it('insert and update generate one ReadModel with version 2', async () => {
+        const entityId: UUID = random.uuid()
+
+        const insertedReadModel = await addConcurrency(client, entityId, 1, 'ConcurrencyReadModel')
+        expect(insertedReadModel.id).to.be.eq(entityId)
+        expect(insertedReadModel.boosterMetadata?.version).to.be.eq(1)
+
+        const updatedReadModel = await addConcurrency(client, entityId, 2, 'ConcurrencyReadModel')
+        expect(updatedReadModel.id).to.be.eq(entityId)
+        expect(updatedReadModel.boosterMetadata?.version).to.be.eq(2)
+      })
+    })
+
+    describe('With two projections for the same ReadModel', () => {
+      if (process.env.TESTED_PROVIDER === 'AWS') {
+        console.log('AWS Provider is not working properly when inserting a ReadModel with two projections') // TODO: Fix AWS Provider
+        return
+      }
+      it('insert and update generate one ReadModel with version 4', async () => {
+        const entityId: UUID = random.uuid()
+
+        const insertedReadModel = await addConcurrency(client, entityId, 2, 'OtherConcurrencyReadModel')
+        expect(insertedReadModel.id).to.be.eq(entityId)
+        expect(insertedReadModel.otherId).to.be.eq(entityId)
+        expect(insertedReadModel.boosterMetadata?.version).to.be.eq(2)
+
+        const updatedReadModel = await addConcurrency(client, entityId, 4, 'OtherConcurrencyReadModel')
+        expect(updatedReadModel.id).to.be.eq(entityId)
+        expect(updatedReadModel.otherId).to.be.eq(entityId)
+        expect(updatedReadModel.boosterMetadata?.version).to.be.eq(4)
+      })
+    })
+  })
+})
+
+async function addConcurrency(
+  client: ApolloClient<NormalizedCacheObject>,
+  entityId: UUID,
+  expectedVersion: number,
+  readModelName: string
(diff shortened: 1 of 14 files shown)


packages/framework-integration-tests/integration/provider-unaware/end-to-end/read-model-concurrency.integration.ts
```diff
@@ -0,0 +1,99 @@
...
+        const updatedReadModel = await addConcurrency(client, entityId, 2, 'ConcurrencyReadModel')
+        expect(updatedReadModel.id).to.be.eq(entityId)
+        expect(updatedReadModel.boosterMetadata?.version).to.be.eq(2)
+      })
+    })
+
+    describe('With two projections for the same ReadModel', () => {
+      if (process.env.TESTED_PROVIDER === 'AWS') {
+        console.log('AWS Provider is not working properly when inserting a ReadModel with two projections') // TODO: Fix AWS Provider
+        return
```
alvaroloes, line 34: We should create an issue to track this and add details

packages/framework-provider-local/src/services/read-model-registry.ts
```diff
@@ -32,27 +37,67 @@ export class ReadModelRegistry {
     const queryPromise = new Promise((resolve, reject) =>
       cursor.exec((err, docs) => {
         if (err) reject(err)
-        else resolve(docs)
+        else {
+          docs.map((doc: ReadModelEnvelope) => {
+            doc.value.boosterMetadata!.optimisticConcurrencyValue = this.generateOptimisticConcurrencyValue(doc.value)
+          })
+          resolve(docs)
```
alvaroloes, line 44: Why do we need to do this when reading the read model? Would it not be already stored in the DB?

packages/framework-provider-local/src/services/read-model-registry.ts
```diff
@@ -32,27 +37,67 @@ export class ReadModelRegistry {
...
   }
 
-  public async store(readModel: ReadModelEnvelope): Promise<void> {
+  public async store(readModel: ReadModelEnvelope, expectedCurrentVersion: number): Promise<void> {
+    const readModelOptimistic: ReadModelEnvelope & { uniqueKey?: string } = readModel
+    readModelOptimistic.uniqueKey = `${readModel.typeName}_${readModel.value.id}_${readModel.value.boosterMetadata?.version}`
+    if (readModelOptimistic.value.boosterMetadata?.version === 1) {
+      return this.insert(readModel)
+    }
+    return this.update(readModelOptimistic, expectedCurrentVersion)
```
alvaroloes, line 57: So we have three values (I think) related to optimistic concurrency, but I don't fully get their purpose. We have:
- readModel.boosterMetadata.version (contains the read model version)
- readModel.uniqueKey (a unique key build from the version so that you can achieve the optimistic concurrency)
- readModel.boosterMetadata.optimisticConcurrencyValue (apparently, it also contains the read model version?)

Could you please clarify their purpose?

packages/framework-provider-local/src/services/read-model-registry.ts
```diff
@@ -32,27 +37,67 @@ export class ReadModelRegistry {
...
     return new Promise((resolve, reject) => {
       this.readModels.update(
-        //use nedb dot notation value.id to match the record (see https://github.com/louischatriot/nedb#finding-documents)
-        { typeName: readModel.typeName, 'value.id': readModel.value.id },
+        {
+          typeName: readModel.typeName,
+          'value.id': readModel.value.id,
+          $or: [
+            { 'value.boosterMetadata.optimisticConcurrencyValue': { $exists: false } },
+            { 'value.boosterMetadata.optimisticConcurrencyValue': expectedCurrentVersion },
```
alvaroloes, line 76: Oh! I thought NeDB didn't support optimistic concurrency... Why do we need this condition if we are already using the unique key approach?

packages/framework-provider-local/src/services/read-model-registry.ts
```diff
@@ -32,27 +37,67 @@ export class ReadModelRegistry {
...
+          ],
+        },
         readModel,
-        { upsert: true },
-        (err) => {
+        { upsert: false, returnUpdatedDocs: true },
+        (err: unknown, numAffected: number) => {
+          if (numAffected === 0) {
```
alvaroloes, line 82: I though when we have an error updating something, numAffected is always 0. How do you know that if this condition is true, then it must be a optimistic concurrency error?

alvaroloes (commented):
Just tons of questions

packages/framework-provider-local/src/services/read-model-registry.ts
gonzalojaubert: There could be old readmodels where the value is undefined

packages/framework-provider-local/src/services/read-model-registry.ts
gonzalojaubert: You are right, the optimisticConcurrencyValue is not needed as it will be the version always. I will remove it and simplify it. The uniqueKey is, as you said, used to avoid **inserting** same ReadModel twice. The value of this field will be `<id>_<boosterMetadata.version>`. If the ReadModel doesn't exists in the database and there are two instances trying to insert the same ReadModel, then both of them will try to do it with version = 0. To be sure that only one of the insert finish successfully we will need to look up into the database before inserting to check if another instance inserted the same read model. As we don't are not able to lock rows or tables we need a unique key to ensure we will never insert a read model with the same and same version (in this case 0).

packages/framework-provider-local/src/services/read-model-registry.ts
gonzalojaubert: I'll change it to use the version as the field is not needed. About your question, the uniqueKey is only to prevent inserting the same read model twice and the version to update only when the version number is equals to the expected one. If the version is not equals to the expected one it means that the row was updated by another instance.

packages/framework-provider-local/src/services/read-model-registry.ts
gonzalojaubert: Yes, it's not a perfect solution but in this case we are updating read models only son we can expect that the updated documents will be greater than 0. In this case, using the query (version = expected version) while updating we will have 0 documents updated if the row was updated by another instance.

alvaroloes (approved):
What a great work, specially with the integration tests!!

charlietfe (approved):

PR #1416, opened 2023-06-14
Remove events

gonzalojaubert:
## Description
Add Remove events action

## Changes
See https://github.com/boostercloud/booster/compare/main...Optum:booster:remove_events?expand=1#diff-f29d96b182c51086eeac1a87c2c597dac61b43e1a28fcc4fbe76a51d4c02fec2

## Checks
- [X] Project Builds


packages/framework-core/src/decorators/projects.ts
```diff
@@ -44,15 +60,46 @@ function registerProjection(
...
+  projectionMetadata: ProjectionMetadata<EntityInterface, ReadModelInterface>,
+  configuration: Record<string, Array<ProjectionMetadata<EntityInterface, ReadModelInterface>>>
+): void {
+  const entityProjections = configuration[originName] || []
+  if (entityProjections.indexOf(projectionMetadata) < 0) {
+    // Skip duplicate registrations
+    entityProjections.push(projectionMetadata)
+    configuration[originName] = entityProjections
+  }
+}
```
davidverdu, line 88: This function receives a param called ´config´ and another called `configuration`. It seems that the latter looks like the unprojection metadata. Could we rename this argument with a different name? (not sure what name should we use)

packages/framework-core/src/services/read-model-store.ts
```diff
@@ -132,6 +135,61 @@ export class ReadModelStore {
...
+        if (missingProjection) {
+          throw new Error(
+            `Missing UnProjection for ReadModel ${missingProjection.class.name} with joinKey ${missingProjection.joinKey} for entity ${entitySnapshotEnvelope.entityTypeName}`
+          )
+        }
+      }
+      return unProjections
+    }
+    return this.entityProjections(entitySnapshotEnvelope)
+  }
```
davidverdu, line 161: this function has 2 responsibilities and they depend on a boolean argument. I would split this function into two, one called `getUnprojections(entitySnapshotEnvelope: EntitySnapshotEnvelope)`  and the other one would be the one you have `this.entityProjections(entitySnapshotEnvelope: EntitySnapshotEnvelope)`. From the caller function I'll use this clause to call one or the other: `const projections = deleteEvent ? this.getUnProjections(entitySnapshotEnvelope) : this.entityProjections(entitySnapshotEnvelope)`

packages/framework-provider-azure/src/library/event-delete-adapter.ts
```diff
@@ -0,0 +1,135 @@
...
+  const eventStore = config.resourceNames.eventsStore
+  if (!events || events.length === 0) {
+    logger.warn('Could not find events to delete')
+    return
+  }
+  for (const event of events) {
+    const newEvent = buildNewEvent(event)
+    const partitionKey = partitionKeyBuilder(event.entityTypeName, event.entityID, 'event')
+    await replaceOrDeleteItem(cosmosDb, eventStore, config, event.id, partitionKey, newEvent)
+  }
```
davidverdu, line 65: I guess that the only way to delete entries from CosmoDB is one by one, isn't it?

packages/framework-integration-tests/integration/provider-unaware/end-to-end/delete.events.integration.ts
```diff
@@ -0,0 +1,485 @@
...
+      limit: 99999,
+    },
+    mutation: gql`
+      mutation EntitiesIdsFinder($entityName: String!, $limit: Float!) {
+        EntitiesIdsFinder(input: { entityName: $entityName, limit: $limit })
+      }
+    `,
+  })
+  return result?.data?.EntitiesIdsFinder
+}
```
davidverdu, line 485: I guess that many of these methods are duplicated elsewhere or can be reused in other tests, and we could create a file with these reusable helpers for all the tests. Nonetheless, I guess this refactor should go in another PR

packages/framework-integration-tests/integration/provider-unaware/end-to-end/delete.events.integration.ts
```diff
@@ -0,0 +1,485 @@
...
+
+    it('then the cart event is not changed', async () => {
+      expect(firstCartEvent.deletedAt).to.be.null
+      expect(firstCartEvent.value.cartId).to.be.eq(firstCartId)
+    })
+
+    it('and ReadModels for the cart id are not deleted', async () => {
+      firstCartReadModel = await getCart(loggedClient, firstCartId)
+      expect(firstCartReadModel.id).to.be.eq(firstCartId)
+    })
```
davidverdu, line 182: are these `it` tests intended to be run sequentially? As far as I understand, `it`tests inside a `describe` should be independent and able to run in whatever order. I'm not sure if we are executing this here with another mindset

davidverdu (commented):
Generally speaking looks good for such a complex feature, congrats. Nonetheless, I left some comments. Let me know when you address them, so that I can do the final review

packages/framework-core/src/decorators/projects.ts
MarcAstr0: Upon closer inspection, the `configuration` parameter is redundant since the projection configuration is already part of the `config` object. It was removed in commit 7b4b1e4.

packages/framework-core/src/services/read-model-store.ts
MarcAstr0: Good suggestion. Applied in commit 200c434.

packages/framework-integration-tests/integration/provider-unaware/end-to-end/delete.events.integration.ts
MarcAstr0: Mocha tests run sequentially. In fact, there's [a separate library](https://www.npmjs.com/package/mocha-parallel-tests) that needs to be used to run Mocha tests in parallel.

packages/framework-provider-azure/src/library/event-delete-adapter.ts
MarcAstr0: Correct. The Cosmos DB client for TypeScript does not provide a method for deleting items in bulk.

packages/framework-integration-tests/integration/provider-unaware/end-to-end/delete.events.integration.ts
MarcAstr0: Absolutely. Some other tests call _similar_ helper methods, but as you say, the work required to create that method probably falls outside this PR's scope.

packages/framework-integration-tests/integration/provider-unaware/end-to-end/delete.events.integration.ts
davidverdu: ok

davidverdu (approved):
LGTM!

PR #1419, opened 2023-06-26
fix(framework-provider-aws): ensure storeSnapshot is idempotent

DKarandikar:
## Description
Fix DynamoDB call in `storeSnapshot` by ensuring it is idempotent
This fixes errors when eventHandlers are accessing entities via `Booster.entity` and also attempting to cache snapshots

## Changes

- Removed ConditionExpression from dyanmoDB call to ensure idempotency

## Checks
- [ ] Project Builds


javiertoledo:
Great catch @DKarandikar. I completely missed removing that conditional the last time we changed this part of the code 😅

NickSeagull:
Awesome! Looking good

PR #1421, opened 2023-07-04
Health sensor

gonzalojaubert:
Add Health HTTP endpoints

See https://github.com/boostercloud/booster/blob/68bbec8da1def7bfa4775c41e277ed19640cfe3c/website/docs/10_going-deeper/health/sensor-health.md


packages/framework-core/src/sensor/health/health-indicators/booster-health-indicator.ts
```diff
@@ -0,0 +1,34 @@
...
+      return result
+    } catch (e) {
+      return { status: HealthStatus.DOWN, details: e }
+    }
+  }
+
+  private async isUp(config: BoosterConfig): Promise<HealthStatus> {
+    const graphqlUp = await config.provider.sensor.isGraphQLFunctionUp(config)
+    const eventsCatalog = await config.provider.sensor.isEventCatalogUp(config)
+    return graphqlUp && eventsCatalog ? HealthStatus.UP : HealthStatus.DOWN
```
alvaroloes, line 32: Shouldn't we also include here the read models health status? I understood this health indicator as a general one that confirms everything is OK

packages/framework-core/src/sensor/health/health-indicators/booster-version.ts
```diff
@@ -0,0 +1,22 @@
...
+  const logger = getLogger(config, 'boosterVersion')
+  try {
+    const packageJsonContents = require(path.join(projectAbsolutePath, 'package.json'))
+    const version = packageJsonContents.dependencies['@boostercloud/framework-core']
+    if (!version) {
+      logger.warn('Could not get Booster Version')
+      return ''
+    }
+    const versionParts = version.replace('workspace:', '').replace('^', '').replace('.tgz', '').split('-')
+    return versionParts[versionParts.length - 1]
```
alvaroloes, line 17: Shouldn't we calculate the Booster version based on the `package-lock.json`? We could have any semantic version in `package.json` file, but to know the exact version that is used, we need to look at the lock file

packages/framework-core/src/sensor/health/health-indicators/default-booster-health-indicators.ts
```diff
@@ -0,0 +1,69 @@
...
+    BOOSTER_HEALTH_INDICATORS_IDS.DATABASE_READ_MODELS,
+    'Booster Database ReadModels Catalogs',
+    BoosterDatabaseReadmodelsCatalogsHealthIndicator
+  )
+  return {
+    [root.healthIndicatorConfiguration.id]: root,
+    [boosterFunction.healthIndicatorConfiguration.id]: boosterFunction,
+    [boosterDatabase.healthIndicatorConfiguration.id]: boosterDatabase,
+    [databaseEventsCatalog.healthIndicatorConfiguration.id]: databaseEventsCatalog,
+    [databaseReadModelsCatalogs.healthIndicatorConfiguration.id]: databaseReadModelsCatalogs,
```
alvaroloes, line 67: If `boosterDatabase` is a mix of the eventsCatalog and the readModelsCatalogs, Why also including these separately?

packages/framework-core/src/sensor/health/health-indicators/booster-database-readmodels-catalogs-health-indicator.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  BoosterConfig,
+  HealthIndicatorResult,
+  HealthIndicatorMetadata,
+  HealthStatus,
+} from '@boostercloud/framework-types'
+
+export class BoosterDatabaseReadmodelsCatalogsHealthIndicator {
```
alvaroloes, line 8: What does `Catalogs` in the name refers to?

packages/framework-integration-tests/src/config/config.ts
```diff
@@ -25,6 +25,12 @@ function configureInvocationsHandler(config: BoosterConfig) {
   }
 }
 
+function configureBoosterSensorHealth(config: BoosterConfig) {
+  Object.values(config.sensorConfiguration.health.booster).forEach((indicator) => {
+    indicator.enabled = true
```
alvaroloes, line 30: Why is this needed? Does`config.sensorConfiguration.health.booster` refer to Booster default health checks? If so, Is the reason for this loop that those default health checks are disabled by default?

If so, Does a normal application need to do this to enable Booster internal health checks?

(so many questions! 😂 )

packages/framework-integration-tests/src/sensor/health/application-add-booster-database-children-health-indicator.ts
```diff
@@ -0,0 +1,26 @@
...
+  BoosterConfig,
+  HealthIndicatorResult,
+  HealthIndicatorMetadata,
+  HealthStatus,
+  BOOSTER_HEALTH_INDICATORS_IDS,
+} from '@boostercloud/framework-types'
+
+@HealthEndpoint({
+  id: `${BOOSTER_HEALTH_INDICATORS_IDS.DATABASE}/myApplication`,
+  name: 'Indicator added to the Booster Database indicator through My Application',
```
alvaroloes, line 12: This confused me.... So you can add sensors... to specific parent sensors? Curious!

packages/framework-integration-tests/src/sensor/health/application-child-health-indicator.ts
```diff
@@ -0,0 +1,25 @@
...
+import {
+  BoosterConfig,
+  HealthIndicatorResult,
+  HealthIndicatorMetadata,
+  HealthStatus,
+} from '@boostercloud/framework-types'
+
+@HealthEndpoint({
+  id: 'myApplication/child',
+  name: 'My Application child',
```
alvaroloes, line 11: What would happen if we add this Health sensor but we don't add the parent one (`myApplication`)?

packages/framework-integration-tests/src/sensor/health/application-health-indicator.ts
```diff
@@ -0,0 +1,25 @@
...
+})
+export class ApplicationHealthIndicator {
+  public async health(
+    config: BoosterConfig,
+    healthIndicatorMetadata: HealthIndicatorMetadata
+  ): Promise<HealthIndicatorResult> {
+    return {
+      status: HealthStatus.UP,
+    } as HealthIndicatorResult
+  }
```
alvaroloes, line 24: Question about parent -> children relationship: 
~When your child doesn't stop crying and you have tried everything~ errr, Is the parent sensor required to also return information about their children? Or is that already managed automatically?

packages/framework-provider-aws-infrastructure/src/test-helper/aws-test-helper.ts
```diff
@@ -6,6 +6,7 @@ import { AWSQueries } from './aws-queries'
 interface ApplicationOutputs {
```
alvaroloes, line 6: ----------------------------- END FIRST REVIEW -----------------------

alvaroloes (commented):
End of first partial review

packages/framework-core/src/sensor/health/health-endpoint.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @param {Object} attributes
+ * @param {string} attributes.id - Unique indicator identifier
+ * @param {string} attributes.name - Indicator description
+ * @param {boolean} attributes.enabled - If false, this indicator and the components of this indicator will be skipped
+ * @param {boolean} attributes.details - If false, the indicator will not include the details
+ * @param {boolean} [attributes.showChildren] - If false, this indicator will not include children components in the tree.
+ *      Children components will be shown through children urls
+ * @constructor
+ */
+export function HealthEndpoint(
```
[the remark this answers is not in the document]
gonzalojaubert: Yeah, makes sense

packages/framework-core/src/sensor/health/booster-health-service.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export class BoosterHealthService {
+  constructor(readonly config: BoosterConfig) {}
+
+  public async boosterHealth(request: any): Promise<unknown> {
+    try {
+      const healthEnvelope: HealthEnvelope = this.config.provider.sensor.rawRequestToHealthEnvelope(request)
+      await this.validate(healthEnvelope)
+      const healthProviders = this.getHealthProviders()
+      const head = this.headHealthProviders(healthEnvelope, healthProviders)
```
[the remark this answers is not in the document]
gonzalojaubert: The name doesn't makes sense. The steps are:
1. Line 23: Get all configured health providers  
2. LIne 24: Get the first level of health providers. Those is the parent or parents providers. I'll rename it to `parents`
3. Line 25: Starting from parents, resolve those and all children providers

packages/framework-core/src/sensor/health/health-indicators/default-booster-health-indicators.ts
gonzalojaubert: BoosterDatabase returns details related to the database only (the url right now), and use the events and readmodels catalogs status to calculate the database status, but it will not show information about the catalogs, this is done in each catalog indicator. We need to add them manually to be shown

packages/framework-core/src/sensor/health/health-indicators/booster-database-readmodels-catalogs-health-indicator.ts
gonzalojaubert: It refers to the metadata in which definitions of tables are defined. The idea is that indicator will provide this information, not the table contents. Any suggestion to improve it will be welcomed

packages/framework-integration-tests/src/config/config.ts
gonzalojaubert: hahaha, np, thanks for your questions and suggestions, all of them makes sense. Let me try to answer to all of them:

1. Why is this needed: By default all the health checks are disabled by default. We could add a method to enable all the defaults at once. Thoughts?
2. Doesconfig.sensorConfiguration.health.booster refer to Booster default health checks?: Yes
3. Is the reason for this loop that those default health checks are disabled by default?: Yes
4. Does a normal application need to do this to enable Booster internal health checks?: Yes

packages/framework-integration-tests/src/sensor/health/application-add-booster-database-children-health-indicator.ts
gonzalojaubert: Yes, for instance, if you have a Rocket that a new database container and want to add this information to the database health indicator

packages/framework-integration-tests/src/sensor/health/application-child-health-indicator.ts
gonzalojaubert: Suppose you have the following sensors:
  A
  A/B/C

If you navigate to `/A` then only **A** sensor will be returned. But you still will be able to navigate to `/A/B/C` and get this specific sensor. If you navigate to `/A/B` it will not be found

packages/framework-integration-tests/src/sensor/health/application-health-indicator.ts
gonzalojaubert: it's managed automatically. See the code you suggested the refactor about children process: https://github.com/boostercloud/booster/pull/1421/files#diff-f5101051c59853404201e54e7e21c9aee4585ace4f8077257a0b166d772f5a40R45

packages/framework-core/src/sensor/health/health-indicators/booster-version.ts
alvaroloes: Ok

packages/framework-core/src/sensor/health/health-indicators/default-booster-health-indicators.ts
alvaroloes: Aha! Thanks

packages/framework-core/src/sensor/health/health-indicators/booster-database-readmodels-catalogs-health-indicator.ts
alvaroloes: @gonzalogarciajaubert Mmm mayby I still don't get it, but to me it can be removed and leave the class name as `BoosterDatabaseReadmodelsHealthIndicator`

What do you think

packages/framework-integration-tests/src/config/config.ts
alvaroloes: @gonzalogarciajaubert Thanks for answering. Regarding point 1. First a question: Are the booster health sensor referred by name so that we can enable just some of them? I mean, can we do something like:
```typescript
const sensors = config.sensorConfiguration.health.booster
sensors.readmodels.enabled = true
sensors.eventStore.enabled = true
// etc
```
If so, I think we don't need that method, because it is always good to have that explicitly in the configuration file. If not, then it could be hard for the user to enable some of them and not others, as they would need to do an... array search? In that case, I would provide some methods to facilitate this or store the sensors with a name, as written above

packages/framework-integration-tests/src/sensor/health/application-child-health-indicator.ts
alvaroloes: Aha! Ok, got it. So the "parent -> child" relationship is a loose relationship. Just like a namespace relationship. Great!

packages/framework-provider-azure-infrastructure/src/test-helper/azure-test-helper.ts
```diff
@@ -68,6 +70,11 @@ export class AzureTestHelper {
     return url
   }
 
+  public static async healthURL(resourceGroup: ResourceGroup): Promise<string> {
+    const environment = process.env.BOOSTER_ENV ?? 'azure'
```
alvaroloes, line 74: Does `azure` as default value for the environment have a special meaning?

packages/framework-provider-azure/src/library/health-adapter.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export async function getConstant(
+  cosmosDb: CosmosClient,
+  config: BoosterConfig,
+  catalog: string
+): Promise<string | undefined> {
+  const container = getContainer(cosmosDb, config, catalog)
+  const { resources } = await container.items.query('SELECT 1 FROM c', { maxItemCount: -1 }).fetchAll()
+  return resources && resources.length > 0 ? resources[0] : undefined
```
alvaroloes, line 23: What kind of magic this is? 😳 
You are doing a `SELECT 1` query with an item count of... -1? And then we return the first result which is... 1 I guess?
Why do get need this `1`? Also, from the name `getConstant` it seems it is getting a constant. 

I guess this is just doing a dumb query to verify this... `catalog` (still don't fully get what a catalog is) works, correct? 
Maybe we can call this function `isCatalogUp` and return a boolean?

packages/framework-provider-azure/src/library/health-adapter.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export async function eventsCatalogHealthDetails(cosmosDb: CosmosClient, config: BoosterConfig): Promise<unknown> {
+  const container = getContainer(cosmosDb, config, config.resourceNames.eventsStore)
+  const url = container.url
+  const partitionKeyDefinition = await partitionKeys(container)
+  const count = await countAll(container)
+  return {
+    url: url,
+    partitionKeyDefinition: partitionKeyDefinition,
+    count: count,
```
alvaroloes, line 44: Wow, we return the count of events as part of the health check. This is one of the most detailed health check I've seen!

packages/framework-provider-azure/src/library/health-adapter.ts
```diff
@@ -0,0 +1,121 @@
...
+  try {
+    const basePath = process.env[environmentVarNames.restAPIURL]
+    return `${basePath}/graphql`
+  } catch (e) {
+    return ''
+  }
+}
+
+export async function isEventCatalogUp(cosmosDb: CosmosClient, config: BoosterConfig): Promise<boolean> {
+  await getConstant(cosmosDb, config, config.resourceNames.eventsStore)
```
alvaroloes, line 58: Shouldn't we check if the result is "1" or undefined? If it is undefined, I guess the database is not up, correct? 

Are we expecting `getConstant` to throw if it is not up? Asking this because we are always returning true here without controlling the exception, so not sure why this call is needed.

packages/framework-provider-azure/src/library/health-adapter.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export async function areReadModelsCatalogsUp(cosmosDb: CosmosClient, config: BoosterConfig): Promise<boolean> {
+  const promises = Object.values(config.readModels).map((readModel) => {
+    const name = readModel.class.name
+    const container = config.resourceNames.forReadModel(name)
+    return getConstant(cosmosDb, config, container)
+  })
+  await Promise.all(promises)
+  return true
```
alvaroloes, line 69: Same question here. We are not checking neither the value returned by `getConstant` nor whether the promise threw an exception. Why are these calls needed then?

website/docs/10_going-deeper/health/sensor-health.md
```diff
@@ -0,0 +1,402 @@
...
+
+```text
+└── database
+```
+
+but you can access to them using the component url https://your-application-url/sensor/health/database/events
+
+```text
+└── events
+```
```
alvaroloes, line 402: Wow! Thanks a lot for such detailed docs

packages/framework-core/src/sensor/health/health-indicators/booster-database-readmodels-catalogs-health-indicator.ts
gonzalogarciajaubert: Yeah, it could be confusing some times but let me change it and we could review it.

packages/framework-integration-tests/src/config/config.ts
gonzalogarciajaubert: You can do:

```typescript
  const sensors = config.sensorConfiguration.health.booster
  sensors[BOOSTER_HEALTH_INDICATORS_IDS.DATABASE].enabled = true
```

packages/framework-provider-azure-infrastructure/src/test-helper/azure-test-helper.ts
gonzalogarciajaubert: AFAIK it doesn't. It's a copy/paste from `graphqlURL` method. Do you think is safe to remove it?

packages/framework-provider-azure/src/library/health-adapter.ts
gonzalogarciajaubert: :) 

1.`maxItemCount` is not the number of items to be returned (this will be TOP), it only applies to the number of items retrieved in each round trip to the server. Setting it to -1 the service determines the optimal page count dynamically instead of using the default value (IIRC it's 100).
2. `SELECT 1`. We are only checking that the container exists and is able to solve queries. This query is the faster one in this case.
3. Let me rename the method and return a boolean!

packages/framework-provider-azure/src/library/health-adapter.ts
gonzalogarciajaubert: Yes yes, I missed this one here and in the read models, let me fix it

packages/framework-provider-azure/src/library/health-adapter.ts
gonzalogarciajaubert: Yes, it is fixed with the previous one, thanks!

packages/framework-integration-tests/src/config/config.ts
alvaroloes: @gonzalojaubert That's enough. Could you please add an example like this in the docs?

packages/framework-provider-azure-infrastructure/src/test-helper/azure-test-helper.ts
alvaroloes: @gonzalogarciajaubert mmm, the `BOOSTER_ENV` var must always be set, but I think it is safer to check it. I would use a common word that doesn't refer to the provider. What do you think about `default` or `booster`?

packages/framework-provider-azure/src/library/health-adapter.ts
alvaroloes: Thanks for the explanation!

alvaroloes (approved):

charlietfe (approved):

PR #1423, opened 2023-07-24
Update dependencies

gonzalojaubert:
Update Booster dependencies:

* jsonwebtoken
* underscore
* mocha
* jwks-rsa


common/changes/@boostercloud/framework-core/update_dependencies_2023-07-24-09-16.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@boostercloud/framework-core",
+      "comment": "Update booster dependencies",
+      "type": "minor"
```
javiertoledo, line 6: Do we need to perform a release that is just upgrading dependencies? Unless these dependencies were upgraded for security reasons, I guess we can set this to "none" and get these upgraded dependencies on the next release along with some features 🤔

javiertoledo (approved):
LGTM, let's merge this ASAP as it doesn't change any behavior and all tests are passing. The only detail is that I'm not sure it's worth creating a minor release with this, but I'll let you decide @gonzalogarciajaubert.

charlietfe (approved):

alvaroloes (approved):

common/changes/@boostercloud/framework-core/update_dependencies_2023-07-24-09-16.json
gonzalogarciajaubert: It's a security update

PR #1430, opened 2023-09-19
Add Azure Event Hub

gonzalojaubert:
Add Azure Event Hub

Update your index.ts to add new methods:

```
boosterProduceEventStream,
  boosterConsumeEventStream,
```


packages/framework-integration-tests/src/config/config.ts
```diff
@@ -25,6 +25,17 @@ function configureInvocationsHandler(config: BoosterConfig) {
   }
 }
 
+function configureEventHub(config: BoosterConfig) {
+  config.eventStreamConfiguration = {
+    enabled: true,
+    parameters: {
+      streamTopic: 'test',
+      partitionCount: 3,
+      messageRetention: 1,
```
alvaroloes, line 34: Questions (take into account that my knowledge about Event Hub is very limited):

- Is there a way to manage the partitionCount depending on load? I mean, automatically?
- What would happen if the messageRetention is met? Are events never delivered anymore? Is there a fallback mechanism we could apply? 
    - For example (probably a bit crazy), if the messageRetention is met for some event and we can execute a function before deleting the event, we could fallback to the CosmosDB default "event dispacher" to update read models and call event handlers (good thing is that the event is already stored in the CosmosDB event-stream)

website/docs/10_going-deeper/azure-scale.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+## Infrastructure
+
+Enabling `EventStreamConfiguration` will apply some changes to the infrastructure:
+
+* Two functions will be created
+  * One function with a CosmosDB consumer that will produce Event Hubs events. Also, it will include the readModels functions, schedule functions app, etc...
+  * One function with an Event Hub consumer function app. This function will allow you to define the number of instances to be created
+* A new container to handle duplicated consumed events
+* A new Event Hub will be added for event handling.
```
alvaroloes, line 39: Aren't we missing "A Function App per partition"?

packages/framework-provider-azure-infrastructure/src/infrastructure/index.ts
```diff
@@ -36,7 +42,13 @@ async function deployApp(config: BoosterConfig, rockets?: InfrastructureRocket[]
...
   }
 
-  await applicationBuilder.uploadFile(applicationBuild.zipResource)
+  const resourceGroupName = createResourceGroupName(config.appName, config.environmentName)
+  const functionAppName = createFunctionResourceGroupName(resourceGroupName)
+  await FunctionZip.deployZip(config, functionAppName, resourceGroupName, applicationBuild.zipResource)
+  if (config.eventStreamConfiguration.enabled) {
+    const streamFunctionAppName = createStreamFunctionResourceGroupName(resourceGroupName)
+    await FunctionZip.deployZip(config, streamFunctionAppName, resourceGroupName, applicationBuild.consumerZipResource)
+  }
```
alvaroloes, line 51: If I understood correctly, now we have two different zips, although the code is the same... Is this true?
If so, Can't we use the same zip resource as we do for all the other functions?

packages/framework-provider-azure-infrastructure/src/infrastructure/synth/terraform-containers.ts
```diff
@@ -156,4 +152,31 @@ export class TerraformContainers {
...
+      partitionKeyPath: `/${dedupAttributes.partitionKey}`,
+      uniqueKey: [{ paths: [`/${dedupAttributes.partitionKey}`] }],
+      partitionKeyVersion: 2,
+      defaultTtl: -1,
+      autoscaleSettings: {
+        maxThroughput: MAX_CONTAINER_THROUGHPUT,
+      },
+      provider: azureProvider,
+    })
+  }
```
alvaroloes, line 181: ---------------- End partial review 1 -------------------

packages/framework-core/src/booster-event-stream-producer.ts
```diff
@@ -0,0 +1,18 @@
...
+import { RawEventsParser } from './services/raw-events-parser'
+
+export class BoosterEventStreamProducer {
+  @Trace(TraceActionTypes.PRODUCE_STREAM_EVENTS)
+  public static async produce(request: unknown, config: BoosterConfig): Promise<void> {
+    const logger = getLogger(config, 'BoosterEventStreamProducer#produce')
+    logger.debug('Produce event workflow started for request:', require('util').inspect(request, false, null, false))
+    try {
+      const eventEnvelopes = config.provider.events.rawToEnvelopes(request)
+      await RawEventsParser.streamPerEntityEvents(config, eventEnvelopes, config.provider.events.produce)
```
alvaroloes, line 13: I still have a question here: Why do we need to produce the events grouped by entity name and ID (sharding key) if EventHub will already manage that grouping by itself?

My guess is that EventHub will automatically split events among partitions by sharding key

packages/framework-provider-azure/src/library/events-stream-consumer-adapter.ts
```diff
@@ -0,0 +1,65 @@
...
+  const events = (context.bindings.eventHubMessages as EventStream) || []
+  logger.debug(`Dedup ${events.length} events`)
+
+  const resources: EventStream = []
+  for (const event of events) {
+    const rawParsed = JSON.parse(event as string)
+    const eventTag: DedupEventStream = {
+      primaryKey: rawParsed._etag,
+      createdAt: new Date().toISOString(),
+      ttl: config.eventStreamConfiguration.parameters?.dedupTtl ?? 86400,
```
alvaroloes, line 28: Two things:
- Great point on being able to define the TTL
- A bit nitpicking: I guess this default TTL is in seconds (24 hours). Could you please put this in a constant with a name?

packages/framework-provider-azure-infrastructure/src/infrastructure/synth/terraform-event-hub.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (!eventHubNamespace) {
+      throw new Error('Undefined eventHubNamespace resource')
+    }
+    const idApiManagement = toTerraformName(appPrefix, 'eh')
+    return new eventhub.Eventhub(terraformStack, idApiManagement, {
+      name: eventHubName,
+      resourceGroupName: resourceGroupName,
+      provider: azureProvider,
+      namespaceName: eventHubNamespace.name,
+      partitionCount: config.eventStreamConfiguration.parameters?.partitionCount ?? this.DEFAULT_PARTITION_COUNT, // Changing this will force-recreate the resource. Cannot be changed unless Eventhub Namespace SKU is Premium
```
alvaroloes, line 31: Do you know what would happen if the resource is recreated? I mean, imagine we have a partition count of 1 and now we want to us 3. The resource will be created but... will some events be lost? I know the events are already in the event-store, but we still need to deliver them to the event processor. Do you know if the "recreation" will happen flawlessly?

packages/framework-provider-azure/src/index.ts
```diff
@@ -41,6 +44,25 @@ if (typeof process.env[environmentVarNames.cosmosDbConnectionString] === 'undefi
...
+  typeof eventHubName === 'undefined' ||
+  eventHubConnectionString === '' ||
+  eventHubName === ''
+) {
+  producer = {} as any
+} else {
+  producer = new EventHubProducerClient(eventHubConnectionString, eventHubName, {
+    retryOptions: {
+      maxRetries: 5,
+      mode: RetryMode.Exponential,
```
alvaroloes, line 61: Not sure if it will be possible but, Could we not include this in the configuration?

packages/framework-provider-azure/src/library/events-stream-producer-adapter.ts
```diff
@@ -0,0 +1,56 @@
...
+    const isAdded = batch.tryAdd({ body: eventEnvelope })
+
+    if (isAdded) {
+      logger.info(`Added ${JSON.stringify(eventEnvelope)} to the batch`)
+      ++i
+      continue
+    }
+
+    if (batch.count === 0) {
+      throw new Error(`Message was too large and can't be sent until it's made smaller. ${eventEnvelope}`)
```
alvaroloes, line 34: This is a "just in case": I think that including the `eventEnvelope` in the string like this will make it appear as `[object Object]` so, if we face this error, it will be almost impossible to know which event is too large. 
Could you please check if this is the case and, if so, do a, for example, JSON.stringify with the event envelope?

website/docs/10_going-deeper/azure-scale.mdx
```diff
@@ -0,0 +1,42 @@
...
+      streamTopic: 'test',
+      partitionCount: 3,
+      messageRetention: 1,
+    },
+  }
+```
+
+## Parameters
+* StreamTopic: Define the internal topic name Booster will use.
+* PartitionCount: Number of Event Hub partitions. The number of functions app consumer instances will match the partition count
```
alvaroloes, line 28: I saw a comment in the code saying that changing this will make the resource to be recreated. I would be ideal to include this as a warning in these docs, along with the effects this "recreation" could cause (are events lost?) so that uses can decide accordingly

alvaroloes (commented):
This PR is amazing! Everything is very well thought 🏅

packages/framework-integration-tests/src/config/config.ts
gonzalogarciajaubert: 1. No, the partition count should be defined
2. If the message retention is met events are removed from the event hub. If we need to archive events beyond the allowed retention period, we can have them automatically stored in Azure Storage or Azure Data Lake. AFAIK there is not mechanisms to execute before deleting an event

website/docs/10_going-deeper/azure-scale.mdx
gonzalogarciajaubert: There will be a Function App to consume the events with X instances. If the partition number is 3, there should be 3 instances of the `Event Hub consumer function app`

packages/framework-provider-azure-infrastructure/src/infrastructure/index.ts
gonzalogarciajaubert: It's not exactly the same as the definitions of the functions in each file are different. The `zipResource` and the `consumerZipResource` are differents:

```
    const zipResource = await FunctionZip.copyZip(azureStack.applicationStack.functionDefinitions!, 'functionApp.zip')
    const consumerZipResource = await FunctionZip.copyZip(
      azureStack.applicationStack.consumerFunctionDefinitions!,
      'consumerFunctionApp.zip'
    )
```

packages/framework-integration-tests/src/config/config.ts
gonzalogarciajaubert: About the retention: Azure allows you to define a cleanup policy. This policy could be `compact` for an infinite retention time. This option is not available in Terraform

packages/framework-provider-azure-infrastructure/src/infrastructure/synth/terraform-event-hub.ts
gonzalogarciajaubert: It looks like this issue was solved [here](https://github.com/hashicorp/terraform-provider-azurerm/pull/15088). It should be possible to increase the partition number. I will update the comment.

One thing about increasing the partition number is (from the Azure documentation):

```
If you use partition key with your producer applications and depend on key hashing to ensure ordering in a partition, dynamically adding partitions isn't recommended.

 Important

While the existing data preserves ordering, partition hashing will be broken for messages hashed after the partition count changes due to addition of partitions.
```

packages/framework-provider-azure/src/index.ts
gonzalogarciajaubert: You don't want this in the configuration?

packages/framework-provider-azure-infrastructure/src/infrastructure/synth/terraform-event-hub.ts
alvaroloes: @gonzalogarciajaubert Thanks! It would be grate to have this in the Booster documentation, pointing to that Azure link

packages/framework-provider-azure/src/index.ts
alvaroloes: @gonzalogarciajaubert HAHAHA Now that I read the question, I though exactly the same as you, which is exactly the opposite of what I wanted to ask XD. 

What I meant is to _include_ this in the configuration. Many projects would probably want to configure this, and it is generic enough (not strictly associated to Azure) to include it in the config

alvaroloes (approved):

charlietfe (approved):

Issue #1429, opened 2023-09-18
TypeError: Cannot set property constructor of [object Object] which has only a getter


javiertoledo:
`ttsc -b tsconfig.json` Fails with TypeError During Compilation on Different Environments (Windows, WSL2, AWS EC2)

## Current Behavior

Two users reported that running the `boost deploy -e production` command or the `boost build` command results in a TypeError during the TypeScript compilation step. The error manifests as "Cannot set property constructor of [object Object] which has only a getter". This happens in multiple environments (Windows, WSL2, and AWS EC2) and seems to be independent of the environment and versions used. The error log points to a line in the `ttypescript` package.

The error message is as follows:

```plaintext
[2023-09-09T21:06:46.813Z] [91m[ChildProcessError] Project contains compilation errors:[39m
[2023-09-09T21:06:46.813Z] Command failed: npm run build
...
TypeError: Cannot set property constructor of [object Object] which has only a getter
...
```

## Expected behavior

The `boost deploy -e production` command and `boost build` command should successfully compile the TypeScript code and deploy/build the project without errors.

## Possible Solution

The issue might be related to a mismatch between versions or some outdated dependencies, although this needs to be confirmed. Another possible issue could be some incompatibility within the TypeScript configuration or some third-party library.

## Additional information

This report aggregates the experiences of multiple users from the Discord channel, verifying that this is not an isolated issue but one that affects multiple setups.

@javiertoledo haven't been able to reproduce the issue in macOS Ventura nor on a freshly created Github Codespace. All unit and integration tests have passed successfully.

### Environment

- Booster version: 1.18.1 (can be checked with `boost version`)
- Node/npm version: Node 16.20.2, ttypescript 1.5.13, typescript 4.7.4 (affected versions)
- OS: Windows, WSL2, AWS EC2

javiertoledo:
This will be solved once we switch to ts-patch: #1432

PR #1432, opened 2023-09-24
Dropped `ttypescript` in favor of `ts-patch`

javiertoledo:
## Description

This PR drops the use of `ttypescript`, which has been deprecated, in favor of the more modern library `ts-patch`.

There have been reports of `ttypescript` yielding compilation errors in some environments, so this change should also solve them https://github.com/boostercloud/booster/issues/1429

## Checks
- [x] Project Builds


gonzalogarciajaubert (approved):
LGTM

PR #1436, opened 2023-09-26
Store Azure events in batch

gonzalojaubert:


packages/framework-provider-azure/src/library/events-store-adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+ * Split events in chunks of size DEFAULT_CHUNK_SIZE
+ *
+ * @param arr
+ * @param size
+ */
+function chunkEvents(arr: Array<NonPersistedEventEnvelope>, size: number): Array<Array<NonPersistedEventEnvelope>> {
+  return Array.from({ length: Math.ceil(arr.length / size) }, (_: NonPersistedEventEnvelope, i: number) =>
+    arr.slice(i * size, i * size + size)
+  )
+}
```
alvaroloes, line 140: You can replace this function by using the lodash equivalent of `_.chunk(array, size)`

packages/framework-provider-azure/src/library/events-store-adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+  config: BoosterConfig
+): Promise<Array<EventEnvelope>> {
+  const logger = getLogger(config, 'store-events-adapter#storeEvents')
+  logger.debug('Storing EventEnvelopes with eventEnvelopes:', eventEnvelopes)
+  const envelopesWithCreatedAt: Array<EventEnvelope> = []
+
+  const eventsPerPartitionKey = eventEnvelopes.reduce(groupByPartitionKey, {})
+  for (const eventsPerPartitionKeyEntry of Object.entries(eventsPerPartitionKey)) {
+    const partitionKey = eventsPerPartitionKeyEntry[0]
+    const eventsInPartitionKey = eventsPerPartitionKeyEntry[1]
```
alvaroloes, line 35: You can simplify this with:
```suggestion
  for (const [partitionKey, eventsInPartitionKey] of Object.entries(eventsPerPartitionKey)) {
```

packages/framework-provider-azure/src/library/events-store-adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+    const chunksOfEventsInPartitionKey = chunkEvents(eventsInPartitionKey, DEFAULT_CHUNK_SIZE)
```
alvaroloes, line 36: Here we can use (as I commented in another comment):
```suggestion
    const chunksOfEventsInPartitionKey = _.chunk(eventsInPartitionKey, DEFAULT_CHUNK_SIZE)
```

packages/framework-provider-azure/src/library/events-store-adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+    for (const eventListInChunk of chunksOfEventsInPartitionKey) {
+      const eventEnvelopesChunkWithCreatedAt = toEventEnvelopes(eventListInChunk)
+      envelopesWithCreatedAt.push(...eventEnvelopesChunkWithCreatedAt)
+      const inputOperations: Array<CreateOperationInput> = toInputOperations(eventEnvelopesChunkWithCreatedAt, config)
+      const batchResponse = await batchEvents(cosmosDb, config, inputOperations, partitionKey)
+
+      // Batch is transactional and will roll back all operations if one fails
+      if (batchResponse?.code !== 200) {
+        logger.error(batchResponse)
```
alvaroloes, line 45: Here it would be better if we prefix this log with something like `"An error ocurred storing a batch of events. Batch response: " and then the batch response.

packages/framework-provider-azure/src/library/events-store-adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+        throw new Error(`Error ${batchResponse.substatus} storing events`)
```
alvaroloes, line 46: Besides the above log, we should also include information about the specific error in this throw. Here we are just saying "Hey, there was an error", and it could be a nightmare to debug if there are tons of unrelated logs and we need to find the above log. Better to have everything together

packages/framework-provider-azure/src/library/events-store-adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+function storableResource(config: BoosterConfig, eventEnvelope: EventEnvelope): JSONObject {
+  const logger = getLogger(config, 'store-events-adapter#storableResource')
+  try {
+    const partitionKey = partitionKeyForEvent(eventEnvelope.entityTypeName, eventEnvelope.entityID)
+    const eventEnvelopeWithEntityID = {
+      ...eventEnvelope,
+      [eventsStoreAttributes.partitionKey]: partitionKey,
+      [eventsStoreAttributes.sortKey]: eventEnvelope.createdAt,
+    }
+    return JSON.parse(JSON.stringify(eventEnvelopeWithEntityID))
```
alvaroloes, line 123: I guess this is done to verify the circular dependencies and the "BigInt" kind of value stated in the above comment, but it seems a bit hacky (and time consuming... as this will be used to store batches of hundred of events).

My question is: is this really needed? a Javascript object like the `eventEnvelope` is already a JSONObject. 
Also, if cosmos will fail when cyclic dependencies or BibInts are present, then we can let it fail.

In brief: could we just return here `eventEnvelopeWithEntityID`?

packages/framework-provider-azure/src/library/events-store-adapter.ts
MarcAstr0: We don't have lodash (and its type declarations) as a dependency. For just a single function that does exactly the same, I wouldn't bother adding another dependency, do you agree?

packages/framework-provider-azure/src/library/events-store-adapter.ts
MarcAstr0: See [this comment](https://github.com/boostercloud/booster/pull/1436#discussion_r1550392068).

packages/framework-provider-azure/src/library/events-store-adapter.ts
MarcAstr0: @gonzalogarciajaubert what do you think about @alvaroloes's suggestion?

packages/framework-provider-azure/src/library/events-store-adapter.ts
MarcAstr0: Simpler and more elegant. Suggestion applied in commit 11bbb506.

packages/framework-provider-azure/src/library/events-store-adapter.ts
MarcAstr0: Expanded log message in commit 11bbb506.

packages/framework-provider-azure/src/library/events-store-adapter.ts
MarcAstr0: Added the full `batchResponse` object to the `Error` in commit 11bbb506.

packages/framework-provider-azure/src/library/events-store-adapter.ts
MarcAstr0: @alvaroloes in commit f9793a60 this return was simplified following your suggestion. I spoke to @gonzalogarciajaubert and it appears there shouldn't be an issue with this (very unlikely to get a BigInt or an object with circular dependencies)

packages/framework-provider-azure/src/library/events-store-adapter.ts
alvaroloes: Oh I though we had it. It's fine then

packages/framework-provider-azure/src/library/events-store-adapter.ts
```diff
@@ -0,0 +1,138 @@
...
+    for (const eventListInChunk of chunksOfEventsInPartitionKey) {
+      const eventEnvelopesChunkWithCreatedAt = toEventEnvelopes(eventListInChunk)
+      envelopesWithCreatedAt.push(...eventEnvelopesChunkWithCreatedAt)
+      const inputOperations: Array<CreateOperationInput> = toInputOperations(eventEnvelopesChunkWithCreatedAt, config)
+      const batchResponse = await batchEvents(cosmosDb, config, inputOperations, partitionKey)
+
+      // Batch is transactional and will roll back all operations if one fails
+      if (batchResponse?.code !== 200) {
+        logger.error(`An error ocurred storing a batch of events. Batch response: ${batchResponse}`)
+        throw new Error(`Error ${batchResponse.substatus} storing events: ${batchResponse}`)
```
alvaroloes, line 44: Interpolating an object into a string like this doesn't produce the result you are specting. See the image I attached.
Captura de pantalla 2024-04-10 a las 16 31 28

So you either you interpolate every important property of `batchResponse` _individually_, or you interpolate something like `JSON.stringify(batchResponse)`

alvaroloes (approved):
Please see the comment about interpolation before merging

PR #1439, opened 2023-09-27
Consistent versioning for rimraf dependencies across packages

javiertoledo:
Updated rimraf to use version `^5.0.0` in all packages


packages/cli/src/templates/project/package-json.ts
```diff
@@ -47,7 +47,6 @@ export const template = `{
     "lint:fix": "eslint --quiet --fix --ext '.js,.ts' **/*.ts",
     "build": "ttsc -b tsconfig.json",
     "clean": "rimraf ./dist tsconfig.tsbuildinfo",
```
javiertoledo, line 49:
```suggestion
    "clean": "rimraf ./dist tsconfig.tsbuildinfo"
```

PR #1431, opened 2023-09-20
Upgrade for Node 18 compatibility

javiertoledo:
In this branch we're working on a big upgrade to remove deprecated dependencies and start using Node 18.

## TODO

- [x] Raised the node version to 18.18.0
- [x] Review why the core & cli packages unit tests are failing
- [x] Review why the AWS package unit tests are failing
- [x] Increase the node version for Azure Functions - currently set to node 14

## Checks
- [ ] Project Builds
- [ ] Project passes tests and checks
- [ ] Updated documentation accordingly

PR #1440, opened 2023-09-27
Closes #1118

### LINKED ISSUE #1118: CI/CD: Ensure all Node.js versions are supported
# Bug Report

## Current Behavior

There are times that unit tests fail on some Node.js versions. For example, CLI tests that check for environment variables complain on Node.js v18 and other multiple ones.

## Expected behavior

We should support a specific number of Node.js versions, unit tests should pass on all of them equally. This behavior should be working on the latest minor version of the major versions we support.

E.g. if we support v18, v16 and v14, the CI/CD process should work on the latest versions of those.

## Possible Solution

Modify the GitHub actions workflows so unit tests are run on the versions of Node.js we support

rishabh1S:
## Description
This pull request addresses Issue #1118  
Ensure support for multiple Node.js versions.

## Changes

- Updated the GitHub Actions workflow to support multiple Node.js versions.
- Added a strategy matrix for testing on Node.js versions 14.x, 16.x, and 18.x.

### DIFF
--- a/.github/workflows/re_test-integration-prepare.yml
+++ b/.github/workflows/re_test-integration-prepare.yml
@@ -28,7 +28,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node }}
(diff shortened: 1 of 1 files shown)


javiertoledo:
Thanks for the PR, but it seems that the project is not currently working on Node 18, so we'll need to keep this on hold a little bit until it's fixed. We're working on the node18 upgrade in #1431

PR #1444, opened 2023-09-29
Upgrade node 18 -  Upgrade Azure Cosmos to ^4.0.0

santigamo:
## Description
Applied the necessary changes to update the Azure Cosmos dependencies to the latest version

## Checks
- [x] Project Builds


packages/cli/test/commands/deploy.test.ts
```diff
@@ -12,6 +12,7 @@ import * as packageManagerImpl from '../../src/services/package-manager/live.imp
 import * as configService from '../../src/services/config-service'
 import * as projectChecker from '../../src/services/project-checker'
 import { makeTestPackageManager } from '../services/package-manager/test.impl'
+import { root } from '../configRoot'
```
javiertoledo, line 15: I don't fully understand how this `configRoot` file and manually setting the path `/someDir` in the tests works. Could you explain why you needed to do this and how the solution works?

This is not your fault, but the CLI tests are currently using way too much "real stuff" to call them "unit tests". Using actual environment variables and depending on certain files in the filesystem is a design flaw. It makes tests slower and less reliable, and the root reason for the current issue with the CLI unit tests likely comes from this faulty design... If this is a patch to workaround this problematic design, that's okay for now, because we want to add support for Node 18 ASAP, but we should probably consider re-designing the CLI unit tests to make sure they perform no actual syscalls nor access real environment variables... We can leave this for another future PR in any case :-D

packages/cli/test/services/environment.test.ts
```diff
@@ -21,7 +21,6 @@ describe('environment service', (): void => {
   describe('currentEnvironment', (): void => {
     it('get current environment: testing', async () => {
       process.env.BOOSTER_ENV = 'testing'
-      stub(process.env, 'BOOSTER_ENV').value('testing')
```
javiertoledo, line 24: We shouldn't set environment variables in unit tests, especially if we don't restore them in an after-block. Maybe we could use here the `replace` and `restore` functions from `sinon` to avoid setting the actual variable and ensure the changes are reset afterward. I think that something like this should work:

```typescript
  ...
  beforeEach(() => {
    replace(process.env, 'BOOSTER_ENV', 'testing')
  }

  afterEach(() => {
    restore()
  }
  ...
```

javiertoledo (changes requested):
Great work @santigamo! I left a couple of comments with some details I'd love to discuss a bit more before merging.

packages/cli/test/commands/deploy.test.ts
santigamo: I'm not sure why, but at least from the upgrade node branch, some tests were failing due to none command generated with oclif being found. Debugging the test I found out that the configuration was not being loaded so the only way I found (that is also used in the oclif repo tests) was creating a const that joins `__dirname` with the parent folder and use this variable to imperatively load the config. So I created the configRoot helper to not init the const in each file.

packages/cli/test/services/environment.test.ts
santigamo: I agree, test changed but at least we need to create the variable empty to being able to replace it later.

Are you okey with that?

```
...
beforeEach(() => {
        process.env.BOOSTER_ENV = ''
        replace(process.env, 'BOOSTER_ENV', 'testing')
      })

      afterEach(() => {
        restore()
      })
...
```

javiertoledo: Oh I see T_T

If that's the case, there's not much difference between setting it to `''` or any other value.

Anyways, inspecting the code again, I found that we were already handling the environment variables state in this code block between lines 7 and 19:

```typescript
  beforeEach(() => {
    if (process.env.BOOSTER_ENV !== undefined) {
      formerValue = process.env.BOOSTER_ENV
    }
    delete process.env.BOOSTER_ENV
  })
  afterEach(() => {
    process.env.BOOSTER_ENV = formerValue
    restore()
  })
```

With this code in place, I'd say that your initial solution of setting the BOOSTER_ENV value without using `replace` and `restore` was correct. Sorry about that. I'll add a commit to your branch removing the unnecessary calls to these methods

packages/cli/test/services/environment.test.ts
```diff
@@ -38,12 +37,15 @@ describe('environment service', (): void => {
...
     describe('process.env.BOOSTER_ENV set', (): void => {
       beforeEach(() => {
-        process.env.BOOSTER_ENV = 'testing'
-        stub(process.env, 'BOOSTER_ENV').value('testing')
+        process.env.BOOSTER_ENV = ''
+        replace(process.env, 'BOOSTER_ENV', 'testing')
+      })
+
+      afterEach(() => {
+        restore()
```
javiertoledo, line 45: This is not actually needed because we're handling the environment variable state manually at the beginning of this file

```suggestion
        process.env.BOOSTER_ENV = 'testing'
```

javiertoledo (approved):
Let's merge this!

PR #1446, opened 2023-10-02
Fix event api for events and notifications not found

gonzalojaubert:
When we use the event API to get events Booster try to creates instances from them. This could cause issues if the event class had been removed. If the event class doesn’t exist Booster should not create the instance and the content will be a plain object

Also, when we use the event API to get notifications events Booster is not able to find the event class. Also, if the class doesn’t exist Booster should not fail but keep the event as a plain object.


javiertoledo (approved):
LGTM!

alvaroloes (approved):

PR #1452, opened 2023-10-03
Upgrade CDK to v2

javiertoledo:
## Description

This is an exploration branch to upgrade the CDK to v2 for AWS.

**Update:** The CDK v2 comes with major changes in how CDK applications are handled. The CDKToolkit has been extracted into a CLI tool, so upgrading to v2 would force us to rewrite a significant portion of our deployment and nuke scripts, and the new implementation would require either [running shell commands or using alpha packages](https://github.com/aws/aws-cdk-rfcs/issues/300). Given the situation, we want to explore the cost of migrating the AWS provider to the CDKTF before going further on this path.

## Checks
- [ ] Project Builds


javiertoledo:
Closing as this no longer seem to make sense for the future of the project. See #1477 for more detail.

PR #1481, opened 2023-10-25
Allow define projection fields in read model queries

gonzalojaubert:
Add the `select` method to project which fields you want to get in your read model query:

```typescript
@Command({
  authorize: 'all',
})
export class GetProductsCount {
  public constructor(readonly filters: Record<string, any>) {}

  public static async handle(): Promise<unknown> {
    const searcher = Booster.readModel(ProductReadModel)
            .filter({
              sku: { contains: 'toy' },
              or: [
                {
                  description: { contains: 'fancy' },
                },
                {
                  description: { contains: 'great' },
                },
              ],
            })
            .select(['sku', 'description'])
            .skipInstance(true)
    const result = await searcher.search()
    return { count: result.length }
  }
}
```

The searcher result using `skipInstance` will generate an array of objects with the `sku` and `description` fields of the `ProductReadModel` read model. If you don't use `skipInstance` the result will be an array of `ProductReadModel` instances.

Note: Using `skipInstance` will skip any Read Models migrations that need to be applied to the result. If you need to apply migrations to the result, don't use `skipInstance`.


packages/framework-types/src/searcher.ts
```diff
@@ -114,11 +129,40 @@ export class Searcher<TObject> {
...
+// type KeyOf<
+//   TType,
+//   R = {
+//     [K in keyof TType & (string | number | bigint | boolean | null | undefined)]: TType[K] extends object
+//       ? `${K}` | `${K}.${KeyOf<TType[K]>}`
+//       : `${K}`
+//   }[keyof TType & (string | number)]
+// > = R
+
+export type ProjectionFor<TType> = Array<KeyOf<TType>>
```
alvaroloes, line 164: If I understood correctly, this types is an array of all the properties of the object, including sub-properties. 

If so, this type can consume a lot of memory and time when compiling. Newer versions of Typescript allows you to do some easier black magic which is much much faster and requires much less memory (up to 6 times less)

```suggestion
type Paths<T> = T extends object ? { [K in keyof T]: `${Exclude<K, symbol>}${'' | `.${Paths<T[K]>}`}` }[keyof T] : never

export type ProjectionFor<TType> = Array<Paths<TType>>
```

MarcAstr0: Suggestion applied in commit e2d6ac5.

MarcAstr0:
**Important Update**

Changes have been made so that `select` can be applied not only in code but also when _querying read models via GraphQL_.

The [`getFields`](https://github.com/Optum/booster/blob/09d7db60d6eb540e98936a40b94870b1d50b61ec/packages/framework-core/src/services/graphql/graphql-generator.ts#L225) method in the `GraphQLGenerator` class will build the `select` that is passed on to the searcher based on what fields came in the GraphQL query. It will also identify which fields are arrays of objects so that the selected properties are treated correctly when querying the database.

The [`buildProjections`](https://github.com/Optum/booster/blob/09d7db60d6eb540e98936a40b94870b1d50b61ec/packages/framework-provider-azure/src/helpers/query-helper.ts#L222) method found in [`query-helper.ts`](https://github.com/Optum/booster/blob/09d7db60d6eb540e98936a40b94870b1d50b61ec/packages/framework-provider-azure/src/helpers/query-helper.ts) of the Azure provider will build the SQL query and handle object and array fields accordingly.

To be able to keep the structure of object properties and their sub-properties, when fetching those fields from the database, they're fetched in dot notation, e.g., `root.property.subProperty`. The helper [`nestProperties`](https://github.com/Optum/booster/blob/09d7db60d6eb540e98936a40b94870b1d50b61ec/packages/framework-provider-azure/src/helpers/query-helper.ts#L255) method is then applied to the results to convert it to nested properties, e.g., `{root: {property: {subProperty: "some value"}}}`.

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -198,13 +221,78 @@ export class GraphQLGenerator {
...
+      parentType: GraphQLObjectType | any
+    ): string[] => {
+      let subFields: string[] = []
+      if (selectionSet && selectionSet.selections) {
+        selectionSet.selections.forEach((selection: any) => {
+          const fieldName = selection.name.value
+          const field = parentType.getFields()[fieldName]
+
+          if (!field) {
+            console.error(`Field ${fieldName} not found on type ${parentType.name}. Skipping.`) // @TODO: remove
```
alvaroloes, line 247: Just checking: there is a TODO here that says remove... I don't fully understand this yet, but we are logging an error but saying "skipping".

My question is: Is this truly an error or some condition that can appear and it is ok to continue?

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -198,13 +221,78 @@ export class GraphQLGenerator {
...
+      ) as unknown as GraphQLObjectType
+
+      if (fieldNode.selectionSet) {
+        fields = fields.concat(extractFields(fieldNode.selectionSet, [], rootType))
+      }
+    })
+
+    return fields
+  }
 }
```
alvaroloes, line 288: Amazing. I though it was going to be much easier to get the selection set from the graphQL query, but you needed to implement this crazy recursive method. Great job!

packages/framework-core/src/services/graphql/graphql-generator.ts
```diff
@@ -90,10 +104,19 @@ export class GraphQLGenerator {
...
+      let select: ProjectionFor<unknown> | undefined = fields.length > 0 ? fields : undefined
       if (info?.fieldName === `List${pluralize(readModelClass.name)}`) {
         isPaginated = true
+        if (select) {
+          select = select
+            .map((field: string) => {
+              return field.split('.').slice(1).join('.')
+            })
+            .filter((str: string) => str.trim().length > 0) as ProjectionFor<unknown>
+        }
```
alvaroloes, line 117: I'm still unsure why you are doing this map. We are removing the first component of every projection path. What for? 
And if we are always removing, why returning it in the first place? We can make the `this.getFields` method not to return it.

packages/framework-integration-tests/package.json
```diff
@@ -111,11 +112,12 @@
...
     "integration/aws-load": "echo 'Skipping integration tests.' # TS_NODE_PROJECT=\"./tsconfig.integration.json\" TESTED_PROVIDER=AWS AWS_SDK_LOAD_CONFIG=true BOOSTER_ENV=production mocha --forbid-only --exit --config \"integration/provider-unaware/load/.mocharc.yml\" \"integration/provider-unaware/load/**/*.load.ts\"",
     "integration/aws-nuke": "TS_NODE_PROJECT=\"./tsconfig.integration.json\" AWS_SDK_LOAD_CONFIG=true BOOSTER_ENV=production mocha --forbid-only --exit --config \"integration/provider-specific/aws/nuke/.mocharc.yml\" \"integration/provider-specific/aws/nuke/**/*.integration.ts\"",
-    "integration/local": "TS_NODE_PROJECT=\"./tsconfig.integration.json\" npm run integration/local-start && npm run integration/local-end-to-end && npm run integration/local-stop",
+    "integration/local": "concurrently \"npm run integration/local-start\" \"npm run integration/local-delayed-end-to-end\"",
```
alvaroloes, line 115: 🫨 What is this black magic of "concurrently"? I normally do "script_1; script_2"... Is this word an oficial operator in bash?

packages/framework-provider-azure/src/helpers/query-helper.ts
```diff
@@ -214,3 +218,83 @@ function toLocalSortFor(
...
+    groupedFields[root].push(field)
+  })
+
+  return Object.keys(groupedFields)
+    .map((root: string): string => {
+      const fields = groupedFields[root]
+      if (root.endsWith('[]')) {
+        const arrayRoot: string = root.slice(0, -2)
+        const subFields = fields.map((f: string) => f.replace(`${root}.`, 'item.')).join(', ')
+        return `ARRAY(SELECT ${subFields} FROM item IN c.${arrayRoot}) AS ${arrayRoot}`
```
alvaroloes, line 243: Out-of-curiosity question: Who adds the `[]` to the fields path? Is it GraphQL?

Another question: What would happen if the array field is in the middle of the path? for example: `lastPurchase.items[].id`? Or is this something not supported?

packages/framework-provider-azure/src/helpers/query-helper.ts
```diff
@@ -214,3 +218,83 @@ function toLocalSortFor(
...
+      } else if (fields.length === 1 && !fields[0].includes('.')) {
+        // Simple field
+        return `c.${fields[0]}`
+      } else {
+        // Nested object fields
+        return fields.map((f: string) => `c.${f} AS "${f}"`).join(', ')
```
alvaroloes, line 249: In the previous `else if` branch you checked if the field didn't contain dots (`.`). Shouldn't we also check here?
Another question: Why in this case we add an `AS` but not in the previous branch?

packages/framework-provider-azure/src/helpers/query-helper.ts
```diff
@@ -214,3 +218,83 @@ function toLocalSortFor(
...
+        const keys = key.split('.')
+        setNestedProperty(output, keys, value)
+      }
+    }
+  }
+
+  function processArray(arr: any[]): any[] {
+    return arr.map((item: any): any => {
+      if (Array.isArray(item)) {
+        return processArray(item)
```
alvaroloes, line 282: I can't grasp my head around this.. When we do a query to the DB, we either fetch one item (`{'a.b.c': 42}`) or an array of items (`[{'a.b.c': 42}, {'a.b.c': 43}, ...]`. Why/when would we get an array of arrays?

website/docs/03_architecture/06_read-model.mdx
```diff
@@ -310,6 +310,11 @@ export class MessageReadModel {
 }
 ```
 
+## Projecting fields from a Read Model
```
alvaroloes, line 313: In the docs, I would change the word `Project` and its derivatives to "Selecting fields" or similar. The reason is that we already have the concept "Project" associated to read model projections, and it can be confusing to the users if we also use the same word for this. 

Besides, in the code we use `.select`, so it makes sense to talk about is using the "Select fields" expresion

alvaroloes (commented):
What an great work @MarcAstr0 🥇 🥇🥇

packages/framework-core/src/services/graphql/graphql-generator.ts
MarcAstr0: @alvaroloes this `map` takes care of the `List<Read Model name>` GraphQL queries (paginated queries), which will wrap the fields inside an `items` property. For example:

```graphql
query ListMyReadModels {
  items {
    foo
    bar
  }
}
```

Calling `getFields` on this query will return `['items[].foo', 'items[].bar']`, but our read model, in this case, `MyReadModel`, doesn't have an `items` property when stored on the database, it has `foo` and `bar` properties, so the actual fields we want to select are these two. Every time one of these types of queries comes, it will have an `items[]` prepended. This `map` cleans the `select`.

packages/framework-integration-tests/package.json
MarcAstr0: [concurrently](https://github.com/open-cli-tools/concurrently) fixes a very annoying problem that occurred when running the integration tests on the local provider. Ideally, it was as simple as running `integration/local-start`, `integration/local-end-to-end`, and `integration/local-stop` in sequence because when `integration/local-start` finishes, it leaves the server running in the background until `integration/local-stop` is executed. However, for some unknown reason, the server was being stopped/killed right in the middle of `integration/local-end-to-end`. I observed that when keeping `integration/local-start` open (i.e., removing the `--exit` flag from the `mocha` command) and then running `integration/local-end-to-end` in another terminal window the end-to-end tests finished just fine. So what concurrently does is "simulate" having those two terminals open. The `sleep` is added to `integration/local-end-to-end` so that it waits enough time for the server to start since concurrently spawns the commands it receives simultaneously.

packages/framework-provider-azure/src/helpers/query-helper.ts
MarcAstr0: @alvaroloes the `[]` is added by the `getFields` method in `GraphQLGenerator` (see [here](https://github.com/Optum/booster/blob/09d7db60d6eb540e98936a40b94870b1d50b61ec/packages/framework-core/src/services/graphql/graphql-generator.ts#L257)).

Array fields in the middle of the path were not initially supported but were added in commit 8726201. However, only array fields that are part of a top-level object property are supported for selecting properties. Arrays nested deeper are not supported at the moment (in other words, `foo.items[].bar` is supported but not `foo.bar.items[].somethingElse`)

packages/framework-provider-azure/src/helpers/query-helper.ts
MarcAstr0: @alvaroloes the previous `else if` condition was when there was a single field and that it didn't contain a dot, which translates into meaning that the field is a top-level field. The `AS` is added for the `else` branch because it indicates nested properties in the read model, e.g., `foo.bar` indicates `{"foo" : {"bar": "some value"}}`. The `AS` is used to retrieve the entire path in your DB query result, e.g., `{"foo.bar": "some value"}` and then the `nestProperties` method converts it back to its original structure. The reason for this is that if one queries without the `AS`, CosmosDB doesn't retain the object structure. In the example, querying for `foo.bar` without the `AS` would return `{"bar": "some value"}`

packages/framework-provider-azure/src/helpers/query-helper.ts
MarcAstr0: @alvaroloes it shouldn't be necessary for the DB-specific case, as you correctly point out, but the `nestProperties` and its helper methods were made to be generic.

packages/framework-core/src/services/graphql/graphql-generator.ts
MarcAstr0: @alvaroloes it is not an error; it is expected, actually. It was used for debugging during development to indicate when a field was not part of the selection set. That's why the `TODO` was there but it was removed in commit f498f423. :wink:

website/docs/03_architecture/06_read-model.mdx
MarcAstr0: @alvaroloes changed in commit c4b7ca6c. Also, in the same commit, a warning message was added indicating the limitations of selecting fields inside arrays, related to [this comment](https://github.com/boostercloud/booster/pull/1481#discussion_r1616552349)

packages/framework-core/src/services/graphql/graphql-generator.ts
alvaroloes: @MarcAstr0 Ahhh! Got it, thanks. Do you mind writing a comment saying that please 🙏 ? Something like "Removing `items[].` from the selection set". That way future us or other devs will have an easier time understanding it. Thanks a lot!

packages/framework-integration-tests/package.json
alvaroloes: All right. Thanks for the explanation. Wondering why that started happening now.

packages/framework-provider-azure/src/helpers/query-helper.ts
alvaroloes: Perfect, thanks!

packages/framework-provider-azure/src/helpers/query-helper.ts
alvaroloes: Woah! Nice trick what you did with the AS there. Thanks!

packages/framework-provider-azure/src/helpers/query-helper.ts
alvaroloes: All right

website/docs/03_architecture/06_read-model.mdx
alvaroloes: Great!

packages/framework-provider-local/src/services/read-model-registry.ts
```diff
@@ -44,7 +50,48 @@ export class ReadModelRegistry {
...
+        if (part.endsWith('[]')) {
+          const arrayField = part.slice(0, -2)
+          if (!Object.prototype.hasOwnProperty.call(currentLevel, arrayField)) {
+            currentLevel[arrayField] = {}
+          }
+          currentLevel = currentLevel[arrayField]
+          isArrayField = true
+        } else {
+          if (i === parts.length - 1) {
+            if (isArrayField) {
```
alvaroloes, line 70: A simplification here: 
```suggestion
        } else if (i === parts.length - 1 && isArrayField) {
```

alvaroloes (approved):
💯 Thanks for all the clarifications

packages/framework-provider-local/src/services/read-model-registry.ts
MarcAstr0: This suggestion broke the unit tests, so we'll ignore it for now.

PR #1484, opened 2023-10-27
Add missing link to website sidebar

DKarandikar:
## Description

Add link to doc which is missing from website


javiertoledo (approved):
Oh! That would definitely explain why most people are surprised when we mention queries, it has been hidden from the documentation menu all this time! 🤦🏼 

Thanks a lot for the edit!

PR #1491, opened 2023-11-02
Refactor: reworked core exports

javiertoledo:
## Description

Cleanup of exported methods of the core package to avoid extra levels of indirection. The core index interface has not changed, but I've moved the implementation of the static methods from the `booster.ts` file to the `index.ts` to avoid a level of indirection. The `BoosterApp` type was made redundant as it was just describing the `Booster` class signature.

## Changes

- [x] Moved static methods implementation to `index.ts`
- [x] Export the `Booster` object directly without proxying it through the `BoosterApp` type
- [x] Remove the `BoosterApp` type, as the `Booster` object already exports the correct signature. This type was used only to type-check loading user apps from the CLI, and while the type was exported, it was not used in practice. Users use the `Booster` singleton exported by the core package instead.
- [x] Remove no longer used methods from the `Booster` class, as they were never exported; thus, removing them is not a breaking change.
- [x] Fix all the tests


packages/framework-core/src/sensor/health/booster-health-service.ts
```diff
@@ -13,10 +13,14 @@ import { defaultBoosterHealthIndicators } from './health-indicators'
...
 
+/**
+ * This class is in charge of handling the health check requests
+ * and dispatching the health checks to the corresponding health indicators
+ */
 export class BoosterHealthService {
   constructor(readonly config: BoosterConfig) {}
 
-  public async boosterHealth(request: any): Promise<unknown> {
+  public async boosterHealth(request: unknown): Promise<unknown> {
```
NickSeagull, line 23: aaaawwww yeah 😍

NickSeagull (approved):
LGTM

PR #1525, opened 2024-02-27
Fix annotation for schema migrations in data-migrations.md

JorgeRdg:
## Description
Docs show `@Migrates` as the annotation to use for schema migrations, but that's not the correct one. This PR fixes it to indicate it's `@SchemaMigration` the one to use.


NickSeagull (approved):
Thanks!

PR #1527, opened 2024-03-15
Track processed events in Azure provider

MarcAstr0:
## Description

The changes in this PR aim to correct a behavior noticed in the Azure provider, where the Cosmos DB change feed would get triggered more than once for the same event, resulting in the same event being processed twice. This behavior has been seen to happen during load-balancing scenarios.

To correct this duplicate-event behavior, a new processed events container is added to Cosmos DB to store the IDs of events that the event processor has processed. The processor will query this container for the event IDs and verify the event has not been processed (i.e., it can't find its ID in this new processed-events container). If an event with the same ID is found, to avoid duplication, the event processor will ignore it and log a warning indicating the duplication.

The IDs store in the processed events container have a time-to-live (TTL) and will get cleared automatically.

## Changes

- Modifies `framework-provider-azure-infrastructure` package adding the new container for processed events.
- Adds a configurable `processedEventsTtl` to `BoosterConfig` for setting the TTL of the event IDs in the processed events container (defaults to 300 seconds, i.e., 5 minutes).
- Makes the necessary changes to the `BoosterEventProcessor` in the `framework-core` package for querying and storing the processed event IDs.
- Modifies the adapters in the `framework-provider-azure` package for interacting with the processed events container.
- For backwards compatibility, adds dummy functions to the (now deprecated) AWS and the local provider packages that will skip any verification for duplicate events since this behavior hasn't been reported in these providers.

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-core/src/booster-event-processor.ts
+++ b/packages/framework-core/src/booster-event-processor.ts
@@ -24,8 +24,11 @@ export class BoosterEventProcessor {
    */
   public static eventProcessor(eventStore: EventStore, readModelStore: ReadModelStore): EventsStreamingCallback {
     return async (entityName, entityID, eventEnvelopes, config) => {
+      const logger = getLogger(config, 'BoosterEventDispatcher#eventProcessor')
+      const unprocessedEvents = await BoosterEventProcessor.filterProcessed(config, eventEnvelopes, eventStore)
+      logger.debug(`Total events that haven't been processed: ${unprocessedEvents.length}`)
       const eventEnvelopesProcessors = [
-        BoosterEventProcessor.dispatchEntityEventsToEventHandlers(eventEnvelopes, config),
+        BoosterEventProcessor.dispatchEntityEventsToEventHandlers(unprocessedEvents, config),
       ]
 
       // Read models are not updated for notification events (events that are not related to an entity but a topic)
@@ -34,9 +37,30 @@ export class BoosterEventProcessor {
           BoosterEventProcessor.snapshotAndUpdateReadModels(config, entityName, entityID, eventStore, readModelStore)
         )
       }
+      // Store events that were just processed
+      await eventStore.storeProcessedEvents(unprocessedEvents)
 
       await Promises.allSettledAndFulfilled(eventEnvelopesProcessors)
     }
+  }
+
+  private static async filterProcessed(
+    config: BoosterConfig,
+    eventEnvelopes: Array<EventEnvelope>,
+    eventStore: EventStore
+  ): Promise<Array<EventEnvelope>> {
+    const logger = getLogger(config, 'BoosterEventDispatcher#filterProcessed')
+    const filteredResults = await Promise.all(
+      eventEnvelopes.map(async (eventEnvelope) => {
+        const result = await eventStore.searchProcessed(eventEnvelope)
+        if (!result || result.length > 0) {
+          logger.debug('Event has already been processed. Skipping.', eventEnvelope)
+        }
+        return result?.length === 0
+      })
+    )
+
+    return eventEnvelopes.filter((_, index) => filteredResults[index])
   }
 
   private static async snapshotAndUpdateReadModels(
--- a/packages/framework-core/src/booster-event-processor.ts
+++ b/packages/framework-core/src/booster-event-processor.ts
@@ -1,11 +1,11 @@
 import {
-  TraceActionTypes,
   BoosterConfig,
   EventEnvelope,
   EventHandlerGlobalError,
   EventHandlerInterface,
   EventInterface,
   Register,
+  TraceActionTypes,
   UUID,
 } from '@boostercloud/framework-types'
 import { EventStore } from './services/event-store'
@@ -24,9 +24,7 @@ export class BoosterEventProcessor {
    */
   public static eventProcessor(eventStore: EventStore, readModelStore: ReadModelStore): EventsStreamingCallback {
     return async (entityName, entityID, eventEnvelopes, config) => {
-      const logger = getLogger(config, 'BoosterEventDispatcher#eventProcessor')
       const unprocessedEvents = await BoosterEventProcessor.filterProcessed(config, eventEnvelopes, eventStore)
-      logger.debug(`Total events that haven't been processed: ${unprocessedEvents.length}`)
       const eventEnvelopesProcessors = [
         BoosterEventProcessor.dispatchEntityEventsToEventHandlers(unprocessedEvents, config),
       ]
@@ -54,7 +52,7 @@ export class BoosterEventProcessor {
       eventEnvelopes.map(async (eventEnvelope) => {
         const result = await eventStore.searchProcessed(eventEnvelope)
         if (!result || result.length > 0) {
-          logger.debug('Event has already been processed. Skipping.', eventEnvelope)
+          logger.warn('Event has already been processed. Skipping.', eventEnvelope)
         }
         return result?.length === 0
       })
--- a/packages/framework-core/src/booster-event-processor.ts
+++ b/packages/framework-core/src/booster-event-processor.ts
@@ -24,9 +24,9 @@ export class BoosterEventProcessor {
    */
   public static eventProcessor(eventStore: EventStore, readModelStore: ReadModelStore): EventsStreamingCallback {
     return async (entityName, entityID, eventEnvelopes, config) => {
-      const unprocessedEvents = await BoosterEventProcessor.filterProcessed(config, eventEnvelopes, eventStore)
+      const eventsNotDispatched = await BoosterEventProcessor.filterDispatched(config, eventEnvelopes, eventStore)
       const eventEnvelopesProcessors = [
-        BoosterEventProcessor.dispatchEntityEventsToEventHandlers(unprocessedEvents, config),
+        BoosterEventProcessor.dispatchEntityEventsToEventHandlers(eventsNotDispatched, config),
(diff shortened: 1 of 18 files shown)


packages/framework-core/src/booster-event-processor.ts
```diff
@@ -34,11 +35,32 @@ export class BoosterEventProcessor {
           BoosterEventProcessor.snapshotAndUpdateReadModels(config, entityName, entityID, eventStore, readModelStore)
         )
       }
+      // Store events that were just processed
+      await eventStore.storeProcessedEvents(unprocessedEvents)
 
       await Promises.allSettledAndFulfilled(eventEnvelopesProcessors)
     }
```
alvaroloes, line 42: @MarcAstr0 The operation of checking if the event and the insertion of that event in the table must be one atomic operation. If not, we could still process the event twice because the second event could come _after_ we searched for it but _before_ we store it into the processed table.

Also, we need to insert the event as soon as the event starts processing, not after it has been processed (for the same reason as above). I think it is a good idea to **change the name of the table to `dispatched_events`**. This way, it is clear that if an event is there, it has been dispatched, and it can be processed or being processed.

So the idea would be to only have one method in the `EventStore` called like `startEventProcessing` (the same method would be in each provider). That method will just try to insert the event in the `dispatched_events` table. If it succeeds, it returns true telling that you can proceed with the processing. If the insertion fails because of a "duplicate ID key" violation, it means the event was already dispatched, so the method will return false, meaning that event can't be processed (again).

This way, we achieve the insertion and "search for presence" operations in one atomic operation, avoiding any possible race condition when we have the two operations separated.

MarcAstr0: @alvaroloes take a look at the latest changes, especially commit 7a26a48c3c3cf83d5733f6d11af36d974b8ea8c5. :wink:

packages/framework-core/src/services/event-store.ts
```diff
@@ -73,6 +73,18 @@ export class EventStore {
     }
   }
 
+  @Trace(TraceActionTypes.CUSTOM)
+  public async storeDispatchedEvent(eventEnvelope: EventEnvelope) {
```
alvaroloes, line 77: Not sure how this passes lint, but you missed the return type `Promise<boolean>`

packages/framework-core/src/services/event-store.ts
```diff
@@ -73,6 +73,18 @@ export class EventStore {
...
+    const logger = getLogger(this.config, 'EventStore#storeDispatchedEvent')
+    try {
+      logger.debug('Storing event in the dispatched event store:', eventEnvelope)
+      return await this.config.provider.events.storeDispatched(eventEnvelope, this.config)
+    } catch (e) {
+      logger.error('Could not store dispatched event: ', eventEnvelope)
+      return
```
alvaroloes, line 84: I think it is a good idea what you are doing here. I mean, not to stop the processing of the event if there was an error. I would just return true and say so in the log. Something like: 
```suggestion
      logger.error('Could not store dispatched event. Continue its processing.', eventEnvelope)
      return true
```

packages/framework-provider-azure/src/library/events-adapter.ts
```diff
@@ -180,3 +180,30 @@ export async function storeSnapshot(
...
+  } catch (e) {
+    if (e.code === 409) {
+      // If an item with the same ID already exists in the container, it will return a 409 status code.
+      // See https://learn.microsoft.com/en-us/rest/api/cosmos-db/http-status-codes-for-cosmosdb
+      logger.warn('[EventsAdapter#storeDispatchedEvent] Event has already been dispatched', eventEnvelope.id)
+    } else {
+      logger.error('[EventsAdapter#storeDispatchedEvent] Error storing dispatched event', e)
+    }
+  }
+  return false
```
alvaroloes, line 208: This will return **false** for any kind of error, which will prevent processing the event. I don't think that's a good idea. If there was a error storing the event in the dispatched table different than 409, we want to keep processing the event. It would be like a fallback to the actual behavior.

I would return false in case of a 409 and let any other error bubble up so that the core can decide what to do (which you already did in the method `EventStore.storeDispatchedEvent`

packages/framework-provider-azure/src/library/events-adapter.ts
```diff
@@ -180,3 +180,30 @@ export async function storeSnapshot(
...
+      .container(config.resourceNames.dispatchedEventsStore)
+      .items.create({
+        eventId: eventEnvelope.id,
+      })
+    return true
+  } catch (e) {
+    if (e.code === 409) {
+      // If an item with the same ID already exists in the container, it will return a 409 status code.
+      // See https://learn.microsoft.com/en-us/rest/api/cosmos-db/http-status-codes-for-cosmosdb
+      logger.warn('[EventsAdapter#storeDispatchedEvent] Event has already been dispatched', eventEnvelope.id)
```
alvaroloes, line 203: You already log a warning in the "core" logic (EventProcessor.filterDispatched), so I think this log could be `info` or `debug` level. This way, we don't get the same warning twice

packages/framework-types/src/config.ts
```diff
@@ -241,15 +244,20 @@ export class BoosterConfig {
       }
     }
   }
+
+  // TTL for events stored in dispatched events table. Default to 5 minutes (i.e., 300 seconds).
+  public dispatchedEventsTtl = 300
```
alvaroloes, line 249: Do you mind placing this along with the other properties, before all the methods? I will be easier to find there as it is together with the other props

packages/framework-types/src/provider.ts
```diff
@@ -144,6 +144,15 @@ export interface ProviderEventsLibrary {
     snapshotEnvelope: NonPersistedEntitySnapshotEnvelope,
     config: BoosterConfig
   ): Promise<EntitySnapshotEnvelope>
+
+  /**
+   * Stores events envelopes that have been dispatched in the dispatched events table
+   *
+   * @param eventEnvelope - The `Array<EventEnvelope>` to store
```
alvaroloes, line 151: It is not an array, it is a single `EventEnvelope`

packages/framework-types/src/provider.ts
```diff
@@ -144,6 +144,15 @@ export interface ProviderEventsLibrary {
...
+   * @param config - The Booster configuration object
+   * @returns `true` if the dispatched event was stored, otherwise, returns `false`
```
alvaroloes, line 153: I would be more restrictive and explicit here and say:
```suggestion
   * @returns `true` if the dispatched event was stored, `false` if the event already exists, and will throw in case of any other error
```

alvaroloes (commented):
I left some nitpicking comments and one a bit more important about what to return when any other error happens while storing the dispatched event

MarcAstr0:
@alvaroloes all of your comments have been addressed in commit 8b06598a412b227c4b93d4152d4ff6bba64a3c5a :wink: :+1:

packages/framework-core/src/services/event-store.ts
```diff
@@ -73,6 +73,18 @@ export class EventStore {
...
   }
 
+  @Trace(TraceActionTypes.CUSTOM)
+  public async storeDispatchedEvent(eventEnvelope: EventEnvelope): Promise<boolean | undefined> {
+    const logger = getLogger(this.config, 'EventStore#storeDispatchedEvent')
+    try {
+      logger.debug('Storing event in the dispatched event store:', eventEnvelope)
+      return await this.config.provider.events.storeDispatched(eventEnvelope, this.config)
+    } catch (e) {
+      logger.debug('Could not store dispatched event. Continue its processing.', eventEnvelope)
```
alvaroloes, line 83: I missed this before: Please log the error too, if not we will swallow it and it would be really hard to find out in the deployed environments. Something like: 
```suggestion
      logger.debug('Could not store dispatched event. Continue its processing.', {error: e, eventEnvelope} )
```

packages/framework-provider-azure/src/library/events-adapter.ts
```diff
@@ -180,3 +180,31 @@ export async function storeSnapshot(
...
+    return true
+  } catch (e) {
+    if (e.code === 409) {
+      // If an item with the same ID already exists in the container, it will return a 409 status code.
+      // See https://learn.microsoft.com/en-us/rest/api/cosmos-db/http-status-codes-for-cosmosdb
+      logger.debug('[EventsAdapter#storeDispatchedEvent] Event has already been dispatched', eventEnvelope.id)
+      return false
+    } else {
+      logger.error('[EventsAdapter#storeDispatchedEvent] Error storing dispatched event', e)
+      return true
```
alvaroloes, line 207: Here we should let the error bubble up so that the core can act upon it if needed (or log it), as other methods do. Basically rethrow it:
```suggestion
      throw e
```

alvaroloes (approved):
Approved, but please take a look at the error "bubble up" comment and the login thing, so that it is easier to debug in prod.

packages/framework-core/src/services/event-store.ts
MarcAstr0: Changed in commit 1eb76723145dabcaf935257523f9743dc66f23e7

packages/framework-provider-azure/src/library/events-adapter.ts
MarcAstr0: Changed in commit 1eb76723145dabcaf935257523f9743dc66f23e7

PR #1539, opened 2024-07-03
Calculated fields

MarcAstr0:
## Description

Due to the changes introduced in version [2.11.0](https://github.com/boostercloud/booster/releases/tag/v2.11.0), this PR introduces the `@CalculatedField` decorator for read model fields, which are calculated using getters using other properties of the read model. e.g.,

```typescript
@ReadModel({
  authorize: 'all',
})
export class PersonReadModel {
  public constructor(
    readonly id: UUID,
    readonly firstName: string,
    readonly lastName: string
  ) {}

  @CalculatedField(['firstName', 'lastName'])
  public get fullName(): Promise<string> {
    return Promise.resolve(`${this.firstName} ${this.lastName}`)
  }
}
```

This decorator receives a list of the read model's properties as an input parameter. The calculated field uses these properties to calculate its value. The decorator ensures that these properties are fetched from the database so that the calculated field is calculated correctly, even if the user did not explicitly request these dependencies.

## Changes

* Add the `CalculatedField` decorator to the `framework-core` package, which stores the calculated fields and their dependencies in the read model's metadata.
* Modify the `search` method in the `BoosterReadModelsReader` class so it handles calculated fields accordingly.
* Add necessary helper methods for working with the calculated fields and instantiating the read model class.
* Fix broken code in the `cli` package due to updated Node types.
* Updated documentation.

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-core/src/decorators/read-model.ts
+++ b/packages/framework-core/src/decorators/read-model.ts
@@ -45,14 +45,18 @@ export function ReadModel(
   }
 }
 
+interface CalculatedFieldOptions {
+  dependsOn: string[]
+}
+
 /**
  * Decorator to mark a property as a calculated field with dependencies.
- * @param dependencies - An array of strings indicating the dependencies.
+ * @param options - A `CalculatedFieldOptions` object indicating the dependencies.
  */
-export function CalculatedField(dependencies: string[]): PropertyDecorator {
+export function CalculatedField(options: CalculatedFieldOptions): PropertyDecorator {
   return (target: object, propertyKey: string | symbol): void => {
     const existingDependencies = Reflect.getMetadata('dynamic:dependencies', target.constructor) || {}
-    existingDependencies[propertyKey] = dependencies
+    existingDependencies[propertyKey] = options.dependsOn
     Reflect.defineMetadata('dynamic:dependencies', existingDependencies, target.constructor)
   }
 }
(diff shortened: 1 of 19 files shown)


packages/framework-core/src/decorators/read-model.ts
```diff
@@ -24,12 +24,35 @@ export function ReadModel(
       }
 
       const authorizer = BoosterAuthorizer.build(attributes) as ReadModelAuthorizer
+      const classMetadata = getClassMetadata(readModelClass)
+      const dynamicDependencies = Reflect.getMetadata('dynamic:dependencies', readModelClass) || {}
```
alvaroloes, line 28: Great!

packages/framework-integration-tests/src/read-models/cart-read-model.ts
```diff
@@ -38,14 +38,17 @@ export class CartReadModel {
     return this.checks
   }
 
+  @CalculatedField(['cartItems'])
```
alvaroloes, line 41: From a UX point of view (and also in case we need to expand the customization properties in the future) I'd say it is a bit better to make the user write `dependesOn` here. If not, it is not clear what you are defining here. What I mean is this:
```suggestion
  @CalculatedField({dependsOn: ['cartItems']})
```
This way when the user sees this, they don't even need to go to the documentation to understand it 😃

alvaroloes (commented):
Amazing job! Just a small comment about UX

alvaroloes (approved):

packages/framework-integration-tests/src/read-models/cart-read-model.ts
MarcAstr0: Suggestion applied in commit 2077b2c.

PR #1544, opened 2024-08-06
Add a .vscode/launch.json file to the default project template to ease out-of-the-box debugging in VSCode.

Virajjai:
## Description
Added a .vscode/launch.json file to the default project template to ease out-of-the-box debugging in VSCode

## Changes
-made launch.json template.
-added to project-initializer.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,13 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    // Runs npx boost -e local
+    {
+      "command": "npm run build && npx boost start -e local",
+      "name": "Start Debug Server",
+      "request": "launch",
+      "type": "node-terminal",
+      "cwd": "${workspaceFolder}"
+    },
+  ]
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,13 +1,22 @@
+
 {
   "version": "0.2.0",
   "configurations": [
-    // Runs npx boost -e local
     {
-      "command": "npm run build && npx boost start -e local",
-      "name": "Start Debug Server",
-      "request": "launch",
+      "name": "Launch Booster App",
       "type": "node-terminal",
-      "cwd": "${workspaceFolder}"
-    },
+      "request": "launch",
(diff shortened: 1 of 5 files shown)


Virajjai:
Hi there,

I’m a first-time contributor and I’m encountering an issue with workflow access. The GitHub Actions workflow requires permissions that I might not have. Could you please provide guidance on how to proceed?

Thank you!

packages/cli/src/services/project-initializer.ts
```diff
@@ -53,6 +54,7 @@ export async function generateRootDirectory(config: ProjectInitializerConfig): P
     [srcDir, 'event-handlers'],
     [srcDir, 'read-models'],
     [srcDir, 'scheduled-commands'],
+
```
javiertoledo, line 57:
```suggestion
```

packages/cli/.vscode/launch.json
```diff
@@ -4,6 +4,7 @@
   // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
   "version": "0.2.0",
   "configurations": [
+
```
javiertoledo, line 7:
```suggestion
```

packages/cli/src/templates/project/launch-json.ts
```diff
@@ -0,0 +1,14 @@
+
```
javiertoledo, line 1:
```suggestion
```

.vscode/launch.json
```diff
@@ -0,0 +1,22 @@
+
```
javiertoledo, line 1:
```suggestion
```

javiertoledo (changes requested):
Looking good, thanks for the change! I just removed a couple of white lines. It'd be nice if you could add a test to check that the launch.json file was created. You can add a line with the check here: https://github.com/boostercloud/booster/blob/main/packages/cli/test/services/project-initializer.test.ts#L80

Once you add the test I can enable the workflow runs

Virajjai:
I have added test you can check it out.

PR #1545, opened 2024-08-08
Update project-initializer.test.ts

Virajjai:


Virajjai:
I have added test for check of launch.json file during initialization of booster project.

PR #1469, opened 2023-10-16
Read Models projections by ReadModel query

gonzalojaubert:
TODO refactor

PR #1547, opened 2024-09-04
Fix projects decorator

MarcAstr0:
## Description

After the changes introduced in #1469, it was found that read model projections where the join key was a nullable array gave a "TS1241: Unable to resolve signature of method decorator when called as an expression" compilation error. This happened in read models and entities like the following:

```typescript
@Entity
export class MyEntity {
  public constructor(
    readonly id: UUID,
    readonly someListOfIds?: Array<string>
    // other properties
  ) {}
  // rest of the entity code
}
```

```typescript
@ReadModel
export class SomeReadModel {
  @Projects(MyEntity, 'someListOfIds')
  public static projectMyEntity(
    entity: MyEntity,
    readModelId: UUID,
    currentReadModel?: SomeReadModel
  ): ProjectionResult<SomeReadModel> {
    // projection logic
  }
  
  // rest of the read model code
}
```

## Changes

* Modified `ProjectionMethod` type to appropriately handle this edge case.

## Checks
- [x] Project Builds

### DIFF
--- a/packages/framework-core/src/decorators/projects.ts
+++ b/packages/framework-core/src/decorators/projects.ts
@@ -105,7 +105,7 @@ type ProjectionMethod<
 > = TJoinKeyType extends ReadModelJoinKeyFunction<TEntity, TReadModel>
   ? ProjectionMethodWithEntityConditionalReadModelIdAndReadModel<TEntity, TReadModel>
   : TJoinKeyType extends keyof TEntity
-  ? PropertyType<TEntity, TJoinKeyType> extends Array<UUID>
+  ? NonNullable<PropertyType<TEntity, TJoinKeyType>> extends Array<any>
     ? ProjectionMethodWithEntityReadModelIdAndReadModel<TEntity, TReadModel>
     : ProjectionMethodWithEntityAndReadModel<TEntity, TReadModel>
   : never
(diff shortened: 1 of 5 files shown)


packages/framework-core/src/decorators/projects.ts
```diff
@@ -105,7 +105,7 @@ type ProjectionMethod<
 > = TJoinKeyType extends ReadModelJoinKeyFunction<TEntity, TReadModel>
   ? ProjectionMethodWithEntityConditionalReadModelIdAndReadModel<TEntity, TReadModel>
   : TJoinKeyType extends keyof TEntity
-  ? PropertyType<TEntity, TJoinKeyType> extends Array<UUID>
+  ? NonNullable<PropertyType<TEntity, TJoinKeyType>> extends Array<any>
```
davidverdu, line 108: I think that the `any` modifier should be used as the last resource in TypeScript. Have you tried these approaches before choosing `any` as the solution?

(1) Maybe is suitable to create a union type for the allowed types for this type, something like:
```typescript
type AllowedPropertyType = UUID | string //we should find a better name for this

...
? NonNullable<PropertyType<TEntity, TJoinKeyType>> extends Array<AllowedPropertyType>
```

(2) another option is to try the `unknown` type instead of `any`

If both options above don't work, let's go for `any`

MarcAstr0: @davidverdu look at commit 51dfedc. I added a `UUIDLike` type following your suggestion. :wink: :+1:

davidverdu (approved):
LGTM. I guess that you don't need other types different from `string` and `UUID` in the union, do you?

PR #1550, opened 2024-09-11
Node 20 support

MarcAstr0:
## Description

Node.js 18 [will reach EOL in April 2025](https://learn.microsoft.com/en-us/azure/azure-functions/functions-versions?tabs=isolated-process%2Cv4&pivots=programming-language-typescript#languages).

This PR includes the necessary changes for Booster to run with the latest LTS release of Node.js (20), considering that Azure Functions has a GA support level for Node.js 20.

## Changes

* Update dependencies to work with Node.js 20.
* Apply small changes to the Azure provider infrastructure package due to changes in the Terraform CDK.
* Add an `upgrade-v3.md` doc listing breaking changes when upgrading from 2.x to 3.x.

## Checks
- [x] Project Builds


packages/framework-provider-azure-infrastructure/src/infrastructure/synth/gateway/terraform-application-gateway.ts
```diff
@@ -157,7 +157,7 @@ export class TerraformApplicationGateway {
         ...gatewayConfiguration,
         wafConfiguration: {
           enabled: true,
-          firewallMode: 'Detection',
+          firewallMode: 'Prevention',
```
davidverdu, line 160: Why did you need to change the firewall mode to let Node 20 work? I guess that the firewall is more protective now with this change, isn't it?

common/git-hooks/pre-push
davidverdu, line 1: why do we need this empty hook?

common/config/rush/pnpm-lock.yaml
davidverdu, line 1: I checked out this branch to explore this file thoroughly. I found in this file some `devDependencies` items like this:

```yaml
'@types/node':
        specifier: ^18.18.2
        version: 18.19.74
```

I guess that the dev dependencies need to be also updated to node 20. Could you check that?

davidverdu: I think this file has this config because `@types/node` is still set to version 18 on the `package.json` files of the packages. They should be checked as well

packages/framework-provider-azure-infrastructure/src/infrastructure/synth/gateway/terraform-application-gateway.ts
MarcAstr0: This change has nothing to do with Node 20, actually. It was introduced by mistake because some modifications to the firewall policies had to be made so that Booster infra could be deployed in the corporate environment that the code was worked on. It has been rolled back in commit 4b28b1fc.

common/git-hooks/pre-push
MarcAstr0: This hook is actually not empty, it prints a message to console when a `git push` is made. It appears that the change committed to this file is encoding-related. Hence, why the diff appears empty.

common/config/rush/pnpm-lock.yaml
MarcAstr0: Good catch! The right version of @types/node was set in commit 475e9473.

davidverdu (approved):
LGTM!

common/git-hooks/pre-push
davidverdu: ok

PR #1552, opened 2024-09-24
Fix read model queries when selecting deeply-nested arrays

MarcAstr0:
## Description

The changes in this PR fix some issues when selecting fields in read model objects that are nested deeply. Previously, if one were to select a field from an object in an array nested deeper than a top-level property, e.g., `['foo.bar.arr[].id']`, it would generate the wrong SQL code. To keep things simple, when this case is encountered, it will retrieve all the properties of the objects in the array.

## Changes

* Modify `buildProjections` and add a `preprocessProjections` method to the Cosmos query helper to identify and alter the projections with deeply nested arrays.

## Checks
- [ ] Project Builds

### DIFF
--- a/packages/framework-provider-azure/src/helpers/query-helper.ts
+++ b/packages/framework-provider-azure/src/helpers/query-helper.ts
@@ -317,6 +317,12 @@ function buildProjections(projections: ProjectionFor<unknown> | string = '*'): s
     .join(', ')
 }
 
+/**
+ * Preprocesses the projections to handle nested arrays and objects.
+ *
+ * @param {ProjectionFor<unknown>} projections - The projections to preprocess.
+ * @returns {ProjectionFor<unknown>} - The preprocessed projections.
+ */
 function preprocessProjections(projections: ProjectionFor<unknown>): ProjectionFor<unknown> {
   const processed = new Set<string>()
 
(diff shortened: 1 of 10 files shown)


packages/framework-provider-azure/src/helpers/query-helper.ts
```diff
@@ -314,6 +317,57 @@ function buildProjections(projections: ProjectionFor<unknown> | string = '*'): s
...
+      } else {
+        // Case 2.3.2: Nested arrays after it
+        const processedField = parts.slice(0, arrayIndices[1] + 1).join('.')
+        processed.add(processedField.slice(0, -2)) // Remove '[]' from the last part
+      }
+    } else {
+      // Case 2.4: Array nested deeper than a top-level property
+      const processedField = parts.slice(0, arrayIndices[0] + 1).join('.')
+      processed.add(processedField.slice(0, -2)) // Remove '[]' from the last part
+    }
```
davidverdu, line 357: I'm not sure but the three `else` blocks for cases 2.2, 2.3.2 and 2.4 are repeated. Maybe if we swap the nested conditions and put in line 333 the condition `else if (arrayIndices.length === 1)` instead, we could save some repetition of code

packages/framework-provider-azure/test/helpers/query-helper.test.ts
```diff
@@ -169,6 +169,46 @@ describe('Query helper', () => {
...
+        mockCosmosDbClient.database(mockConfig.resourceNames.applicationStack).container(`${mockReadModelName}`).items
+          .query
+      ).to.have.been.calledWith(
+        match({
+          query:
+            'SELECT c["id"], ' +
+            'ARRAY(SELECT item["z"] FROM item IN c["x"]["arr"]) AS "x.arr", ' +
+            'c["foo"]["bar"]["items"] AS "foo.bar.items", c["foo"]["bar"]["baz"]["items"] AS "foo.bar.baz.items", ' +
+            'ARRAY(SELECT item["subArr"], item["id"] FROM item IN c["arr"]) AS arr ' +
+            'FROM c ',
```
davidverdu, line 206: You may want to use backticks ``` instead for this multiline string, as it is explained https://eranstiller.com/javascript-multiline-strings

```typescript
query: `
SELECT c["id"], 
ARRAY(SELECT item["z"] FROM item IN c["x"]["arr"]) AS "x.arr", 
c["foo"]["bar"]["items"] AS "foo.bar.items", c["foo"]["bar"]["baz"]["items"] AS "foo.bar.baz.items", 
ARRAY(SELECT item["subArr"], item["id"] FROM item IN c["arr"]) AS arr 
FROM c
`
```

packages/framework-provider-azure/src/helpers/query-helper.ts
```diff
@@ -314,6 +317,57 @@ function buildProjections(projections: ProjectionFor<unknown> | string = '*'): s
...
+  })
+
+  // Convert the Set back to the original type of projections
+  if (Array.isArray(projections)) {
+    return Array.from(processed) as ProjectionFor<unknown>
+  } else {
+    return Array.from(processed).reduce((acc, field) => {
+      ;(acc as any)[field] = field
```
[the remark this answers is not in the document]
alvaroloes: 😳 It seems accidental to me

packages/framework-provider-azure/src/helpers/query-helper.ts
alvaroloes: This code is a bit hard to read. Question: Why `arrayIndices[0] === 1` means "Array nested within a top-level property"?

packages/framework-provider-azure/src/helpers/query-helper.ts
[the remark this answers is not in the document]
MarcAstr0: It was actually required by Prettier. It's called a "defensive semicolon" or "leading semicolon."

packages/framework-provider-azure/test/helpers/query-helper.test.ts
MarcAstr0: The problem with the backticks is that they add newline characters that are not in the generated string we're comparing it to. Therefore, we have to use string concatenation. Otherwise, the tests fail.

packages/framework-provider-azure/src/helpers/query-helper.ts
MarcAstr0: Thanks for the suggestion. Applied in commit 6813d46.
Initially, it was this lengthy because it was easier to follow which cases apply.

davidverdu (approved):
LGTM. But Integration tests should pass as well

PR #1553, opened 2024-09-27
Fix read model queries when selecting deeply-nested arrays (local provider)

MarcAstr0:
## Description

This PR is related to #1552. It fixes the same issue but for the local provider. It changes the approach to selecting fields when querying read models in the local provider. Instead of passing projections to NeDB, the full object will be retrieved, and the fields indicated in the selection will be filtered in.

## Changes

* Modifies `read-model-registry.ts` to change how read models are queries.
* Test added to `read-model-registry.test.ts` for querying a complex read model.

## Checks
- [x] Project Builds


davidverdu (approved):
LGTM

PR #1558, opened 2024-11-24
Fix optimistic concurrency issue with read model projections

MarcAstr0:
## Description

The provider libraries have two methods for read model retrieval: `fetch` and `search`. In the implementation for the Azure provider, only the former was adding an `optimisticConcurrencyValue` property to the `boosterMetadata`. This value stores the Cosmos DB `_etag` property for the document and is used to handle optimistic concurrency scenarios.

Read model projection issues related to optimistic concurrency have been found due to the `search` method being used instead of `fetch`. Some read models weren't properly projected when a simultaneous update occurred because the optimistic concurrency verification wasn't happening due to the `optimisticConcurrencyValue` property not being in the `boosterMetadata`. 

This PR fixes this by including this property in the `search` method implementation results.

## Changes

* Modified the Azure provider's read model `search` method so that it writes the `_etag` as the `optimisticConcurrencyValue` in the `boosterMetadata`.

## Checks
- [x] Project Builds

### DIFF
--- a/common/changes/@boostercloud/framework-core/fix_optimistic_concurrency_2024-11-24-17-37.json
+++ b/common/changes/@boostercloud/framework-core/fix_optimistic_concurrency_2024-11-24-17-37.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@boostercloud/framework-core",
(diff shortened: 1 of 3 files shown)


davidverdu (approved):
LGTM

PR #1560, opened 2024-12-17
fix: check if property exists on object without executing

hendrik-depauw-lemon:
## Description
When populating the getters on a ReadModel, the getters are executed twice. Once where you would expect it, and once in an `if` statement to check if the property exists on the ReadModel. In this `if` statement, there is no check to see if the getter is a promise or not, so it is never awaited. When the getter-promise would eventually return an error, the whole Booster application crashes. 

This PR thus fixes two things:
- Stops unnecessarily executing all getters on ReadModels twice
- Prevents the app from crashing when an async getter throws an error

## Changes

Replaced `source[key] !== undefined` with `key in source`.

## Checks
- [ ] Project Builds


MarcAstr0 (approved):
Looks good. LGTM!

PR #1561, opened 2024-12-26
Enhance storing events in batches in Azure provider

MarcAstr0:
## Description

PR #1436 added the ability to store events in batches. However, there are times when one might want to store events individually due to cost or performance reasons. This PR adds the ability to enable or disable batching events via configuration. It also adds more options for the batch inserts, such as setting the size of the chunks.

## Changes

* Added the `azureConfiguration` property to the `BoosterConfig` object to set options for batch storage of events.
* Split the `storeEvents`method into two to handle batch and non-batch scenarios.
* Updated unit tests.
* Updated documentation.

## Checks
- [x] Project Builds


davidverdu (approved):
LGTM

PR #1571, opened 2025-03-05
feat: update graphql type informer to handle circular dependencies be…

hendrik-depauw-lemon:
## Description
The graphql type informer could not handle circular dependencies between types. The result would always be JSON. By using thunks, we can fix this issue.

## Changes
- Use thunks to define graphql fields

## Checks
- [ ] Project Builds

## Additional information
We are using `Object.assign` on the `graphqlTypes` map so that the references to the types do not change. However, when we want to update the fields of the type, we need an internal function from the graphql package that is currently not exported. What course of action do you guys suggest? Should we open a PR to graphql? Do we patch graphql? Do we copy the internals of the function? Other options?

### DIFF
--- a/packages/framework-core/src/services/graphql/graphql-type-informer.ts
+++ b/packages/framework-core/src/services/graphql/graphql-type-informer.ts
@@ -14,6 +14,8 @@ import {
   GraphQLType,
   GraphQLUnionType,
 } from 'graphql'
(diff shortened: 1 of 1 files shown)


MarcAstr0 (changes requested):
The only change needed would be to work around the no longer accessible usage of `defineFieldMap`.

PR #1577, opened 2025-04-18
feat: gracefully handle shutdown signals

BelgianNoise:
## Description

I was having trouble generating code coverage for my tests with `nyc`.
Took me some time but I found out that `nyc` requires the process it spawn to gracefully exit.
I was not completely sure what that meant, so I had to do some digging.
This PR adds handling of signals to gracefully shut down the process.

## Changes

- Added process signal listeners to gracefully shut down the server

## Checks
- [x] Project Builds

## Additional information

I added a `cleanup()` function that is currently empty, question is: does Booster have dedicated functions to close down the server, clean data, close connections ?
Didn't add docs on the functions as I find them self explanatory.

This fixes the issue for me locally, but I am not entirely sure if this is the way to go. (approach + location in Booster code itself)

Does this require any documentation to be updated ?

### DIFF
--- a/packages/framework-core/src/booster.ts
+++ b/packages/framework-core/src/booster.ts
@@ -63,6 +63,7 @@ export class Booster {
    * Initializes the Booster project
    */
   public static start(codeRootPath: string): void {
+    registerShutdownSignalHandlers()
     const projectRootPath = codeRootPath.replace(new RegExp(this.config.codeRelativePath + '$'), '')
     this.config.userProjectRootPath = projectRootPath
     Importer.importUserProjectFiles(codeRootPath)
@@ -209,4 +210,23 @@ function checkAndGetCurrentEnv(): string {
     )
   }
   return env
+}
+
+function registerShutdownSignalHandlers(): void {
+  const signals = ['SIGINT', 'SIGTERM', 'SIGQUIT']
+  signals.forEach((signal) => {
+    process.on(signal, () => {
+      shutDownGracefully(signal)
+    })
+  })
+}
+
+function shutDownGracefully(signal: string): void {
+  console.log(`Received signal '${signal}'. Shutting down gracefully...`)
+  cleanup()
+  process.exit(0)
+}
+
+function cleanup(): void {
+  // Does Booster have dedicated clean up to do ?
 }


packages/framework-core/src/booster.ts
```diff
@@ -210,3 +211,22 @@ function checkAndGetCurrentEnv(): string {
...
+}
+
+function shutDownGracefully(signal: string): void {
+  console.log(`Received signal '${signal}'. Shutting down gracefully...`)
+  cleanup()
+  process.exit(0)
+}
+
+function cleanup(): void {
+  // Does Booster have dedicated clean up to do ?
```
MarcAstr0, line 231: @BelgianNoise Booster currently does not have a dedicated cleanup upon shutdown, but I think it would be beneficial to implement one. It should probably:

* Stop accepting incoming GraphQL connections/requests and close WebSocket connections.
* Close connections to DB and read model stores (this would have to be provider-specific).
* Close any other connections, such as to Event Hubs.

arthur-joppart-lemon: Thank you so much for your feedback, we have done some more internal debugging and found that it was an issue in our provider that was causing issues with the child process shenanigans & `nyc`.

That being said, this could still be a nice addition to the core package imo. However I won't be able to focus on it right now as I would need to find time outside working hours to make these changes and get more familiar with the code :)

I am stil eager to go deeper into the code and add the things you mentioned above, but it won't be in teh short term.
I will revert my PR back to a draft and I am leaving it up to you if you would rather see it closed for now.

(Feel free to hijack this PR if desired)

BelgianNoise: Thank you so much for your feedback, we have done some more internal debugging and found that it was an issue in our provider that was causing issues with the child process shenanigans & nyc.

That being said, this could still be a nice addition to the core package imo. However I won't be able to focus on it right now as I would need to find time outside working hours to make these changes and get more familiar with the code :)

I am stil eager to go deeper into the code and add the things you mentioned above, but it won't be in teh short term.
I will revert my PR back to a draft and I am leaving it up to you if you would rather see it closed for now.

(Feel free to hijack this PR if desired)

Edit: I do not have access to revert back to draft :)

(Oops, commented with diff account earlier)

MarcAstr0: @BelgianNoise I think we can go ahead with this addition and eventually implement the cleanup method in future PRs :)

PR #1578, opened 2025-05-17
Fix logger name in event stream consumer

santigamo:
## Summary
- fix incorrect logger identifier in BoosterEventStreamConsumer

## Testing
- `node common/scripts/install-run-rush.js test` *(fails: EHOSTUNREACH)*
- `npm test` in `packages/framework-core` *(fails: cross-env not found)*

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/hydrogen
+lts/iron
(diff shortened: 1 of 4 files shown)


MarcAstr0 (approved):
@santigamo approved! But first, can you add the change file? You generate it with `rush change` command. :)

PR #1579, opened 2025-06-03
feat: Improve event store resilience

MarcAstr0:
## Description

This PR introduces a robust retry mechanism with error filtering capabilities for event store operations in Booster. The implementation includes exponential backoff with jitter, configurable retry attempts, and fine-grained control over which errors should be retried.

## Changes

- Added `retryWithBackoff` function in `framework-common-helpers` with:
  - Exponential backoff with jitter
  - Configurable retry attempts, delays, and backoff factors
  - Error filtering capabilities
  - Comprehensive logging of retry attempts

- Enhanced `RetryConfig` interface with new properties:
  - `retryAllErrors`: Optional boolean to control default retry behavior
  - `nonRetryableErrors`: Optional array of error class names that should never be retried
  - `retryableErrors`: Optional array of error class names that should be retried when `retryAllErrors` is false

- Integrated retry mechanism in core event store operations:
  - Event storage in `RegisterHandler`
  - Dispatched event storage in `EventStore`
  - Snapshot storage in `EventStore`

- Added comprehensive test suite covering:
  - Basic retry functionality
  - Error filtering behavior
  - Delay calculation with jitter
  - Logging behavior
  - Edge cases and error scenarios

- Added Azure provider-specific retry configuration:
  - Introduced environment variables for Cosmos DB client retry behavior:
    - `COSMOSDB_MAX_RETRIES`: Controls maximum retry attempts
    - `COSMOSDB_RETRY_INTERVAL`: Sets fixed retry interval in milliseconds
    - `COSMOSDB_MAX_WAIT_TIME`: Defines maximum wait time in seconds
  - Updated CosmosClient initialization to use these settings when available
  - Maintains backward compatibility by using client defaults when variables are not set

## Checks
- [x] Project Builds

## Additional information

The retry mechanism follows these rules:
1. If `retryAllErrors` is true (default) or undefined, all errors are retried except those in `nonRetryableErrors`
2. If `retryAllErrors` is false, only errors in `retryableErrors` are retried
3. `nonRetryableErrors` takes precedence over other settings
4. All new properties are optional to maintain backward compatibility

The implementation uses exponential backoff with jitter to prevent thundering herd problems and provides detailed logging for debugging purposes.

The Azure provider changes allow for fine-tuning of the Cosmos DB client's retry behavior through environment variables, separate from the general event store retry configuration. This provides additional flexibility for Azure-specific scenarios while maintaining the default behavior when no custom configuration is provided.


website/docs/10_going-deeper/retry-configuration.mdx
```diff
@@ -0,0 +1,100 @@
...
+```
+
+## Retry Behavior
+
+The retry mechanism follows these rules:
+
+1. If `retryAllErrors` is true (default) or undefined, all errors are retried except those in `nonRetryableErrors`
+2. If `retryAllErrors` is false, only errors in `retryableErrors` are retried
+3. `nonRetryableErrors` takes precedence over other settings
+4. The delay beteen retries follows an exponential backoff pattern with jitter to prevent thundering herd problems
```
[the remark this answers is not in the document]
MarcAstr0: Fixed in commit 6f75eba.

packages/framework-core/src/booster-register-handler.ts
```diff
@@ -31,9 +32,16 @@ export class RegisterHandler {
...
+
+    await retryWithBackoff(
+      () =>
+        config.provider.events.store(
+          register.eventList.map((event) => RegisterHandler.wrapEvent(config, event, register)),
+          config
+        ),
+      config.eventStoreRetry,
+      logger
     )
```
davidverdu, line 45: Is correct to assume that if the `store` method throws an error the event is not stored and it is safe to retry, guaranteeing that two copies of the same event are not stored, just only one?

packages/framework-core/src/services/event-store.ts
```diff
@@ -117,7 +121,11 @@ export class EventStore {
     const logger = getLogger(this.config, 'EventStore#storeSnapshot')
     try {
       logger.debug('Storing snapshot in the event store:', snapshot)
-      return await this.config.provider.events.storeSnapshot(snapshot, this.config)
+      return await retryWithBackoff(
+        () => this.config.provider.events.storeSnapshot(snapshot, this.config),
+        this.config.eventStoreRetry,
+        logger
+      )
```
davidverdu, line 128: Same as above. Is correct to assume that if the `storeSnapshot` method throws an error the event is not stored and it is safe to retry, guaranteeing that two copies of the same event are not stored, just only one?

packages/framework-common-helpers/src/retrier.ts
```diff
@@ -40,3 +40,80 @@ function checkRetryError(e: Error, errorClassThatRetries: Class<Error>, logger?:
...
+ * Calculates the delay for a retry attempt using exponential backoff with jitter.
+ * @param attempt Current attempt number (1-based)
+ * @param config Retry configuration
+ * @returns Delay in milliseconds
+ */
+export function calculateRetryDelay(attempt: number, config: RetryConfig): number {
+  const baseDelay = Math.min(config.initialDelay * Math.pow(config.backoffFactor, attempt - 1), config.maxDelay)
+  const jitter = baseDelay * config.jitterFactor * (Math.random() * 2 - 1)
+  return Math.min(baseDelay + jitter, config.maxDelay)
+}
```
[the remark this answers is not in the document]
MarcAstr0: Good suggestion. Applied in commit 3b0eaa0.

packages/framework-core/src/services/event-store.ts
MarcAstr0: @davidverdu Yes, it's safe to retry `storeSnapshot` as well. While it does perform a query to check for existing snapshots before inserting, this is a read-only operation (see [here](https://github.com/boostercloud/booster/blob/017cc1595b2ddd3bc7eff62ccff5b89ae2fe77af/packages/framework-provider-azure/src/library/events-adapter.ts#L93)). The only write operation is the final `container.items.create()`. If this fails, no snapshot is stored, making retry safe. The previous query operation doesn't affect the data, so retrying won't result in duplicate snapshots.

davidverdu (approved):

PR #1587, opened 2025-08-18
Refactor Azure read model search to use Cosmos DB's native continuation token pagination

MarcAstr0:
## Description

Currently, the implementation for read model search queries in the Azure provider uses custom pagination based on `OFFSET` and `LIMIT`. This approach is causing records to either not appear in pages or be duplicated between pages, The root cause is that Cosmos DB does not guarantee consistent ordering for SQL queries without an explicit `ORDER BY` clause, meaning that the same query can return results in different orders between page requests. This makes `OFFSET`-based pagination unreliable even without concurrent writes to the container. Additionally, the current cursor calculation logic simply increments by the limit size, which assumes no gaps in data and stable ordering. Using Cosmos DB's native continuation token pagination should correct this.

## Changes

* Refactors the `search()` method in `query-helper.ts` to handle paginated queries using the continuation token.

## Checks
- [x] Project Builds

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -16,7 +16,7 @@ importers:
         specifier: 3.7.13
         version: 3.7.13(graphql@16.10.0)(react@17.0.2)(subscriptions-transport-ws@0.11.0(graphql@16.10.0))
       '@boostercloud/framework-types':
-        specifier: workspace:^3.4.0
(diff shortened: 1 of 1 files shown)


davidverdu (approved):
Approved with a comment. I would add an integration test or a test in packages/framework-provider-azure/test/helpers/query-helper.test.ts to ensure all the paths of the new code are successfully executed

